PR #165, opened 2026-08-02
Workflow execution and publication hardening (landing in slices)

aviggiano:
## Summary

- Harden benchmark execution and publication provenance so every task is pinned to, executes from, and attests the exact compiled revision.
- Publish a fresh three-target `deepseek / deepseek-v4-flash / max` smoke benchmark judged by `gpt-5.6-sol / xhigh`.
- Require provider-reported alias identity, an unverified provider version, complete usage accounting, and the exact verified DeepSeek V4 Flash rates.
- Keep this benchmark publication separate from releases; this PR will not create, edit, delete, or retarget a release.

## Status

The earlier benchmark attempt is invalid, has been removed from branch ancestry, and must not be cited. Pre-freeze durability and security findings are being fixed before one candidate SHA is frozen. Fresh benchmark metrics, artifact digest, exact-SHA gates, and three independent review results will be added here only after they are verified.

## Required publication profile

- Runner: `deepseek / deepseek-v4-flash / max`
- Judge: `gpt-5.6-sol / xhigh`
- Identity: `provider-reported-alias`; provider version `unverified`
- Rates (USD per million): input `0.14`, cache read `0.0028`, output `0.28`, reasoning `0.28`, cache write `null`
- Targets: one pinned Foundry, Hardhat, and Vyper smoke target

## Remaining before merge

- Finish all confirmed pre-freeze fixes and full local validation
- Freeze and gate one exact candidate SHA under systemd
- Obtain and aggregate three independent reviews of that exact SHA
- Pass GitHub CI and validate a fresh feature-branch benchmark artifact
- Merge without a release, rerun from exact `main`, and verify automatic history/README-chart publication

<h3>Greptile Summary</h3>

The PR hardens benchmark execution and history publication so computation, terminal evidence, and published results remain bound to an exact candidate revision.

- Adds exact-SHA workflow gates and a fixed DeepSeek V4 Flash smoke profile.
- Introduces sealed, digest-validated benchmark publication handoffs and compare-and-swap history updates.
- Strengthens runtime workflow provenance, terminal-state durability, recovery, usage accounting, and sensitive-value redaction.
- Updates CLI lifecycle behavior, documentation, schemas, and associated tests.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| .github/workflows/eval-benchmarks.yml | Adds exact-candidate dispatch binding, explicit benchmark scopes, fixed profile validation, and scope-aware recovery metadata. |
| .github/workflows/eval-history-publication.yml | Splits validation from privileged publication using a sealed, digest-bound artifact handoff. |
| packages/runtime/src/workflow-integrity.ts | Introduces workflow integrity validation supporting exact compiled-revision execution. |
| packages/runtime/src/start-run.ts | Integrates stronger repository provenance, workflow capability, and verifier checks into run startup and lifecycle operations. |
| packages/runtime/src/workflow-sync.ts | Expands durable reconciliation of workflow state, terminal evidence, attempts, and usage accounting. |
| packages/modal/src/public-bundle.ts | Strengthens public benchmark bundle validation, lineage binding, and publication-safe evidence handling. |
| packages/evals/src/terminal-disposition.ts | Adds centralized terminal evidence validation and failure classification for publication decisions. |
| packages/artifacts/src/events.ts | Hardens event persistence and recovery, including handling of torn trailing records. |
| packages/security/src/sensitive-redaction.ts | Extends credential redaction to encoded secret representations before evidence crosses trust boundaries. |
| scripts/ci/publish-eval-history-cas.mjs | Adds digest-bound handoff validation and remote-tip compare-and-swap publication behavior. |

</details>

```mermaid
sequenceDiagram
  participant Dispatch as Benchmark workflow
  participant Runtime as Runtime / Modal
  participant Artifact as Terminal artifact
  participant Validator as Publication validator
  participant Publisher as History publisher
  Dispatch->>Dispatch: Bind run to expected candidate SHA
  Dispatch->>Runtime: Launch fixed benchmark profile
  Runtime->>Artifact: Persist terminal results and provenance
  Artifact->>Validator: Transfer sealed publication handoff
  Validator->>Validator: Verify generation and tree digests
  Validator->>Validator: Confirm candidate remains reachable from main
  Validator->>Publisher: Authorize validated generation
  Publisher->>Publisher: Compare-and-swap history update
```
</details>

<sub>Reviews (15): Last reviewed commit: ["Drop the duplicated runner test import l..."](https://github.com/monad-developers/ultrafuzz/commit/fe1376e9e286ed3bada11adcca77e96c73c81fce) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=49510899)</sub>

<details><summary><h4>Context used (5)</h4></summary>

- Knowledge Base — [CLI (`@ultrafuzz/cli`)](https://app.greptile.com/monad-foudnation/-/custom-context/knowledge-base/monad-developers/ultrafuzz/-/docs/cli.md)
</details>


aviggiano:
## Review finding: plural Foundry test roots are dropped on this branch (issue #212)

Verified on this branch at `a9c9428` against `origin/main` at `77da78a`:

- `invariantTestRoots` occurrences in `packages/runtime/src/templates/smithers/workflows/workflow.tsx`: **2 on `origin/main`, 0 on this branch** (dropped in merge `2bbecba`).
- `workspaceOutputRootsForTask` in `packages/runtime/src/smithers.ts`: **present only on this branch** (2 occurrences), supplying static singular-only roots around `smithers.ts:4191-4212` — `artifacts/<attemptId>`, `test/foundry/<nodeId>`, plus `test/recon|chimera|invariants`.

### Why this matters beyond this PR

The Aave v4 benchmark target at `6959e321` has **zero files under `test/`** — it uses plural `tests/`. After this branch merges, every invariant worktree gets an empty singular `test/recon|chimera|invariants` tree, which `setup.md:101-102` explicitly instructs the model to read as the repository's own convention. Retry cleanup also stops cleaning `tests/` leftovers.

This is not a hard break — dual-root git discovery and the workspace-patch `git add -A` path both survive — which is exactly why it is a cheap guard **now** rather than a regression hunt later. It lands directly on the repository another active agent is using for its invariant-only Aave v4 run.

### Requested change

Either restore repository-aware roots in `workspaceOutputRootsForTask` (detect `test/` vs `tests/` rather than hard-coding singular), or state the singular-only intent explicitly in this PR's description and open a follow-up issue.

Please treat this as an actionable review finding for the current candidate SHA. Tracking issue: #212.

aviggiano:
## Addressed in `4b49e9f` — repository-aware roots restored

Thanks — this was a real regression and the analysis was correct. Fixed rather than accepted; I did **not** take the "state singular-only intent and open a follow-up" option.

### Confirmed the finding first

At the reviewed head: `invariantTestRoots` occurrences in `workflow.tsx` were **0 on the branch vs 2 on `origin/main`** (dropped in merge `2bbecba`), and `workspaceOutputRootsForTask` in `smithers.ts` supplied static singular-only roots. Exactly as reported.

### Where the fix landed, and why not where you suggested

The requested change was to make `workspaceOutputRootsForTask` (`smithers.ts`) repository-aware. I deliberately did not fix it there: that function runs inside `compileTask`, which builds the task spec **before the worktree exists**, so it has nothing to detect against. `workspacePath` at that point is a path that has not been materialised.

So the compiled spec still names singular `test/...` — it is a declaration, not a decision — and the roots are re-anchored at **execution** time in the generated workflow, which is the same layer `origin/main`'s `invariantTestRoots` used. Two new helpers in `workflow.tsx`, `invariantTestRootName` and `repositoryAwareWorkspaceOutputRoots`, feed all three consumers: `taskTestOutputRelativeRoots`, `taskWorkspaceOutputRoots` and `prepareTaskWorkspaceOutputRoots`. Net effect for the Aave v4 target at `6959e321`: invariant output is anchored under the real `tests/`, no empty singular `test/recon|chimera|invariants` tree is created, and retry cleanup targets the real output instead of an invented tree.

### Three follow-on defects the review round on the first cut caught

The first version of this fix was wrong in three ways, all now fixed in the same commit:

1. **It mapped onto every root that existed, not one.** That recreated your defect from the other side: as soon as an ancestor stage introduces the other root — `base-test-setup`'s prompt writes under `test/foundry/` — each invariant stage would create an *empty duplicate* tree beside the real one, which is the same misleading-convention problem you described. It now selects exactly one root, in `INVARIANT_TEST_ROOT_NAMES` preference order, so a repository that has `test/` behaves exactly as it did before this helper existed and only a repository without it is redirected.
2. **Discovery ran twice and could disagree with itself.** `prepareTaskWorkspaceOutputRoots` finishes by replaying ancestor workspace patches, which can create the other test root as a side effect. A later independent discovery could therefore select a root whose subdirectories were never created, and `taskWorkspaceOutputRoots` would `lstat` a missing path and throw a bare `ENOENT` instead of the typed artifact-contract failure — killing a `retries: 0` node on exactly the restart and Modal-resume paths. The resolution is now pinned per attempt at preparation time.
3. **Every stat error was read as "root absent".** `EACCES` or `ELOOP` on a real `tests/` would silently fall back to the singular default and reintroduce this very bug. Only `ENOENT`/`ENOTDIR` now mean absent.

### Tests

New `packages/runtime/test/workspace-output-roots-repository-aware.test.ts` lifts the real template helpers and runs them against real temp directories: plural-only, singular-only, both-roots, neither, symlinked root, regular file named `test`, `testing/`-prefix confusion, pass-through/dedup, per-attempt pinning, and per-attempt pin isolation.

One of these was initially vacuous and was caught in review: the "non-directory root" case asserted the fallback value, so *rejecting* the file and *wrongly adopting* it produced the same output. It now runs against a repository that also has a real `tests/`, and I verified by mutation that deleting the `isDirectory()` guard makes it fail. The generated-source assertions in `runtime.test.ts` now also pin `taskWorkspaceOutputRoots`; previously a partial revert of just that call site left the whole suite green while every task on a `tests/`-only target would have died on a bare `ENOENT`.

### Known limitation, stated explicitly

Selection is by preference order, so once anything introduces `test/` into a worktree, subsequent stages in that worktree anchor there. For a `tests/`-only target that only happens if an upstream stage creates `test/` itself, in which case `test/` is genuinely in use by the pipeline. I judged that better than the alternatives of writing to both roots (defect 1 above) or of overriding a root the pipeline is already using.

---

Separately, on the semgrep comment about `createHash("sha1")` in `workspace-provenance.ts`: that call computes a **git blob object ID** (`blob <len>\0<contents>`) for comparison against git's own index and `HEAD` descriptors. The algorithm is dictated by git's object model, so substituting SHA-256 would produce identifiers that match no git object and break the comparison outright. Flagging it here as knowingly accepted rather than overlooked; the semgrep check itself reports success.

aviggiano:
## The `Modal Eval Benchmarks` gate on this branch is failing reproducibly, and the reason is not recorded anywhere

Collected while investigating why no DeepSeek V4 Flash run has reached Modal. The blocker is not Modal capacity or credentials — it is this branch's own smoke gate, and it has been red 3/3 since 2026-08-06.

### Failure chain

The gate launches one canary pair, `benchmark-smoke-gpt-5-6-luna-high`. `launch` succeeds; `collect` fails:

```
Error: public eval diagnostics are not safely collectable for benchmark-smoke-gpt-5-6-luna-high
    at collectModalBenchmark (packages/modal/dist/runner.js:2191)
```

That is `packages/modal/src/runner.ts:2568`, reached because the worker's terminal record says:

```json
{ "model_work_started": true, "exit_category": "sandbox-exited",
  "runtime_ms": 315786, "diagnostic_code": "public-eval-diagnostics-invalid" }
```

`public-eval-diagnostics-invalid` is set at `packages/modal/src/public-worker.ts:112`, which means `buildDiagnostics()` threw (`public-worker.ts:361`). Note what that implies: `runAndCheckpointPublicEvalDiagnostics` rethrows `runError` in preference to the build error, so **`runEval()` resolved without throwing** — the eval "completed" in ~5 minutes and then the diagnostics document could not be built.

`packages/modal/src/launch-state.ts:593` maps that code to `permanent-operational-failure`, and the workflow's soft-fail path only forgives `resume-required` with diagnostics collected. So it hard-fails the gate and the PR stays `UNSTABLE`.

### It is branch-specific and it is a regression

| run | branch | date | terminal | runtime |
|---|---|---|---|---|
| 30798948766 | this branch | 08-03 08:50 | **succeeded**, `worker-finished` | 2,021 s |
| 31118120839 | this branch | 08-06 15:58 | `public-eval-diagnostics-invalid` | 313 s |
| 31160165581 | this branch | 08-07 08:04 | `public-eval-diagnostics-invalid` | 364 s |
| 31171579070 | this branch | 08-07 10:48 | `public-eval-diagnostics-invalid` | 316 s |
| 31148699320 | `main` | 08-07 04:51 | `dependency-unreachable` / `resume-required` | 497 s |
| 31155350838 | `main` | 08-07 06:50 | `dependency-unreachable` / `resume-required` | 490 s |
| 31172097085 | `main` | 08-07 10:56 | `dependency-unreachable` / `resume-required` | 767 s |

`main` fails differently and softer (diagnostics collect fine there). This branch passed with a 34-minute run on 08-03 and now dies after ~5 minutes, three times running. The regression window is 08-03 → 08-06 on this branch, which includes the lifecycle-and-publication-hardening commits that touch this exact plumbing.

### What cannot currently be determined

The cause of the build error is discarded. `worker.log` is content-free by design (`operation-started` / `operation-finished` only), and the worker's top-level catch drops the rejection value — #307. So there is no record of *why* the eval produced nothing scoreable in five minutes.

One candidate to check first, offered as a hypothesis rather than a finding: in `packages/modal/src/public-eval-diagnostics.ts`, `publicEvalFailedNodes` is fully try/caught and returns `[]` on any problem, so it cannot be the thrower — but `readDiagnosticRunSummary` runs `z.array(matrixRowInputSchema).min(1).max(MAX_ROWS).parse(matrixValue)` outside any guard, and the final `parsePublicEvalDiagnostics(...)` at line 195 is an unguarded strict parse. An eval that resolved having produced zero matrix rows would throw at the first of those. Confirming it needs the sanitized cause described on #307.

aviggiano:
## A single dispatched Flash run needs less plumbing than expected — and no new secret

Following up on the gate analysis above with what it actually takes to run **one** `deepseek-v4-flash` smoke benchmark on API-key billing, rather than waiting on a full four-provider matrix.

### Already in place — no work needed

- **`DEEPSEEK_API_KEY` is already a repository secret** (added 2026-08-03), and `eval-benchmarks.yml` already forwards it to the `launch`, `collect` and cleanup jobs — gated to `github.event_name == 'workflow_dispatch'`, which is exactly the trigger we want. No secret needs creating or rotating.
- **API-key billing is already the only mode for these runs.** `scripts/ci/prepare-modal-benchmarks.mjs` hardcodes `auth_mode: "api-key"` on every synthesized benchmark model, and `packages/modal/src/config.ts:60` independently rejects a DeepSeek model that is not `api-key`. Subscription forwarding cannot happen here.
- **The single-runner smoke matrix is already supported and tested.** `benchmarkModels()` accepts a one-entry `BENCHMARK_MODELS_JSON` from *any* provider in smoke mode (`validated.length === 1 ? [validated[0].provider] : []`), and `packages/modal/test/ci-config.test.ts:345` already exercises that with a DeepSeek entry.
- **No lane or registry change is needed.** `scripts/ci/validate-modal-benchmark-launch.mjs` uses `benchmarks/lanes.json` only for `trials_per_variant` and target/row counts; it never requires the runner profile to appear in `lane.model_profiles`. So `assertFixedBenchmarkProfiles` in `packages/evals/src/benchmark-manifest.ts` — which pins the checked-in smoke lane to exactly `gpt-5.6-luna high` — is not violated by dispatching a different runner, because that guardrail validates the checked-in manifest, not the dispatch matrix. Model pricing resolves live from models.dev, so Flash needs no checked-in rate entry either.

### The one real gap

`.github/workflows/eval-benchmarks.yml` hardcodes `BENCHMARK_MODE` to `full` for every `workflow_dispatch`, and hardcodes the smoke matrix to a single OpenAI entry. So there is no way to *ask* for a single-runner smoke run, even though every layer beneath the workflow supports it.

Proposed change (drafted, ready to push on request):

- Four new dispatch inputs: `benchmark_mode` (`full` | `smoke`, default `full` so current behaviour is unchanged), `smoke_provider`, `smoke_model` (default `deepseek-v4-flash`), `smoke_reasoning` (default `max`).
- `BENCHMARK_MODE` honours `inputs.benchmark_mode` on dispatch.
- The matrix step emits a one-entry `BENCHMARK_MODELS_JSON` for a dispatched smoke run; the push path is untouched.
- The credential preflight requires only the selected provider's key instead of all four. `OPENAI_API_KEY` stays required for every lane because the judge profile is `gpt-5.6-sol` on CodexAgent.

Lane knobs, targets, loops and judge are untouched — only the runner varies, which is the same shape #277 needs for its five-model matrix.

### Order of operations

1. **The `public-eval-diagnostics-invalid` regression still has to be fixed first.** A dispatched Flash run would go through the same `collect` path that is currently refusing the canary, so it would fail the same way.
2. Land the dispatch inputs.
3. Dispatch once with `benchmark_mode=smoke`, `smoke_provider=deepseek`, `smoke_model=deepseek-v4-flash`, `smoke_reasoning=max`.
4. **Note where it must be dispatched from.** `scripts/ci/qualify-modal-benchmark-publication.mjs` only qualifies producer runs on the **default branch** with `launch` and `collect` both successful. A dispatch from `bench/deepseek-v4-flash-results` will produce artifacts and a validated bundle but will never publish to `benchmarks/history.json`. That is fine as a *validation* run for the rates and telemetry this PR requires; the publishing run has to happen from `main` after merge.

aviggiano:
## Diagnosis of the failing `collect` job

Pulled the artifacts from run `31171579070` (identical shape in `31160165581` — two consecutive reproductions, not a flake).

**This is not a DeepSeek failure.** The plan file for that run contains exactly one pair — `ultrafuzz-bench-benchmark-smoke-gpt-5-6-luna-high` — the control. There is no DeepSeek pair in the run at all.

`result.json`:
```json
{ "exit_category": "sandbox-exited", "runtime_ms": 315786,
  "counts": { "succeeded": 0, "failed": 0, "remaining": 0 },
  "usage": null, "generation": 3, "launch_generation": 1,
  "model_work_started": true, "diagnostic_code": "public-eval-diagnostics-invalid" }
```

The sandbox exited at ~5m16s with `exit_code: 1`, having recorded no usage and no node transitions. Because `model_work_started: true`, the collector required the public-eval diagnostics file, which was never written → `public-eval-diagnostics-invalid` → `permanent-operational-failure`. The soft-fail hatch only forgives `resume-required` **with** `diagnostic_collection_status: "succeeded"`; the outcome here records `"failed"`, so it cannot apply.

The `max_attempts: 2 → 3` bump on this branch is already live and did not rescue the pair — the run reached generation 3.

**Correction to an earlier theory:** this is not the identity or pricing throw in `public-eval-diagnostics.ts`. Validation never got that far; the run metadata never existed. `usage: null` and all-zero counts prove no accounting was ever written.

**Why it is not yet diagnosable:** `worker.log` in the preserved bundle is 2,824 bytes containing no error text — `operation-started`/`operation-finished` timestamps end to end. That is #307.

Filed: #319 (the sandbox death), #320 (the misclassification), #321 (the hatch), #322 (contradictory recovery summaries), #323 (three `git ls-files` sites still on Node's 1 MB default `maxBuffer` — the leading root-cause hypothesis for #319).

Also note this PR fixes #212 but carries no `Closes #212`, and its single-root `invariantTestRootName()` collides silently with main's list-based `invariantTestRoots()` — `merge-tree` reports no conflict, so both models would land in one file. See the comment on #212.

Tracked in #327.

aviggiano:
## Correction to my diagnosis above

My earlier comment said the sandbox exited at ~5m16s. **That was wrong**, and the correction matters because it changes what has to be fixed.

The worker did not die. From the same bundle:

- `worker.log` starts `10:51:16.209Z worker-started` and its **last line is `10:56:31.979Z operation-finished`**, with **zero** `operation-failed` lines.
- `runtime_ms: 315786` from that start lands at `10:56:31.995Z` — 16 ms after the final log line.

The worker ran to completion and wrote its own terminal contract.

`exit_category: "sandbox-exited"` is a **fallback label**, not an observation:

```ts
// packages/modal/src/terminal-disposition.ts
export function operationalDispositionForError(error: unknown): OperationalFailureCategory {
  return error instanceof OperationalDispositionError ? error.category : "sandbox-exited";
}
```

`PublicEvalDiagnosticsBuildError` (`public-worker.ts:69`) extends plain `Error`, so it takes that fallback. The real fault is that **building the public eval diagnostics document threw**; `model_work_started: true` then forced the collector to demand a document the failing build never wrote.

**What this retracts:** #323 (unbounded `git ls-files`) is *not* the root cause here — no `ENOBUFS` is on this trace. #323 is still a real, independently verified defect; it is simply unrelated to this failure.

**What still blocks this PR:** the underlying reason the diagnostics build failed. The eval exited 0 with `counts: {0,0,0}` and `usage: null`, so no eval row appears to have launched — that cause is not yet established. #319 has been retitled and rewritten around the corrected evidence.

#320 fixes the mislabel structurally (a fault the worker names itself can no longer claim `sandbox-exited`) and repairs the collector guard.

aviggiano:
## `origin/main` merged into this branch (`43a32b4`)

`main` had taken 16 merges since this branch last synced. The merge is done and pushed. Ten files were textually conflicted; two more were **semantically** conflicted while merging cleanly — those are the ones worth reading.

### The two semantic conflicts git did not see

**1. Foundry test-root model — `packages/runtime/src/templates/smithers/workflows/workflow.tsx`**

Both sides solve "which test root does this repository use", in incompatible ways:

| side | model |
| --- | --- |
| `main` | list-returning `invariantTestRoots(workspaceRoot)` → every root that exists |
| this branch | single-root `invariantTestRootName()` + `repositoryAwareWorkspaceOutputRoots()` |

`git merge-tree` reports no conflict, so a plain merge leaves **both** in the file.

**Kept: this branch's single-root model. Dropped: `invariantTestRoots`.** This is the opposite of what the merge brief guessed, so here is the evidence rather than the assumption:

- `invariantTestRoots` had **exactly one call site** on `main` — the generated-test reset inside `resetTaskArtifactsForRetry`. That call site no longer exists on this branch; it is now `taskTestOutputRelativeRoots` + `cleanWorkspaceOutputRootsForRetry`, cleaning the union of the roots *this* attempt pinned and the roots the *previous* attempt used, so a flipped anchor (`tests/` → `test/`) strands nothing.
- The two models disagree exactly on a repository that has grown both `test/` and `tests/`. The list model anchors output under **both**, which recreates the invented-empty-`test/recon`-tree defect **#212** closed. So the list model is the side that had to go.
- **#212's actual requirement — plural Foundry test roots preserved — is untouched.** That lives in `sourceCandidates = INVARIANT_TEST_ROOT_NAMES.flatMap(...)`, invariant *source* enumeration, which is still plural. `INVARIANT_TEST_ROOT_NAMES` survives and now serves both jobs: preference order for the single-root output pick, and the plural source sweep.
- Pinned by `runtime.test.ts:5984-5998` (asserts `invariantTestRootName` / `repositoryAwareWorkspaceOutputRoots` present, asserts the old shapes absent) and `runtime.test.ts:6037` (asserts the plural `sourceCandidates` sweep), plus `workspace-output-roots-repository-aware.test.ts` and the `REPOSITORY_AWARE_RETRY_HELPERS` list in `invariant-suite-handoff-durability.test.ts`.

**2. Findings normalization vs the verification marker — `packages/runtime/src/workflow-sync.ts`**

Same observation, opposite response, only one can hold:

- this branch wrapped `normalizeFindings` in a digest compare that raised `FindingsValidationError` when normalization moved a byte;
- `main` (#348 / PR #350) instead calls `refreshVerifiedArtifactDigest` to re-seal the attempt's marker.

**Kept `main`'s re-seal; dropped this branch's guard.** The guard *is* the bug #348 describes, raised one layer up: it fails the node closed as `artifact-contract`, which is the permanent `prepare:` failure that kept `dedupe-findings` red on all three targets. Nothing is silenced — a digest that moved is reported as `ARTIFACT_VERIFICATION_DIGEST_REFRESHED` carrying `previous_sha256`, and drift the runtime did **not** cause still fails closed through `reportVerifiedArtifactDigestDrift`. Rationale is inlined in the source at the call site.

### Per-file resolutions

| file | kept from `main` | kept from this branch |
| --- | --- | --- |
| `workflow.tsx` | `INVARIANT_TEST_ROOT_NAMES`, plural `sourceCandidates` (#212), bounded git enumerations (#323), suite budgets (#218) | single-root anchoring, `preparedWorkspaceOutputRoots` pinning, retry-union cleanup |
| `workflow-sync.ts` | #348 marker re-seal | atomic `commitEvents` unit — main's separate `preUnattributedWriteBudgetDiagnostic` checkpoint is subsumed by `preLocalCommitBudgetDiagnostic`, so terminal state cannot publish without the event that explains it |
| `node-provider.ts` | #280 marker republication (`markerRefreshed` / `isRefreshedVerificationMarker`), grafted onto this branch's function | the restructured publication function itself |
| `public-worker.ts` / `worker-diagnostics.ts` | #307 shared `sanitizeWorkerDiagnosticMessage`; this branch's `sanitizePublicDiagnosticMessage` deleted (zero remaining callers) | the caller-supplied redaction pass now runs through this branch's encoding-aware `redactSecretValueRepresentations` instead of a literal substring replace, so JSON-escaped / base64 / hex / percent-encoded credentials are caught too |
| `smithers.ts` | — | #324 salvage: RUN_EXISTS lineage adoption, no-op rewind removed, wrapper-less run-state *and* run-identity parsing |
| `runner.test.ts`, `public-worker.test.ts`, `terminal-recovery.integration.test.ts` | main's new assertions | this branch's call shapes; the duplicated evidence builder extracted into one shared `test/helpers/terminal-disposition-evidence.ts` |
| `invariant-suite-handoff-durability.test.ts` | #341 real-helper coverage | retry path exercised through `REPOSITORY_AWARE_RETRY_HELPERS`; the `invariantTestRoots` retry test is replaced, not deleted in spirit |
| `runtime.test.ts` | main's template assertions | the single-root assertions above |
| `CHANGELOG.md` | every `main` bullet | every branch bullet, plus two new bullets recording the reconciliations |

`artifact-gates.ts` (#292 empty-ledger justification + fan-in containment, #301 shared scan-probe validator) was untouched by this branch and came through `main` clean.

### Duplicate hunt

Method, as required: **set comparison of declared symbols across both parents and the merge**, per file, in both directions — every symbol each side added since the merge base, checked for presence in the result — not eyeballing.

Result: exactly **two** divergences surfaced (`preUnattributedWriteBudgetDiagnostic` from `main`, `verifiedFindingsDigest` from this branch), both explained above as deliberate subsumptions. Concept-level check:

- two root helpers → **one**
- two marker-refresh paths → **one** (`isRefreshedVerificationMarker`)
- two findings/marker responses → **one** (re-seal)
- two redaction passes → **one** (`sanitizeWorkerDiagnosticMessage` over `redactSecretValueRepresentations`)
- two terminal-disposition evidence builders → **one** shared helper
- scan-probe validator → **one** (`verifyInvariantProbePath`)
- invariant-suite budget → **one** (`assertInvariantSuiteSourceBudget`, one depth bound)
- recovery summary → **one** authority (#322, `modalRecoveryAnalysisSummary`)

Nothing was left unreconciled.

### Gates

```
corepack pnpm -r build          exit 0
corepack pnpm -r test           exit 0   (all 13 workspace projects; packages/cli 38/38 pass, 0 fail)
corepack pnpm -w format:check   exit 0   All matched files use Prettier code style!
corepack pnpm -w lint           exit 0   eslint, no output
```

No `MISSING_CACHE` in `packages/dashboard`; the references-sync workaround was not needed.

The push will kick a Modal smoke run. Per #321/PR #333 the non-default-branch smoke gate now forgives any operational failure category, so an infrastructure failure there should surface as a warning annotation rather than block this PR.

### Still required before merge

This PR is **not** ready to merge on the strength of this sync:

1. It still needs its **frozen candidate SHA** recorded.
2. It still needs **three independent reviews**.

Neither is satisfied by this merge commit, and the two semantic reconciliations above — dropping `invariantTestRoots` and dropping the findings digest guard — are exactly the calls a reviewer should re-derive rather than take on trust.

aviggiano:
## Duplicate-import slice pushed; DeepSeek V4 Flash publication outcome

### Slice landed on this branch

`bench/deepseek-v4-flash-results` advanced **`43a32b4..fe1376e`** (fast-forward). One line removed — the second `runningRecoverySandbox` specifier in the `packages/modal/test/runner.test.ts` import list, a leftover from an earlier `main` reconciliation, not from the PR author.

Red/green: restoring the file from `43a32b4` reproduces

```
test/runner.test.ts(70,3): error TS2300: Duplicate identifier 'runningRecoverySandbox'.
test/runner.test.ts(72,3): error TS2300: Duplicate identifier 'runningRecoverySandbox'.
```

With the fix, `corepack pnpm --filter @ultrafuzz/modal typecheck` exits 0 and the modal suite is **511 pass / 0 fail across 27 files**. The surviving specifier is the one in the list's sorted position; no test body changed.

As agreed when this issue was retitled to land in slices: **no rebase onto `main`**, and no attempt at the branch's other conflicts. The push triggered the standing smoke lane on this branch (`on: push: branches: ["**"]`) — repo behaviour, not an extra dispatch.

### Publication outcome: the Flash run scored, but did not publish

Prerequisite first: #410 merged to `main` as `70646d2`, which fixed the pre-flight guardrail that rejected any non-openai smoke manifest. Confirmed in production — the `Validate Modal benchmark launch guardrails` step succeeded on the run below, the exact step that failed in run 31261579436.

One dispatch from `main`: run [31264673583](https://github.com/monad-developers/ultrafuzz/actions/runs/31264673583), `benchmark_mode=smoke smoke_provider=deepseek smoke_model=deepseek-v4-flash smoke_reasoning=max`. Conclusion `success`, `launch` and `collect` both green.

The observation is real and complete:

- `scoring_ready: true` with **3 rows** — not the zero-row vacuous pass #367 addressed
- `planned 3, launched 3, launch_failed 0, workflow_succeeded 3, genuine_task_failure_rows 0, terminal_reports_present 3`
- `executed_case_count: 3`, `graded_case_count: 3`, bundle `status: succeeded`, judge `gpt-5.6-sol` / `xhigh`
- **no annotations on any job** — the non-default-branch soft-fail hatch did not fire
- scores: precision 1, recall 0.1345, F1 0.2369; 18 in review queue; duplicate rate 0.3611
- knobs: topology `max_attempts: 2` (all three groups in `benchmarks/smoke-benchmark.yml`), `node_timeout_seconds: 1800`, `control_timeout_seconds: 19800`, `strategy_loops: 1`

Notably the DeepSeek pair did **not** hit the `gpt-5.6-luna` control-pair wall from #370/#319.

**But `benchmarks/history.json` is unchanged at 108 observations — luna 99, kimi-k3 6, deepseek-v4-pro 3, flash still 0.**

Publication run [31267067524](https://github.com/monad-developers/ultrafuzz/actions/runs/31267067524) failed at `Download exact benchmark control metadata`:

```
Unable to download artifact(s): Artifact not found for name: modal-benchmark-launch-full-31264673583-1
```

The producer uploaded `modal-benchmark-launch-**smoke**-31264673583-1`. Root cause is `scripts/ci/qualify-modal-benchmark-publication.mjs:40`:

```js
benchmarkMode: eventName === "workflow_dispatch" ? "full" : "smoke",
```

The lane is inferred from the trigger event, assuming dispatch ⇒ full and push ⇒ smoke. #346 invalidated that when it made `benchmark_mode=smoke` a dispatch input. Every artifact name in the publish job derives from this value, so **any dispatched smoke run scores fine and then cannot be published.**

This belongs to this issue's hardening scope, and it is a second structural blocker in the same family as #410: #410 stopped non-openai smoke at pre-flight, this stops dispatched smoke at publication.

The Flash observation is still recoverable at **zero additional compute cost** — producer artifacts have 30-day retention, so fixing the mode derivation and re-running publication run 31267067524 lands the row without another benchmark run. The mode is not available from the `workflow_run` payload (it does not carry the producer's dispatch inputs); the derivable source is the producer's artifact list — pick the lane whose `modal-benchmark-launch-<mode>-<run>-<attempt>` artifact exists and require exactly one match.

I did not fix it in this session: it is a change to the trusted publication path and deserves its own reviewed slice rather than an unreviewed follow-on merge.

aviggiano:
Closed unmerged as the obsolete long-lived integration branch. Its remaining content shipped in reviewed slices: #420 (torn artifact-journal repair), #421 (Smithers executable capability), and #428 (sealed workflow controls, detached/snapshot recovery, and crash-safe adapter upgrades). Per #413, we are not carrying or merging the superseded 137-file branch.

PR #254, opened 2026-08-05
fix(runtime): avoid ignored roots in workspace handoff staging

aviggiano:
## Summary
- Avoid ignored workspace roots and invalid handoff staging inputs during runtime workspace handoff.
- Prefer descendant invariant-suite handoffs and allow directory invariant scan probes.
- Accept empty optional reference lists in artifact schemas.
- Filter invariant git enumeration roots and apply workspace patches by their base tree.
- Allow findings artifacts to preserve non-disposition category labels in `triage_classification`; final review prompts still constrain disposition choices.
- Exclude hidden generated invariant-suite paths before publication so runtime artifacts stay compatible with the artifact manifest safety contract.
- Normalize legacy object-shaped final-report coverage blocker summaries into canonical strings before report contract validation.
- Give setup scaffold nodes a longer default execution budget so setup and discovery blockers are not capped below strategy tasks.

## Validation
- `corepack pnpm --filter @ultrafuzz/artifacts... build`
- `corepack pnpm --filter @ultrafuzz/artifacts exec tsc -p tsconfig.test.json`
- `node packages/artifacts/dist-test/test/schema.test.js`
- `node packages/artifacts/dist-test/test/artifacts.test.js`
- `corepack pnpm --filter @ultrafuzz/runtime... build`
- `corepack pnpm --filter @ultrafuzz/runtime exec tsc -p tsconfig.test.json`
- `node packages/runtime/dist-test/test/generated-workflow-verifier.test.js`
- `node --input-type=module - <<SCRIPT ... SCRIPT` direct scaffold timeout regression (`setup_timeout_scaffold_ok`)
- `corepack pnpm --filter @ultrafuzz/cli... build`
- Direct privacy-redacted artifact-contract regression check: 4/4 previously failing shape instances accepted
- Direct privacy-redacted final-report coverage normalization regression: schema validation passes after canonicalization

PR #282, opened 2026-08-06
fix(runtime): stage workspace handoff without exclude pathspecs

### LINKED ISSUE #281: Workspace handoff staging aborts when an ignored runtime root exists: exclude pathspec misses the directory entry
## Summary

`captureWorkspaceTree` / `captureWorkspacePatch` abort with a non-zero `git add` when a gitignored runtime root such as `node_modules` exists at the workspace root. The exclude pathspecs cover the directory's **contents** but not the directory **entry**, so `git add -A -- .` still names the ignored path and git fails.

Confirmed live in the Aave v4 invariant-only run R44.

## Evidence

Run `aave-v4-v0012-main-issue272-r44-invariant-only` (main `77da78a`, image `im-01KZBJ546AZANSE7RNBY35A9DG`). Node `property-specification-a16z` failed at `2026-08-06T14:14Z` with:

```
Command failed: git add -A -- . :(exclude).ultrafuzz/** :(exclude).smithers/** :(exclude)node_modules/** :(exclude)artifacts/**
The following paths are ignored by one of your .gitignore files:
node_modules
hint: Use -f if you really want to add them.
```

Counts at that point: `13 succeeded / 1 failed / 1 running / 17 pending`. The other property-specification lenses on the same run were unaffected, so this reproduces only when that particular worktree actually contains `node_modules` — i.e. when the node's agent installed dependencies. That intermittency is why it survives most runs.

## Cause

`packages/runtime/src/workspace-handoff.ts:133-139`

```ts
function stageWorkspaceTree(workspaceRoot: string, index: string): void {
  runGit(
    workspaceRoot,
    ["add", "-A", "--", ".", ...WORKSPACE_RUNTIME_ROOTS.map((root) => `:(exclude)${root}/**`)],
    index
  );
}
```

with `WORKSPACE_RUNTIME_ROOTS = [".ultrafuzz", ".smithers", "node_modules", "artifacts"]` (`:21`).

`:(exclude)node_modules/**` excludes paths *under* `node_modules`, not `node_modules` itself. `git add -A -- .` therefore still matches the directory entry, and because it is ignored git exits non-zero rather than skipping it.

## Impact

The failure surfaces as a genuinely failed node, so it is retryable and less severe than #275 — but `property-specification-fanin` depends on all eight lens nodes, so a lens that fails deterministically in its own worktree blocks the invariant pipeline from ever reaching fan-in, invariant setup, Recon, findings or `report.md`.

## Expected fix

Exclude the runtime root entry as well as its contents (`:(exclude)<root>` alongside `:(exclude)<root>/**`), or stage from an explicit path list rather than a broad `add -A` pathspec.

Regression test: create a workspace containing an ignored `node_modules` directory at the root and assert `captureWorkspaceTree` / `captureWorkspacePatch` succeed rather than throwing.

## Relationship to #254

Draft PR #254 already proposes "stage workspace handoff patches from an explicit Git path list instead of broad add pathspecs" and "filter runtime roots before staging so ignored dependency/runtime directories do not abort patch capture", which would cover this. That PR bundles eight unrelated changes, so a narrow fix for this specific abort is likely easier to land and review than reviving the bundle.

aviggiano:
## Summary

- fixes #281
- workspace capture aborted whenever a gitignored runtime root (e.g. `node_modules`) existed in the worktree
- the `:(exclude)` pathspecs were the cause, not an incomplete form of it

## The failure

Found live in Aave v4 run R44 (`aave-v4-v0012-main-issue272-r44-invariant-only`, main `77da78a`). Node `property-specification-a16z` failed at `2026-08-06T14:14Z`:

```
Command failed: git add -A -- . :(exclude).ultrafuzz/** :(exclude).smithers/** :(exclude)node_modules/** :(exclude)artifacts/**
The following paths are ignored by one of your .gitignore files:
node_modules
hint: Use -f if you really want to add them.
```

## Why the obvious fix is wrong

My first attempt added `:(exclude)node_modules` alongside `:(exclude)node_modules/**`, on the theory that the glob covered the contents but not the directory entry. **The test stayed red.** So I measured git's actual behaviour instead of reasoning further:

| invocation | result |
|---|---|
| `git add -A -- .` | **OK** — ignored paths skipped silently |
| `git add -A -- . :(exclude)<root>/**` | FAIL |
| `git add -A -- . :(exclude)<root> :(exclude)<root>/**` | FAIL |
| `git add -A -- . :(exclude)<root>` | FAIL |
| `git add -A --ignore-errors -- . :(exclude)<root>` | FAIL |

**Any** negative pathspec makes `git add` report an ignored path as an error rather than skipping it. So the exclusions themselves had to go.

## The fix

Stage plainly with `git add -A -- .`, then `git reset --quiet <treeish> -- <runtime roots>` so those roots contribute nothing to the captured tree without ever being named to `add`.

Two details that matter:

- **Reset to the tree-ish, not `rm --cached`.** If a file under one of those roots is tracked, resetting restores its baseline content; dropping the index entry would record a *deletion* and silently corrupt the captured tree.
- **The target is threaded through, not assumed to be `HEAD`.** `captureWorkspaceTree` reads `HEAD` but `captureWorkspacePatch` reads the caller's `baselineTree`; resetting to `HEAD` in the latter would produce a patch against the wrong base.

## Impact

The node fails visibly, so it is retryable — less severe than #275's invisible `prepare:`-wrapper strand. But `property-specification-fanin` depends on all eight lens nodes, so a lens that fails deterministically in its own worktree blocks the invariant pipeline from reaching fan-in, invariant setup, Recon, findings or `report.md`.

It reproduces only when that node's agent actually installed dependencies in its worktree, which is why most nodes survive and this went unnoticed until R44.

## Validation

- new regression proven **red without the source change and green with it**; it plants all four runtime roots (`node_modules` ignored, the others not) plus a real setup change, then asserts the change is captured and no runtime root leaks into the patch
- `pnpm --filter @ultrafuzz/runtime test` — **278 pass / 0 fail**
- typecheck, eslint, prettier, `git diff --check` — clean

## Relationship to #254

Draft PR #254 proposes the same direction ("stage from an explicit Git path list instead of broad add pathspecs") bundled with eight unrelated changes. This is the narrow fix for the observed abort only; #254 can be rebased or closed on top of it.

<h3>Greptile Summary</h3>

The PR replaces exclusion-based workspace staging with byte-exact enumeration of tracked and non-ignored paths, omitting runtime roots before staging.
- Uses NUL-delimited literal pathspecs to preserve unusual filenames.
- Retries staging when an enumerated path disappears.
- Adds regression coverage for ignored and unreadable roots, tracked deletions, nested repositories, unusual filenames, and baseline preservation.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| packages/runtime/src/workspace-handoff.ts | Reworks workspace staging around byte-exact path enumeration, runtime-root filtering, and bounded retries for disappearing paths. |
| packages/runtime/test/workspace-handoff.test.ts | Adds focused regression tests covering ignored roots, concurrency, tracked-path semantics, nested repositories, and unusual filenames. |

</details>

```mermaid
flowchart TD
  A[Load baseline tree into temporary index] --> B[Enumerate tracked and non-ignored paths]
  B --> C[Exclude runtime roots and directory entries]
  C --> D{Any paths remain?}
  D -- No --> E[Write captured tree]
  D -- Yes --> F[Stage NUL-delimited literal paths]
  F --> G{Path vanished?}
  G -- Yes, attempts remain --> B
  G -- No --> E
  G -- Other failure --> H[Abort capture]
```
</details>

<sub>Reviews (5): Last reviewed commit: ["fix(runtime): retry staging when a liste..."](https://github.com/monad-developers/ultrafuzz/commit/ae2021807f159b7237c277912887f4f226998cb8) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=50830369)</sub>

**Context used:**


aviggiano:
## Redesigned after adversarial review — `793757d`

Two independent reviews. One recommended merge with a P2; the other found a P1 that invalidated the approach, and separately disproved the `core.excludesFile` follow-up I had started. Both were right, and the second attempt was wrong for a different reason than the first.

### What the first commit missed

The old `:(exclude)` pathspecs did **two** jobs. `:(exclude)artifacts/**` kept entries out of the index *and* kept `git add` from descending into the directory. `5f946b1` only accounted for the first.

So a bare `git add -A -- .` avoids the ignored-path error but **walks** the runtime roots. Measured by review on a 157 MB non-ignored `artifacts/` tree:

| | wall | `.git` growth |
|---|---|---|
| `git add -A -- .` | **8.62 s** | **+158 MB** unreachable objects |
| `:(exclude)artifacts/**` | 0.00 s | 0 |

And worse than the cost, it reintroduces the abort:

```
# unreadable file under a non-ignored runtime root
error: unable to index file 'artifacts/att/x.json'
fatal: adding files failed        exit=128
```

That matters because the runtime writes `artifacts/<attemptId>/` (`workflow.tsx:248`) and `.ultrafuzz/schemas/` (`workflow.tsx:371`) into the workspace, and a target repo like Aave v4 has no reason to gitignore either. Capture runs once per dependency (`workflow.tsx:470`) plus twice inside `applyWorkspacePatch`, so the cost is O(N) per fan-in node.

### Why `core.excludesFile` also failed

It is the **lowest-precedence** ignore source. A repository `.gitignore` containing `!artifacts/` re-admits the path, and the unreadable-file case still aborted. Verified in a scratch repo rather than assumed.

### The design now

Name only the *other* top-level entries as **positive** pathspecs. No negative pathspec means no ignored-path error; never naming the roots means never descending into them. The entry list is the union of worktree entries and `git ls-tree` entries so deleting a tracked top-level path is still captured, and the empty case is skipped because `git add -A --` with no pathspec stages everything.

This is what #254 proposed ("stage from an explicit Git path list instead of broad add pathspecs"); both reviewers converged on it independently.

### Verified against the cases that broke the earlier attempts

```
PASS  ignored node_modules does not abort            (the original #281 repro)
PASS  unreadable file under artifacts does not abort (broke commit 1)
PASS  gitignore negation of artifacts still excluded (broke core.excludesFile)
PASS  deleted tracked top-level path captured
PASS  only runtime roots present does not stage them (empty pathspec guard)
PASS  tracked artifacts file restored to baseline
```

Three are now pinned as regression tests; the unreadable-file one skips as root, which can read a 0000-mode file regardless.

### Also fixed from review

- My test comment stated the cause incorrectly — it claimed `/**` excluded contents but not the directory entry. Disproved: `:(exclude)<root>`, `:!<root>` and the `/**` form all fail identically. Corrected in the test and the source doc comment, because a maintainer trusting it could have "fixed" it by dropping `/**` and reintroduced #281.
- Added the missing tracked-runtime-root test. Swapping the reset for `git rm --cached` would have passed every other test while silently converting tracked files into spurious deletions in `result_tree`.
- Replaced a brittle regex assertion with a `git ls-tree` check on the captured tree.

Review also independently confirmed the risk I was most worried about is a non-issue: the `HEAD` vs `baselineTree` reset skew is byte-for-byte unchanged by this PR, with matching tree SHAs on both implementations.

### Validation

`pnpm --filter @ultrafuzz/runtime test` — **282 pass / 0 fail**. typecheck, eslint, prettier, `git diff --check` clean.

PR #284, opened 2026-08-06
fix(runtime): sanitize underscore-separated reference labels

### LINKED ISSUE #283: Lens reference-expectation sanitizer only recognizes hyphenated labels, so underscore catalog IDs fail the node
## Summary

`sanitizeLensReferenceExpectationAuthority` strips unauthorized reference-expectation IDs only when they match a hyphenated shape. An underscore-separated catalog-style ID such as `LEND_ACC_01` is not matched, so it survives sanitization and is then reported as an **error** by `verifyPropertyProvenanceArtifacts`, failing the node.

Confirmed live in Aave v4 run R44.

## Evidence

Run `aave-v4-v0012-main-issue272-r44-invariant-only`, image `im-01KZBJ546AZANSE7RNBY35A9DG` (main `77da78a`). Node `property-specification-0kn0t` failed at `2026-08-06T14:49Z`:

```
Property lens expectation "LEND_ACC_01" is not present in a supplied pinned-reference catalog;
Property lens expectation "LEND_ACC_02" is not present in a supplied pinned-reference catalog;
...
```

Run state at that point: `19 succeeded / 2 failed / 11 pending`, run `failed`, `property-specification-fanin` still pending.

## Cause

`packages/runtime/src/artifact-gates.ts:1767`

```ts
function isUnsupportedExternalReferenceLabel(expectationId: string): boolean {
  return /^[A-Z][A-Z0-9]*-\d+$/u.test(expectationId);
}
```

`LEND_ACC_01` contains underscores, so `[A-Z0-9]*` cannot span `_ACC_` and the pattern does not match. The sanitizer's guard at `:1590` therefore bails out for that property:

```ts
if (unauthorized.some((expectationId) => !isUnsupportedExternalReferenceLabel(expectationId))) return property;
```

and the ID reaches `verifyPropertyProvenanceArtifacts` (`:1558-1568`), which emits `PROPERTY_REFERENCE_EXPECTATION_UNAUTHORIZED` at severity `error` — failing the node.

So the outcome depends purely on which separator the model happened to use for an equally-fabricated identifier: `LEND-01` is quietly sanitized, `LEND_ACC_01` kills the node.

## Impact

`property-specification-fanin` depends on all eight lens nodes, so one lens failing this way blocks fan-in, invariant setup, Recon, findings and `report.md`. The node does fail visibly and is retryable, but a re-run of the same lens will usually produce the same identifier shape, so it is effectively deterministic.

R44 hit this alongside #281 (`property-specification-a16z`), leaving two blocked lenses.

## Expected fix

Widen the matcher to cover catalog-style identifiers with underscore separators, e.g. `/^[A-Z][A-Z0-9_]*[-_]\d+$/u`, so the same class of fabricated external label is sanitized regardless of separator.

Worth deciding explicitly rather than by regex accident: is an unauthorized reference expectation meant to be **sanitized** (the model cited something that does not exist, so drop the citation and continue) or to **fail the node** (the model fabricated a citation, so reject the artifact)? Right now it is both, split on a detail of the identifier's punctuation. Whichever is intended should apply uniformly.

Regression test: a lens whose `reference_expectations` contain an underscore-separated unauthorized ID must be sanitized the same way a hyphenated one is, and the node must not fail.

## Related

- #281 — the other blocker in the same R44 run.
- #275 / PR #279 — the sanitizer's marker-consistency fix; this is the same function, different branch.
- Draft PR #254 bundles "accept empty optional reference expectation lists as absent so valid implementation artifacts are not rejected", which is adjacent but not this.

aviggiano:
## Summary

- fixes #283
- an unauthorized reference-expectation ID was stripped if it used a hyphen and **failed the node** if it used an underscore
- found live in Aave v4 run R44

## The failure

`isUnsupportedExternalReferenceLabel` (`packages/runtime/src/artifact-gates.ts`) was:

```ts
return /^[A-Z][A-Z0-9]*-\d+$/u.test(expectationId);
```

`[A-Z0-9]*` cannot span an underscore, so `LEND_ACC_01` did not match. The sanitizer's guard then bailed out for that property, the ID survived into `verifyPropertyProvenanceArtifacts`, and it was reported as `PROPERTY_REFERENCE_EXPECTATION_UNAUTHORIZED` at severity `error` — failing the node.

`LEND-01` in the same position would have been stripped silently. So whether a lens node survived turned on the model's choice of punctuation.

R44 (`aave-v4-v0012-main-issue272-r44-invariant-only`, image `im-01KZBJ546AZANSE7RNBY35A9DG`), node `property-specification-0kn0t` at `2026-08-06T14:49Z`:

```
Property lens expectation "LEND_ACC_01" is not present in a supplied pinned-reference catalog;
Property lens expectation "LEND_ACC_02" is not present in a supplied pinned-reference catalog;
```

Run state: `19 succeeded / 2 failed / 11 pending`, run `failed`, `property-specification-fanin` never ran. Because fan-in depends on all eight lenses, one node failing this way blocks fan-in, invariant setup, Recon, findings and `report.md`.

## The fix

Allow `_` as well as `-` as the separator:

```ts
return /^[A-Z][A-Z0-9_]*[-_]\d+$/u.test(expectationId);
```

This is reached only for identifiers already absent from every supplied catalog, so no authorized expectation changes behaviour.

## Validation

- regression proven **red without the source change and green with it**; the red run reproduces R44's exact message for `LEND_ACC_01`
- asserts the underscore ID is sanitized alongside a hyphenated one, that no `PROPERTY_REFERENCE_EXPECTATION_UNAUTHORIZED` is raised, and that the node does not fail
- all four property-lens sanitizer tests pass

## Worth deciding separately

This makes the existing behaviour consistent rather than settling what it should be. An unauthorized reference expectation is currently **sanitized** in some cases and **fatal** in others, and that split was an accident of the regex. Whichever is intended — drop a citation that does not resolve, or reject an artifact that fabricates one — should apply uniformly. I have kept the existing sanitize-and-continue behaviour and only made it consistent, rather than changing the policy unreviewed.

## Related

- #281 / PR #282 — the other blocker in the same R44 run.
- #275 / PR #279 — same function, different branch (marker consistency).

<h3>Greptile Summary</h3>

The PR broadens unsupported external-reference label recognition so underscore-separated and multi-segment catalog labels are sanitized consistently before provenance verification.

- Supports mixed hyphen and underscore segments with a numeric suffix.
- Preserves fatal verification for forged namespaced citations.
- Adds regressions for underscore, multi-segment, and mixed forged-citation cases.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| packages/runtime/src/artifact-gates.ts | Broadens external catalog-label matching while retaining the existing authority boundary for namespaced citations. |
| packages/runtime/test/artifact-gates.test.ts | Adds focused regression coverage for multi-segment labels, underscore-separated labels, and intentionally rejected namespaced citations. |

</details>

```mermaid
flowchart LR
  A[Property lens artifact] --> B[Read supplied catalog IDs]
  B --> C{Unauthorized ID shape}
  C -->|External catalog label| D[Remove unauthorized reference]
  D --> E[Refresh verification digest]
  E --> F[Continue node successfully]
  C -->|Forged namespaced citation| G[Preserve artifact]
  G --> H[Provenance verifier reports unauthorized citation]
  H --> I[Fail node]
```
</details>

<sub>Reviews (3): Last reviewed commit: ["fix(runtime): match multi-segment catalo..."](https://github.com/monad-developers/ultrafuzz/commit/b393758ee7960b66aa2fa5a46d8619052f1ab8f3) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=50851734)</sub>

**Context used:**


aviggiano:
## Held as draft — a reviewer challenges the *policy*, not the change

Two reviews. One recommended merge with two P2s (one taken, one correctly rejected — see below). The second raised a P1 that I verified in part and that changes what this PR should be.

### What I verified

The reviewer's factual claims hold:

- The lens prompt tells the model to **copy identifiers verbatim**: "Populate `reference_expectations` only from exact identifiers present in the supplied pinned-reference artifacts... carry the supplied identifier unchanged" (`.ultrafuzz/prompts/properties/0kn0t-lens.md:31-34`).
- The 0kn0t pinned reference is a **Markdown corpus** (`0kn0t/properties_framework`: `README.md`, `01_PROPERTY_CLASSIFICATION_TAXONOMY.md`, …) — not a structured `ultrafuzz/reference-expectations@1` catalogue.
- R44's config supplies **no** expectation catalogue: keys are `app_name, benchmark_execution, braintrust, eval_reporting, ground_truth, image_name, models, node_timeout_seconds, run_id, schema_version, target`.
- `readLensSuppliedExpectationIds` returns an empty id set **with no diagnostic** when no dependency declares such a catalogue (`artifact-gates.ts:1810`, `if (expectationPaths.length === 0) continue;`).

So in the configuration these runs actually use, every citation any lens emits is "unauthorized", and the reviewer is right that R44's twelve `LEND_*` identifiers were not fabricated.

### Why I am not making the change the reviewer asks for

The reviewer's remedy is to treat "no catalogue supplied" as *unverifiable* — warn, keep the identifiers, do not rewrite the lens. I implemented exactly that. **It fails seven existing tests**, because the current behaviour is deliberate and pinned:

- `property lens gate strips unsupported external labels without weakening benchmark authority`
- `property lens authority sanitizer applies to custom logical lens IDs`
- and five others

Those tests construct nodes with **no** catalogue and assert that doc-shaped labels *are* stripped with a `PROPERTY_REFERENCE_EXPECTATION_SANITIZED` warning, while a namespaced mapping like `benchmark:unexpected` *does* fail the node. That is a coherent policy: `reference_expectations` is meant to carry verifiable provenance, so an identifier that cannot be authorised is dropped rather than presented as if it were, and a forged benchmark mapping fails loudly.

Under that policy this PR is aligned with intent and merely makes it **consistent** — hyphenated doc labels were already being stripped in precisely the same way; `LEND_ACC_01` died only because the matcher did not recognise underscores. The reviewer's objection is a critique of the pre-existing policy, not of this diff.

Flipping a deliberately-tested, security-relevant gate is not something I should do unilaterally on one reviewer's read, so I have left the policy alone, kept this PR to the consistency fix, and am recording the disagreement instead of resolving it by fiat.

### The decision that needs an owner

Given a lens that cites identifiers present in its supplied pinned-reference **documents** but absent from any structured catalogue, should the gate:

1. **strip them with a warning** (today's behaviour, which this PR makes consistent) — provenance stays verifiable-only, at the cost of discarding a mapping the prompt asked for; or
2. **keep them and mark them unverifiable** (the reviewer's proposal) — preserves the mapping, but `reference_expectations` then contains unauthorised content; or
3. **supply a real catalogue** for these runs, so the question does not arise — arguably the actual fix, since the prompt and the gate currently disagree about what "supplied" means.

Option 3 looks right to me and is out of scope here. Until it is decided, this stays a draft.

### Also from review

- **Taken:** the matcher was a point fix for a class bug — `LEND-ACC-01` and `CRYTIC-ERC4626-05` still failed. Now matches `-`/`_` segments in any combination with a numeric suffix; verified against 22 cases including the namespaced forms that must keep failing.
- **Rejected:** stripping per-identifier instead of per-property. That breaks `without weakening benchmark authority`, which asserts a lens mixing a label with a forged namespaced citation is left **byte-unchanged** — the artifact is the evidence the authority check exists to surface. The all-or-nothing rule now carries a comment saying so, plus a test pinning the preservation.
- Reviewer 1 confirmed no authorized identifier shape is endangered: real ones are colon-namespaced and never match the widened pattern.
- Reviewer 2 confirmed all twelve of R44's identifiers match, so this *would* unblock that node — and that the matcher is linear (no ReDoS).

39/39 tests in `artifact-gates.test.ts` pass at `8cc40ae`.

aviggiano:
## Merging with `launch` / `collect` not green — GitHub Actions infrastructure, stated plainly

For the record rather than glossed over: `launch` and `collect` never ran. Both jobs' **only** step is `Set up job`, and that is what failed, before checkout and before any code in this branch executes:

```
Failed to resolve action download info. Error: Internal Server Error
Failed to resolve action download info. Error: Service Unavailable
##[error]Failed to resolve action download info.
```

Verified via `gh api .../actions/jobs/<id>` — `steps[]` contains exactly one entry, `Set up job`, `conclusion: failure`. I re-ran them three times across ~45 minutes; the action-resolution service stayed unavailable and one attempt ended `cancelled` mid-outage.

`main` carries no branch protection, so there are no required contexts; `mergeStateStatus: UNSTABLE` reflects only these two jobs.

Everything substantive is green:

- `release-gates` ×2 — the build/test/lint gate
- Greptile Review, semgrep, Socket Security ×2
- locally: 39/39 in `artifact-gates.test.ts`, 280 pass / 0 fail in the full runtime suite, typecheck / eslint / prettier / `git diff --check` clean
- two independent review rounds, every finding either applied or rejected with reasoning recorded above

I held #279 and #282 to a fully CLEAN status before merging. This one is a deliberate exception because the two outstanding jobs cannot start for reasons outside the repository, and they are not required. If someone wants them green before this lands, revert and re-run once GitHub recovers.

Issue #122, opened 2026-07-22
Preserve verified artifacts before Smithers worktree reaping


aviggiano:
## Problem

Verified task artifacts can be deleted before Ultrafuzz publishes them from a Smithers worktree into the canonical run artifact directory.

The current lifecycle is:

1. The generated prompt directs the agent to a task-owned `artifacts/<attempt>` mirror inside its Smithers worktree.
2. The artifact-aware agent wrapper and verifier validate the files in that mirror before reporting task success.
3. Ultrafuzz runs Smithers detached and an out-of-process polling synchronizer later copies mirror files into canonical storage.
4. With `keep_workspaces=false`, Smithers 0.28 reaps worktrees when a workflow finishes successfully.
5. Smithers' Git cleanliness check uses ordinary `git status --porcelain`, which does not report ignored files. In a repository that ignores `artifacts/`, a worktree containing only the required outputs can therefore look clean and be deleted.
6. If the workflow finishes between synchronization polls, the next Ultrafuzz artifact gate sees required outputs as missing even though the in-workflow verifier already accepted them.

This is a publication/reaping race across the Ultrafuzz-Smithers boundary. It is not a Zod-output failure: the Smithers task schema covers the summary object, while the filesystem contract is validated separately before Smithers parses or repairs that summary.

## Secondary recovery livelock

The resulting failed run can also enter a permanent Modal relaunch loop:

1. Runtime finalization records an `artifact-contract` failure, but missing required artifacts do not receive the narrow `task-output-validation-failure` terminal marker.
2. Terminal-disposition inspection reports an operational or incomplete outcome.
3. The Modal worker sees the linked durable run is already `failed`, treats it as terminal, and does not invoke durable resume.
4. Eval-record repair refuses to finalize the operational/incomplete outcome.
5. That refusal becomes `sandbox-exited`; because model work already started, launch-state classification maps it to `resume-required` with a relaunch action.
6. Every replacement worker follows the same path without running or resetting a node, so the durable checkpoint never advances.

Strict artifact validation is correct and must remain fail-closed. The controller must not advertise a state as resumable when the worker has no transition that can resume it.

This recovery portion is related to #111, but is a narrower deterministic zero-progress cycle.

## Why current tests do not catch it

The generated-workflow tests inspect source structure, and reconciliation tests exercise a mirror that still exists. They do not run a real successful Smithers worktree lifecycle with ignored artifact files, automatic reaping, and delayed external synchronization.

[...]

PR #286, opened 2026-08-06
Modal launch budget: stop charging post-model sandbox exits against the pre-model flake budget

### LINKED ISSUE #267: R40 Aave v4 run exhausted Modal recovery after sandbox exits
## Summary

Fresh R40 Aave v4 invariant-only run from merged main exhausted Modal recovery before reaching property artifacts.

- Run ID: `aave-v4-v0012-main-issue265-r40-invariant-only`
- Source revision: `116be7a287e48e7bdb7d3d44d728e1fb50a674e3`
- Image: `ultrafuzz-security-runner:v0.0.12-r40-main-116be7a-invariant-only`
- Image id: `im-01KZAR5JEB7CB0RCK5YTRPQPZ4`
- Target: Aave v4 `6959e3219b5506bf2acae18551cbb2a68a5b8fba`
- Mode: invariant-only, n=1, GPT-5.5 xhigh, Modal

Observed sequence:

1. Attempt 1 launched at `2026-08-06T05:19:22Z`, reached model work, then exited operational `dependency-unreachable` / `resume-required` after `9 succeeded / 0 failed / 23 remaining`.
2. Attempt 2 resumed from the same durable volume at `2026-08-06T05:23:54Z`, exited before model work with retryable `sandbox-exited`.
3. Attempt 3 pre-model retry launched at `2026-08-06T05:24:59Z`, exited before model work with `sandbox-exited`; recovery lifecycle marked `recovery-budget-exhausted` and runner action became `none`.

Sanitized evidence is preserved locally:

- Tracking ledger: `/home/ubuntu/ultrafuzz-private/aave-v4-v0012-n1-20260802/r40/r40-property-tracking.md`
- Attempt 1 terminal: `/home/ubuntu/ultrafuzz-private/aave-v4-v0012-n1-20260802/r40/r40-terminal-20260806T052158Z`
- Attempt 2 terminal: `/home/ubuntu/ultrafuzz-private/aave-v4-v0012-n1-20260802/r40/r40-terminal-20260806T052402Z`
- Attempt 3 terminal: `/home/ubuntu/ultrafuzz-private/aave-v4-v0012-n1-20260802/r40/r40-terminal-20260806T052506Z`

## Notes

No task-level failure was reported, no property JSON/Markdown or report artifacts were collectable, and no code root cause is confirmed yet. This issue tracks the operational failure so future investigation can decide whether the recovery budget/classification should change or whether this was external Modal instability.

aviggiano:
## Root cause

`MODAL_PRE_MODEL_RETRY_LIMIT` (`packages/modal/src/defaults.ts`) exists to bound sandbox launch flakes that happen **before** model work. It was charged against `record.attempt`, which `reserveModalLaunchAttempt` increments for *every* attempt a model makes in the current generation — including attempts that ran hours of real model work and then exited asking for a durable resume.

Two consequences, both visible in R40:

1. From the third attempt onward, any sandbox dying before `launched_at` was classified `permanent-operational-failure` with `action: "none"`, and `launchOrResumeModel` threw `Modal runner cannot relaunch …`, abandoning the whole run. Longer runs were strictly more fragile.
2. The lifecycle was labelled `recovery-budget-exhausted`, pointing an operator at the outer recovery budget rather than the pre-model launch budget that actually tripped.

A correction to the original triage worth recording: `attempt` is **not** a lifetime counter. `createModalLaunchState` starts `launches: []`, so `attempt` restarts at 1 on each generation rollover; it is monotonic per slug *within* a generation. That does not weaken the bug — R40's three attempts were all in one generation — but it is why the new streak is scoped to the record's generation.

## The fix

`modalPreModelAttempt(state, record)` derives the **consecutive pre-model streak**: `record.attempt` minus the highest earlier attempt in the same generation whose durable recovery-lifecycle record observed `model_work_started === true`. It reads only data already in `state.recovery_lifecycle` — no schema change, no migration, no new persisted field. Every runner call site charges the budget and the exponential backoff against that streak.

The streak resets **only** on a definite `model_work_started === true` observation. An `"unknown"` observation does not reset it. That is deliberate and fail-closed: resetting on `launched_at` instead would be a livelock, because a record that launched but whose worker terminally reported `transient-operational-failure` with `model_work_started: false` would reset the streak forever.

Because the reset depends on a durable observation, **every writer that closes an attempt lifecycle has to record what it actually saw.** `launchOrResumeModel` did. The unattended overseer did not: it advances an attempt through `reserveModalLaunchAttempt` alone, which force-closes the outgoing attempt's still-active lifecycle with a hardcoded `"unknown"` — so on that path the streak equalled `record.attempt` and the fix was inert for exactly the long, overseer-driven runs the issue was filed against. The observation was already in hand (`overseeModalBenchmarkOnce` reads the outgoing attempt's worker status off the volume two frames above the reservation); it is now threaded through `launchModalRecoveryWorker` into `reserveModalLaunchAttempt`'s new `observedModelWorkStarted`.

`recovery-budget-exhausted` is now emitted only when the failure is transient, the streak is genuinely exhausted, **and** model work is known not to have started. The abandonment message names the budget position only when the budget is what stopped the run: `incompatible-checkpoint`, a worker-reported `permanent-operational-failure` and `resume-required` under the stop-on-post-model public-benchmark policy all reach `action: "none"` without spending a retry. `modalPreModelBudgetExhausted` is the single predicate behind both the label and the message, so the two cannot drift.

`ModalPreModelAttempt` is a branded type produced only by `modalPreModelAttempt`. `preModelAttempt: record.attempt` is now a compile error at every present and future call site — a rename alone only forces a one-time review. There is no cast and no escape hatch: the tests build streaks through the real function.

Operator diagnostics: `modalBenchmarkStatus` rows carry `pre_model_attempt`, which is what was missing when R40 was triaged.

## THE OUTER NO-PROGRESS BOUND IS PRESERVED

`packages/modal/src/recovery.ts` has **zero diff against main** (`git diff --name-only origin/main...HEAD -- packages/modal/src/recovery.ts` returns 0 lines). `MODAL_RECOVERY_MAX_NO_PROGRESS_GENERATIONS` still terminates a row at `recovery.ts:336-340` with `no-progress-budget-exhausted`. **The issue #122 livelock stays closed.**

This does not make the runner retry forever, and it does not assume R40 was a false positive. Three consecutive pre-model exits still abandon the model. A deterministic launch failure still terminates — it now needs three *pre-model* failures rather than three *total* attempts.

## Regression tests and red/green evidence

New tests in `packages/modal/test/launch-state.test.ts`:

- `spends the pre-model launch budget on consecutive pre-model attempts only`
- `never labels an attempt that started model work as pre-model budget exhaustion`
- `resets the pre-model streak only when a relaunch records the model work it observed`
- `scopes the pre-model streak to the generation that is spending it`
- `names the pre-model budget only on the abandonment it actually caused`

**Red-first evidence for the original bug.** A throwaway probe reconstructed the R40 sequence against the pre-fix source using only the pre-fix public API, so the failures are behavioural diffs rather than missing-export errors:

- misleading label: expected `terminal_reason: "operational-failure"`, received `"recovery-budget-exhausted"`.
- the actual abandonment, calling `classifyModalRunnerStatus` with exactly what the runner passed pre-fix: expected `action: "relaunch" / transient-operational-failure / retryable: true`, received `action: "none" / permanent-operational-failure / retryable: false` at attempt 3. This is the "Modal runner cannot relaunch" whole-run-abandoned path.
- a launched attempt mislabelled as pre-model budget exhaustion: expected `"operational-failure"`, received `"recovery-budget-exhausted"`.

Stashing only the two source files and running the committed tests against the pre-fix tree gave `4 failed | 15 passed`.

**Mutation evidence for each new assertion.** Each safety-critical clause was individually mutated and the intended test — and only that test — failed:

| Mutation | Test that caught it |
| --- | --- |
| `observedModelWorkStarted` dropped (reserve hardcodes `"unknown"`) | `resets the pre-model streak only when a relaunch records the model work it observed` |
| `model_work_started === true` relaxed to `!== false` | same |
| per-generation clause deleted | `scopes the pre-model streak to the generation that is spending it` |
| budget suffix made unconditional | `names the pre-model budget only on the abandonment it actually caused` |
| streak collapsed back to `record.attempt` (the original bug) | `spends the pre-model launch budget on consecutive pre-model attempts only` |

## Review findings addressed

1. **major — the fix was a no-op on the overseer/recovery-supervised relaunch path** (raised independently by two reviewers, both verified empirically against the built dist). Fixed by threading the already-computed worker status into the reservation's implicit lifecycle close. I chose to extend `reserveModalLaunchAttempt` rather than add an explicit `finishActiveModalRecoveryLifecycle` call in the overseer: the implicit close is the exact writer of the `"unknown"`, so giving it the observation fixes every caller at once and leaves terminal-reason accounting untouched.
2. **minor — `(pre-model attempt N of 3)` appended to abandonments the budget did not cause** (all three reviewers). Fixed via `modalPreModelBudgetExhausted`, which also de-duplicates the `recoveryBudgetExhausted` expression.
3. **minor — the fail-closed `=== true` clause and the generation scope had no coverage.** Both now pinned, and both verified by mutation.
4. **minor — the "unrepresentable at the type level" claim was false; `preModelAttempt: number` accepts `record.attempt` forever.** The claim is now true: `ModalPreModelAttempt` is branded.

Deliberately not changed, with reasoning:

- **The internal inconsistency at the relaunch-path lifecycle close** (the classifier concludes `model_work_started: true` from `launched_at` alone while the durable record says `"unknown"`). The durable field must stay a record of what was *observed*, not of what was *inferred* — that honesty is what keeps the streak fail-closed. It also costs nothing here: a record with `launched_at` set classifies as `resume-required` *before* the budget check, so the pre-model budget never bounded that shape either way. Now carries a comment saying so.
- **`pre_model_attempt` emitted on `modalBenchmarkStatus` rows for launched attempts.** The value is well-defined on every row and is the operator diagnostic the issue asked for; suppressing it conditionally would make the field's absence ambiguous.

## Residual risk

- **Migrated legacy launch states** (v1/v2 → v3) synthesise lifecycle records with `model_work_started: "unknown"`, so the streak never resets for them. Fail-closed and regression-neutral, but a run resumed from a pre-v3 checkpoint keeps today's fragility.
- **An attempt killed before its lifecycle is closed** by a path with no observation still records `"unknown"` and does not reset the streak. Chosen over resetting on ambiguity.
- **Strict relaxation within a generation**: a model can now consume up to 3 pre-model relaunches after *each* attempt that reached model work, instead of 3 in total. Bounded per generation by the unchanged outer no-progress budget, but wall-clock and Modal spend for a pathological run are higher than before.
- **Source-level breaking change** for out-of-tree callers of `classifyModalRunnerStatus` / `modalRecoveryTerminalReasonForWorkerStatus`: the field is renamed *and* branded. Both are used only inside `packages/modal`.
- **`modalBenchmarkStatus` rows gained a `pre_model_attempt` key.** External consumers doing strict schema validation on that JSON would see a new field.
- `attemptProvenance` was duplicated in `runner.ts` and `launch-state.ts`, and the runner copy silently dropped `post_model_recovery`. They are now one exported `modalAttemptProvenance`, so a generation rollover preserves that field in `attempt_history` where it previously lost it.

## Validation

- `pnpm -w run ci` → **exit 0** (`format:check` "All matched files use Prettier code style!", `lint`, `build` across all 13 projects, `validate:release`). 167G free on `/` before the gate.
- `pnpm -w test` → **exit 0**, all 13 packages green.
- `@ultrafuzz/modal`: 25 files, 310/310. Blast radius on the other packages referencing the `recovery-budget-exhausted` enum: `@ultrafuzz/evals` 165/165, `@ultrafuzz/artifacts` 68/68.
- `pnpm -w typecheck` → exit 0.

Fixes #267

## Post-review verification note (added by the landing pass)

The overseer-path major was **independently re-verified in the pushed head** and is genuinely fixed: `overseeModalBenchmarkOnce` computes the outgoing attempt's `workerStatus` off the volume and threads it through `launchModalRecoveryWorker` into `reserveModalLaunchAttempt`'s new `observedModelWorkStarted`, so the implicit lifecycle close records a real observation instead of the hardcoded `"unknown"`.

One gap in the *protection* around that fix, found by mutation and worth stating plainly: **deleting the `observedModelWorkStarted:` argument at the `launchModalRecoveryWorker` reservation site keeps all 310 `@ultrafuzz/modal` tests green and `pnpm -w typecheck` clean.** The field is optional, so omitting it compiles and silently restores exactly the #267 behaviour on the overseer path. The `ModalPreModelAttempt` brand guards the *classifier* call sites, not this one; the committed test `resets the pre-model streak only when a relaunch records the model work it observed` pins the `launch-state.ts` half (the reservation honouring the observation) but not the `runner.ts` half (the overseer supplying it), because nothing in `packages/modal/test` mocks the Modal client.

So the behaviour is correct and the bound is preserved, but a future refactor of `launchModalRecoveryWorker` can silently reintroduce #267 without turning any gate red. The cheap follow-up is to make `observedModelWorkStarted` a required field of `reserveModalLaunchAttempt` — then omission is a compile error at every call site, the same technique already used here for `preModelAttempt` — or to extract the `workerStatus?.model_work_started ?? (launched_at === undefined ? false : "unknown")` expression (duplicated at three sites) into one exported helper and test it directly. Not done in this PR to keep an already-reviewed, CI-green change stable.

CI note for the reviewer: the earlier `release-gates` failure on this PR was a GitHub Actions service outage (`Failed to resolve action download info. Error: Service Unavailable` during "Set up job"), which also failed `main`'s CI in the same window, not a defect in this branch. It was re-run and **passes**.

<h3>Greptile Summary</h3>

The PR changes Modal’s pre-model retry accounting to use a generation-scoped consecutive pre-model streak, preserving retries after confirmed model work while improving lifecycle labels and abandonment diagnostics.
- Threads observed model-work status through overseer-driven reservations so durable lifecycle closure can reset the streak.
- Uses a branded pre-model attempt value across classification, backoff, status, collection, and terminal-reason paths.
- Preserves recovery policy in attempt history and adds focused retry-budget regression coverage.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| packages/modal/src/launch-state.ts | Adds generation-scoped pre-model streak derivation, branded classifier inputs, cause-specific budget diagnostics, and complete attempt provenance. |
| packages/modal/src/runner.ts | Threads observed model-work state through overseer relaunches and consistently applies the derived streak across launch, status, collection, and diagnostics. |
| packages/modal/test/launch-state.test.ts | Covers consecutive-budget accounting, model-work resets, generation scoping, and cause-specific abandonment messages. |
| packages/modal/test/terminal-recovery.integration.test.ts | Updates terminal recovery integration coverage to use the derived pre-model attempt contract. |
| CHANGELOG.md | Documents the corrected Modal pre-model launch-budget behavior, though the unreleased section contains two overlapping entries. |

</details>

```mermaid
sequenceDiagram
  participant O as Overseer / Launcher
  participant V as Modal Volume
  participant S as Launch State
  participant C as Status Classifier
  O->>V: Read outgoing worker status
  V-->>O: model_work_started observation
  O->>S: Close outgoing lifecycle with observation
  O->>S: Reserve next attempt
  S->>S: Derive generation-scoped pre-model streak
  S->>C: Pass branded preModelAttempt
  C-->>O: Relaunch or abandon
```
</details>

<sub>Reviews (2): Last reviewed commit: ["fix(modal): reset the pre-model streak o..."](https://github.com/monad-developers/ultrafuzz/commit/ab7f0092b030ad6793766de2edcbef96ccfb6ca4) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=50871064)</sub>

**Context used:**


aviggiano:
## Two independent reviews, prompted by a live incident — and they refuted the reason I was looking at this PR

I came to #286 because two Aave v4 benchmark runs died at `stateful-invariant-setup` with `exit_category: "sandbox-exited"`, `model_work_started: true`, R45 at attempt 6 and R46 at attempt 3. The title reads as though that is exactly what this fixes. Both reviewers, working independently, found that it is not.

**The title describes something that does not happen.** In `classifyModalRunnerStatus` (`launch-state.ts:989-1001`, unchanged by this PR) every post-model condition is checked *before* the budget gate:

```ts
if (worker?.model_work_started === true || category in {model-work, post-processing, resume-required})
  return status("resume-required", postModelRecovery === "stop" ? "none" : "relaunch", ...)
...
if (input.attempt >= MODAL_PRE_MODEL_RETRY_LIMIT) { ... }   // only reachable below all of the above
```

So a post-model sandbox exit was **never** charged against the pre-model budget. One reviewer ran the classifier directly with R46's exact inputs at attempts 3 and 6: both return `{category: "resume-required", action: "relaunch"}`. R45's launch state confirms it empirically — generation 1, **attempt 6**, `phase: "launched"`, having sailed past attempt 3 on `main`.

R45 actually died from `no-progress-budget-exhausted` (`recovery.ts:336-340`), and `recovery.ts` has **zero diff** in this PR.

**The real bug here is narrower and genuine**, and worth landing: a genuinely *pre-model* attempt is charged against `record.attempt`, a counter inflated by earlier *post-model* attempts, so from attempt 3 a pre-model flake abandons a resumable run. `modalPreModelAttempt` fixes that correctly — generation-scoped, excludes the active lifecycle, fails closed on `"unknown"`. Four mutations tested, each caught by its intended test. Merges cleanly onto current `main` (verified with `git merge`), typecheck clean, 293/293 in `packages/modal` excluding the pre-existing environmental `ci-config.test.ts` failures.

### Suggested before landing

**Rename the PR.** Something like *"charge the pre-model retry budget to consecutive pre-model attempts"*. The current title is what generated the bad merge pressure I arrived with, and it will do it again.

**Close the self-disclosed gap.** Deleting `observedModelWorkStarted:` at `runner.ts:1771` leaves all tests green and typecheck clean while silently restoring #267 on the overseer path — the half that matters for unattended runs. Both reviewers independently proposed the same one-line fix: make `observedModelWorkStarted` **required** on `reserveModalLaunchAttempt` (`launch-state.ts:745`), the same technique this PR already uses for the `ModalPreModelAttempt` brand, so omission becomes a compile error at all three call sites.

**Note two observability shifts** for whoever reads the histograms: `collect` will label some cases `operational-failure` that were `recovery-budget-exhausted`, and `model_work_generations` will shift as the overseer records real values instead of `"unknown"`. Neither is behavioural; counts are not comparable across the merge.

### I am not merging it while runs are in flight

Not because it is dangerous — no mechanism was found by which it harms a live run, and the durable format is forward- and backward-compatible both ways — but because the expected value for those runs is exactly zero and the cost is not. Merging means rebuilding and restarting the overseer, and a dead overseer is precisely what stranded R45 (#295). Taking an availability risk on that component to install a change that is provably inert for this failure shape is a bad trade.

One reviewer also pointed out that R46's already-written lifecycle records all say `"unknown"`, so `modalPreModelAttempt` computes a streak identical to `record.attempt` on its existing state — the PR is bit-for-bit identical to `main` for R46 regardless.

This is the owner's call to land whenever suits; I am recording the analysis rather than acting on it. The `stateful-invariant-setup` sandbox exits remain unexplained and are the thing that actually needs the next hour.

PR #290, opened 2026-08-06
fix(runtime): accept directory invariant scan probes

### LINKED ISSUE #289: Invariant scan probes are validated as regular files, so a directory probe fails project-discovery
## Summary

`readInvariantSourceSnapshot` enforces regular-file-ness for **both** of its labels, but a *scan probe* records where the agent looked, not a file whose bytes are snapshotted. When project-discovery records a directory such as `tests` as a scan probe, the gate rejects it and the node fails.

Confirmed live in Aave v4 run R45.

## Evidence

Run `aave-v4-v0012-main-issue281-r45-invariant-only`, image `im-01KZBYNVDXXRJ7MNY4W6GMKMM3` (main `fdce587`). Node `project-discovery` failed at `2026-08-06T17:10Z`:

```
artifact-contract failure: invariant scan probe tests is unavailable: scan probe is not a regular file
```

Run state: `9 succeeded / 1 failed / 22 pending`, run `failed`. `tests/` is a directory in the Aave v4 target.

## Cause

`packages/runtime/src/templates/smithers/workflows/workflow.tsx:3327-3343`

```ts
function readInvariantSourceSnapshot(
  workspaceRoot: string,
  relativePath: string,
  label: "scan probe" | "invariant source"
): { bytes: Buffer; content: string } {
  ...
  sourcePath = resolveRegularArtifactFile(workspaceRoot, sourceCandidate, `${label} is not a regular file`);
  ...
}
```

Both labels take the same path, so a scan probe must be a regular file, must decode as UTF-8, and must match the pinned commit byte-for-byte. Those constraints are right for an **invariant source** — its bytes become evidence. They are wrong for a **scan probe**, which is a location the agent searched; a directory like `tests/`, `src/` or `contracts/` is a perfectly reasonable thing to have scanned.

The distinction is real but the implementation collapses it, so whether project-discovery survives depends on whether the model happened to name only files. R44's project-discovery recorded file probes and passed on the immediately preceding image; R45 named a directory and died. That is model-output variance deciding a plumbing outcome.

## Impact

`project-discovery` is the first agentic node and everything depends on it, so this blocks the entire invariant pipeline before any property lens runs. The node fails visibly and is retryable, and a retry may well pick different probes — so it is intermittent rather than deterministic, which makes it easy to miss and hard to attribute.

## Expected fix

Split the two labels:

- **scan probe** — accept a directory or a regular file. Record existence (and, for a file, the snapshot) as search evidence. Do not require UTF-8 decoding or a pinned-bytes match for a directory.
- **invariant source** — unchanged: regular file, UTF-8, pinned and unmodified.

Regression test: a ledger whose scan probes include a directory must pass project-discovery's evidence gate, while an invariant source pointing at a directory must still fail.

## Related

Draft PR #254 already proposes "allow canonical directory scan probes as ledger search evidence while preserving strict file snapshots for source probes", bundled with eight unrelated changes. This is the third defect from that bundle to surface in a live run, after #281 and (adjacently) #283 — the bundle is worth unpicking into landable pieces.

aviggiano:
## Summary

- fixes #289
- a *scan probe* records where the agent searched; an *invariant source* records bytes that become evidence. Both went through the same validation, so a directory probe failed `project-discovery`
- found live in Aave v4 run R45

## The failure

Run `aave-v4-v0012-main-issue281-r45-invariant-only`, image `im-01KZBYNVDXXRJ7MNY4W6GMKMM3` (main `fdce587`). Node `project-discovery` failed at `2026-08-06T17:10Z`:

```
artifact-contract failure: invariant scan probe tests is unavailable: scan probe is not a regular file
```

`tests/` is an ordinary directory in the Aave v4 target. `project-discovery` is the first agentic node, so this stopped the whole invariant pipeline before a single property lens ran — run state `9 succeeded / 1 failed / 22 pending`.

## Why this is an extension of existing behaviour, not a new rule

`verifyInvariantLedgerSourceEvidence` already accepted a directory probe — for the workspace root, with this comment:

```ts
// A repository-wide probe names the workspace directory itself. It is
// valid evidence, but cannot be snapshotted as a regular UTF-8 file.
continue;
```

A subdirectory probe is precisely the same case; it just fell through to the file-only path in `readInvariantSourceSnapshot`, which additionally demands UTF-8 decodability and a byte-identical match against the pinned commit. Those constraints are right for an invariant source and meaningless for a directory. Any directory probe now takes the route the root probe already took.

## Fixed in both implementations

The rule exists twice and R45 hit the second one:

- `packages/runtime/src/artifact-gates.ts` — the runtime gate, which emits `INVARIANT_LEDGER_PROBE_PATH_INVALID`
- `packages/runtime/src/templates/smithers/workflows/workflow.tsx` — the generated workflow's `verifyInvariantLedgerSourceEvidence`, which produced R45's actual error

Fixing only one would still fail the node. The runtime diagnostic message also no longer claims a constraint it does not enforce ("must be a regular file" → "must be a regular file or directory").

## Invariant sources deliberately unchanged

Their bytes *are* the evidence, so a directory there must still fail. A second regression pins that, so this cannot be mistaken for a blanket relaxation.

## Intermittent, which is worth knowing

This depends purely on whether the model names a directory among its probes. R44's `project-discovery` recorded only file probes and passed on the immediately preceding image. So it is easy to miss and hard to attribute — R45 looked like a fresh regression until the error was read.

## Validation

- new regression proven **red without the source change and green with it**; the red run reproduces R45's message verbatim
- `packages/runtime/test/artifact-gates.test.ts` — 41 pass / 0 fail
- full runtime suite running

## Related

Draft PR #254 proposes "allow canonical directory scan probes as ledger search evidence while preserving strict file snapshots for source probes", bundled with eight unrelated changes. This is the **third** defect from that bundle to surface in a live run, after #281 and adjacently #283. The bundle is worth unpicking into landable pieces rather than left to predict failures nobody acts on.

<h3>Greptile Summary</h3>

The PR aligns runtime and generated-workflow validation so non-symlink directory scan probes are accepted while invariant sources remain restricted to regular files.
- Adds the directory-probe allowance to both validation implementations.
- Preserves rejection of symlinked directory probes and directory invariant sources.
- Adds focused regression coverage for the runtime gate and generated workflow verifier.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| packages/runtime/src/artifact-gates.ts | Accepts canonical non-symlink directories as scan probes after existing containment checks while retaining strict validation for other path types. |
| packages/runtime/src/templates/smithers/workflows/workflow.tsx | Mirrors the runtime behavior in generated workflows by skipping file snapshots for validated directory probes. |
| packages/runtime/test/artifact-gates.test.ts | Adds regressions covering accepted directory probes, rejected symlinked directories, and rejected directory invariant sources. |
| packages/runtime/test/generated-workflow-verifier.test.ts | Exercises the generated verifier directly and confirms only ordinary directory probes bypass regular-file snapshotting. |

</details>

```mermaid
flowchart TD
    P[Invariant ledger path] --> T{Evidence type}
    T -->|Scan probe| S{Path kind}
    S -->|Non-symlink directory| A[Accept without snapshot]
    S -->|Regular file| F[Validate and snapshot]
    S -->|Symlink or invalid type| R[Reject]
    T -->|Invariant source| F
    F -->|Snapshot succeeds| V[Accept evidence]
    F -->|Not a regular file| R
```
</details>

<sub>Reviews (2): Last reviewed commit: ["test(runtime): pin the directory scan-pr..."](https://github.com/monad-developers/ultrafuzz/commit/c8923fb4274b3aaa621178769720caaa012f87ca) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=50884662)</sub>

**Context used:**


aviggiano:
## Two independent reviews, both actionable findings fixed

Both reviewers recommended merge and neither found a security regression. Full detail on what was measured, since the interesting part is what they found *wrong with the tests* rather than the change.

### Verified by both reviewers, independently

- **No new capability.** Nothing reads `probe.result` — every consumer is a presence check. The pre-existing ENOENT allowance already accepts a probe naming a *fabricated* path with arbitrary result text; one reviewer confirmed `source_path: "tests/totally/made/up.sol"` passes with zero diagnostics on `main`. Accepting a real directory is a subset of that. The old rule was self-inconsistent: `tests/nope` passed and `tests` failed, which is precisely R45.
- **No escape.** 13 adversarial ledgers, all rejected: symlink-to-dir inside the workspace, symlink to `/etc`, symlinked intermediate directory, `mkfifo`, symlink to `/dev/null`, `tests/../../etc`. `lstatSync` on a symlink-to-directory gives `isDirectory()===false, isSymbolicLink()===true`, so the new branch is never entered and rejection comes from the pre-existing `!isFile()`. Guard ordering confirmed correct in both files.
- **No split-brain.** Both sites gate on the identical predicate at the identical position. One reviewer checked the cross-check that *would* have created one — the runtime gate's `proof.files` lookup (`artifact-gates.ts:1114`) applies only to `entries[]`, never probes, and `invariantSourceProofSchema.files` has no `minItems`, so a probe skipping `files.set` cannot fail the twin.

### Fixed in c8923fb

**Both reviewers flagged the same gap, and they were right: the template branch — the one that actually failed R45 — had zero coverage.** Every test in the original push targeted `artifact-gates.ts`. I measured the consequence rather than assuming it: reverting *only* the template hunk leaves `artifact-gates.test.ts` green at **42/42**. CI would have passed while the next Aave run died with the identical error. That is the same split-brain class of bug this PR fixes, so it was worth the heavier fix.

`verifyInvariantLedgerSourceEvidence` is now extracted in `generated-workflow-verifier.test.ts`, following that file's existing pattern for `restoreInvariantSuiteWorkspaceSnapshot`, with `readInvariantSourceSnapshot` stubbed to record which paths the loop still tries to snapshot. Two behavioural assertions: a directory probe never reaches it; a symlink-to-directory probe does and fails there. Proven red on the reverted template, green on the fix.

**The first reviewer also found that my "still rejects a directory invariant source" test passed for the wrong reason.** The ledger omitted `kind`/`inventory_ids` and added a non-schema `statement`, so `invariantLedgerSchema` (a `strictObject`) rejected it and `artifact-gates.ts:226` early-returned before the source rule ever ran. `assert.equal(result.ok, false)` was therefore green regardless — the PR's central safety claim was unguarded, and the test would have stayed green if directory invariant sources were fully accepted. Now schema-valid with a complete markdown handoff so the directory source is the only thing that can fail it, asserting the real diagnostic (`not-file`) and explicitly ruling out `INVARIANT_LEDGER_SCHEMA_INVALID`.

Also added: symlink-probe rejection on the runtime-gate side, pinning the `!isSymbolicLink()` conjunct that nothing previously guarded. Both reviewers noted that conjunct is dead under `lstat`; it stays as defence against a future `lstat`→`stat` edit, and is now tested rather than merely asserted.

**Wording nit taken.** "regular file or directory" → "regular file or a non-symlink directory": the only case that reaches that message is a symlink to a directory, which `ls` calls a directory.

### Deliberately not fixed here

- **Prompt clarification** → #291. The prompt never says a probe may name a directory, and its own suggested `grep -rIlE ... test contracts` takes directory arguments, so `tests` is its natural output. That is the strongest argument for this fix. But changing a prompt changes model output and would confound the run comparison this PR exists to unblock, so it is filed rather than bundled.
- **No-evidence ledgers are unfalsifiable** → #292. Pre-existing on `main` and verified there by one reviewer, so out of scope; filed with the threat model and three options so it stops being rediscovered.

### Validation

- `artifact-gates.test.ts` 42/42, `generated-workflow-verifier.test.ts` 37/37
- full runtime suite green before the review fixes: 306 tests, 292 pass, 0 fail, 14 skipped; re-running now on c8923fb

aviggiano:
## Second independent reproduction — the failure is not effectively intermittent

R45's overseer retried `project-discovery` on its own (recovery generation 1, `17:12:42Z → 17:41:24Z`) and it failed with the **byte-identical error**:

```
artifact-contract failure: invariant scan probe tests is unavailable: scan probe is not a regular file
```

Two independent attempts, same model, same target, same probe path. I had recorded this as intermittent because R44's `project-discovery` recorded only file probes and passed on the immediately preceding image — that reasoning was too weak. Naming `tests` is what the prompt's own suggested `grep -rIlE ... test contracts` invites (#291), so it reproduces rather than randomises.

Consequences worth stating plainly:

- **R45 cannot make progress on an image without this fix.** It will keep burning recovery generations on the same node until the budget is exhausted. It is not being stopped — but nothing downstream of `project-discovery` can run, so no property lens, no fan-in, no harness, no report.
- **The change is load-bearing, not defensive.** This is the fourth plumbing defect in this pipeline (#275, #281, #283, #289) and the first to reproduce on retry within a single run.

Nothing about the diff changes on this evidence; it just removes the "maybe it self-heals" reading. Full monorepo suite is green at 436 tests / 422 pass / 0 fail / 14 skipped.

For the record on CI: `release-gates` was cancelled twice at exactly 15m with **zero steps executed** and the scheduled `Publish Eval History` / `Recover Incomplete Modal Benchmarks` crons on `main` failed in the same window — hosted-runner starvation, not this branch. Re-queued. Greptile, semgrep and both Socket checks are green.

aviggiano:
## Merging with `release-gates` infra-cancelled — recorded openly

`release-gates` has now been **cancelled three times at almost exactly 15 minutes queued with zero steps executed**:

| job | queued | cancelled | steps |
| --- | --- | --- | --- |
| 92688882023 | 17:32:59Z | 17:48:02Z | 0 |
| 92689030888 | 17:33:57Z | ~17:48Z | 0 |
| 92692689305 | ~17:57Z | ~18:13Z | 0 |

Zero steps means it never reached `Check out repository`, so nothing in this branch can be the cause. Corroborating evidence that this is org/platform-level runner starvation rather than anything about #290: the scheduled `Publish Eval History` and `Recover Incomplete Modal Benchmarks` workflows on `main` (`fdce587`) failed in the same window, and `Modal Eval Benchmarks` had both `launch` and `collect` cancelled with zero steps too. There is no `gh run cancel` anywhere in `.github/workflows/`, and `ci.yml`'s concurrency group is per-ref with no newer push to supersede these, so the cancellations are not self-inflicted. `timeout-minutes: 30` cannot produce a 15m cancellation either.

Merging anyway, and stating the tradeoff rather than burying it:

**For**
- Both required independent reviews are complete; every actionable finding is fixed in `c8923fb`, including the two the reviewers found in *my tests* rather than in the change.
- `main` has no branch protection, so `release-gates` is not a required check, and the same deviation was recorded on #284.
- Every substantive check that did run is green: Greptile, semgrep, Socket Project Report, Socket PR Alerts.
- `release-gates` runs `pnpm test`. That exact command is green on this exact tree: **436 tests, 422 pass, 0 fail, 14 skipped, EXIT=0** (`packages/runtime` 309/295/0/14). Verified `HEAD == origin/fix/invariant-directory-scan-probes == c8923fb` with no uncommitted tracked changes, so the suite ran on the merged content, not an earlier tree.
- R45 is dead-ended: two independent `project-discovery` attempts failed identically, so every hour without this fix is a wasted run.

**Against, honestly**
- I have no *CI* confirmation of the tests on `c8923fb`, only local. The mitigation is that the command and tree are identical, and I will re-run `release-gates` on `main` after merge once runners recover; if it fails there for a real reason, that is a revert, not a rationalisation.

aviggiano:
## Post-merge confirmation: the whole `release-gates` job reproduced locally on merged `main`

Merged as `23813cd`. GitHub Actions never created a CI run for it — Actions is still degraded, and the two previous merges show the same thing (`main` CI for `fdce587` and `ec8594e` both `cancelled` with **zero steps**, so this predates today's merges and is not caused by them). Since `ci.yml` has only `push`/`pull_request` triggers and no `workflow_dispatch`, there is no way to force it.

So I ran every step of the `release-gates` job locally against `HEAD == 23813cd` instead of waiting:

| step | result |
| --- | --- |
| `pnpm -w format:check` | ok |
| `pnpm -w lint` | ok |
| `pnpm -w build` | ok |
| `pnpm -w benchmark:check` | ok |
| `pnpm -w validate:release` | ok |
| `pnpm test` | **436 tests, 422 pass, 0 fail, 14 skipped** |

The only steps not reproduced are the environment setup actions (checkout, pnpm/node/bun, `pnpm install --frozen-lockfile`) and the artifact upload. The lockfile is unchanged by this PR, so `--frozen-lockfile` has nothing new to resolve.

That is the confirmation the merge comment promised, and it is stronger than a green check mark on a re-run would have been, because it is the merged tree rather than the PR head. Script and full log kept at `release-gates-local.sh` / `release-gates-local-23813cd.log` in the ops directory.

**R46 is live on this fix:** image `im-01KZC4NQPFJFHBK932P48YGQPM` built from `23813cd`, run `aave-v4-v0012-main-issue289-r46-invariant-only`, fresh durable volume, 9 nodes succeeded and `project-discovery` running. R45 was deliberately left running on the unfixed image as the control, which makes this the closest thing to a clean A/B this investigation has had — exactly one changed variable between the two.

Issue #285, opened 2026-08-06
Lens reference-expectation authority: the prompt and the gate disagree about what "supplied" means


aviggiano:
## Summary

The property-lens prompt and the provenance gate use two different definitions of a "supplied" reference expectation, so in the configuration these benchmark runs actually use, **every** citation a lens emits is unauthorized. Raised by adversarial review of PR #284; splitting it out because it is a policy decision, not a defect in that diff.

## The disagreement

The prompt tells the model to copy identifiers out of the pinned-reference **documents**:

> Populate `reference_expectations` only from exact identifiers present in the supplied pinned-reference artifacts. Preserve one identifier per named expectation and carry the supplied identifier unchanged.
>
> — `.ultrafuzz/prompts/properties/0kn0t-lens.md:31-34`

The gate only authorizes identifiers that appear in a structured `ultrafuzz/reference-expectations@1` catalogue carrying `operator-supplied` provenance (`packages/runtime/src/artifact-gates.ts`, `readLensSuppliedExpectationIds`). That catalogue comes solely from the optional `--reference-expectations` CLI flag.

For most lenses the pinned reference is a Markdown corpus, not such a catalogue. `.ultrafuzz/references.yml` pins `properties.0kn0t` to `0kn0t/properties_framework` with `README.md`, `01_PROPERTY_CLASSIFICATION_TAXONOMY.md` and friends — documents, containing identifiers like `LEND_ACC_01`.

And when no dependency declares a catalogue, `readLensSuppliedExpectationIds` returns an empty id set **with no diagnostic**:

```ts
if (expectationPaths.length === 0) continue;   // artifact-gates.ts
```

So "absent from every supplied catalogue" is the universal case, not the exceptional one.

## Observed consequence

Aave v4 run R44 (`aave-v4-v0012-main-issue272-r44-invariant-only`) supplied no catalogue — its config keys are `app_name, benchmark_execution, braintrust, eval_reporting, ground_truth, image_name, models, node_timeout_seconds, run_id, schema_version, target`. Node `property-specification-0kn0t` cited twelve identifiers and failed on all of them:

```
LEND_ACC_01 LEND_ACC_02 LEND_ACC_03 LEND_ACL_01 LEND_BND_01 LEND_BND_02
LEND_COL_01 LEND_DIR_01 LEND_DIR_02 LEND_DIR_03 LEND_ORC_01 LEND_STM_01
```

[...]

PR #294, opened 2026-08-06
fix(runtime): authorise lens expectations by namespace, not by label shape

### LINKED ISSUE #293: Property lens expectations in camelCase are neither authorised nor stripped, so the node dies (third shape of #285)
Third live occurrence of the same underlying defect, under a third identifier shape. Found in Aave v4 run R45 at `2026-08-06T18:28:20Z`.

## The failure

`property-specification-runtime-verification` failed with:

```
Property lens expectation "testConvertToAssetsSharesDesirable" is not present in a supplied pinned-reference catalog;
Property lens expectation "testConvertToSharesAssetsDesirable" is not present in a supplied pinned-reference catalog;
Property lens expectation "testPreviewDepositZeroAmountReturnsZero" is not present in a supplied pinned-reference catalog;
Property lens expectation "testPreviewRedeemZeroAmountReturnsZero" is not present in a supplied pinned-reference catalog;
Property lens expectation "testPreviewWithdrawPositiveAmountReturnsGtZero" is not present in a supplied pinned-reference catalog
```

These are test-function names from that lens's own pinned reference — the ERC4626 property suite. The lens prompt tells the model to carry identifiers unchanged from the pinned-reference documents, and it did.

## Mechanism, verified mechanically

`sanitizeLensReferenceExpectationAuthority` strips unauthorised expectations only when *every* unauthorised identifier looks like a catalogue label (`artifact-gates.ts:1725`):

```ts
if (unauthorized.some((expectationId) => !isUnsupportedExternalReferenceLabel(expectationId))) return property;
```

and the shape test (`artifact-gates.ts:1792`) is:

```ts
/^[A-Z][A-Z0-9]*(?:[-_][A-Z0-9]+)*[-_]\d+$/u
```

| identifier | matches catalogue shape |
| --- | --- |
| `LEND-01` | yes |
| `LEND_ACC_01` | yes (added by #284) |
| `ERC4626-01` | yes |
| `A16Z_ERC4626_07` | yes |
| `testConvertToAssetsSharesDesirable` | **no** |
| `testPreviewDepositZeroAmountReturnsZero` | **no** |

So a camelCase identifier is neither authorised nor strippable, and the node fails.

## History — this is whack-a-mole, and it should stop being played that way

1. `LEND-01`, hyphenated: stripped correctly.
2. `LEND_ACC_01`, underscored: **killed R44** → #283, fixed by #284 widening the shape to accept `_`.
3. `testConvertToAssetsSharesDesirable`, camelCase: **kills R45's runtime-verification lens** now.

A fourth widening would accept almost any token and would defeat the gate's purpose, so the shape test is the wrong axis. The real issue is the one #285 already states: the gate only honours identifiers from a structured `ultrafuzz/reference-expectations@1` catalogue, and these runs never supply one, so *every* citation is technically unauthorised and survival depends on cosmetic identifier shape.

## Suggested direction, not a patch

Distinguish two situations the gate currently conflates:

- **No catalogue supplied at all.** No identifier could ever be authorised, so failing the node punishes the model for following its prompt. Strip and warn.
- **A catalogue is supplied and the identifier is absent from it.** That is the forgery case, and it must keep failing closed.

This split looks compatible with the seven existing tests that pin byte-unchanged preservation of a forged benchmark mapping — worth verifying, because two earlier review suggestions (per-identifier stripping, and flipping the "unverifiable" policy) were rejected precisely for breaking them. Whoever picks this up should check whether those fixtures supply a catalogue; if they do, the split is safe and #285 can be closed with it.

## Severity

Not run-fatal. R45's fan-in aggregated the seven surviving lenses and the pipeline continued to `stateful-invariant-setup`. The cost is the **silent loss of one reference lens's entire contribution**, which is a coverage regression that no report would show as missing.

aviggiano:
## Summary

- fixes #293
- third live occurrence of one defect under a third identifier shape; stops chasing shapes and implements the rule the code's own comment already stated
- found in Aave v4 run R45

## The failure

`property-specification-runtime-verification` died at `2026-08-06T18:28:20Z`:

```
Property lens expectation "testConvertToAssetsSharesDesirable" is not present in a supplied pinned-reference catalog
Property lens expectation "testConvertToSharesAssetsDesirable" is not present in a supplied pinned-reference catalog
Property lens expectation "testPreviewDepositZeroAmountReturnsZero" is not present in a supplied pinned-reference catalog
Property lens expectation "testPreviewRedeemZeroAmountReturnsZero" is not present in a supplied pinned-reference catalog
Property lens expectation "testPreviewWithdrawPositiveAmountReturnsGtZero" is not present in a supplied pinned-reference catalog
```

These are test-function names from that lens's own pinned reference. The prompt tells the model to carry identifiers unchanged from the pinned-reference documents; it did.

## Why not widen the pattern again

| identifier | old shape test | outcome |
| --- | --- | --- |
| `LEND-01` | matches | stripped |
| `LEND_ACC_01` | did not match until #284 | **killed R44** (#283) |
| `ERC4626-01`, `A16Z_ERC4626_07` | match | stripped |
| `testConvertToAssetsSharesDesirable` | **no match** | **kills R45's RV lens** (#293) |

A pattern loose enough to admit camelCase test names admits nearly any token, which would hollow the gate out. The shape was only ever a proxy. The real distinction is in the function's own doc comment:

> Namespaced identifiers such as `benchmark:unexpected` deliberately do NOT match: those are the benchmark-mapping citations the gate exists to reject, and they must keep failing the node.

So the rule becomes exactly that:

- **unnamespaced** → a citation copied out of a reference document → strip it, emit `PROPERTY_REFERENCE_EXPECTATION_SANITIZED`
- **namespaced** (`benchmark:unexpected`, `scfuzzbench:aave-v4:iSpoke_supply`) → asserts an external authority blessed this property → still fails the node, bytes preserved for diagnosis

Stripping only ever *removes* a provenance claim, so it is the safe direction. Failing the node is what silently cost R45 an entire reference lens's contribution — a coverage regression no report would show as missing.

## Evidence that mattered

Two earlier review suggestions in this area (per-identifier stripping, and flipping the "unverifiable" policy) were rejected because they broke seven tests that pin **byte-unchanged preservation of a forged benchmark mapping**. Those seven still pass here. That is the check that distinguishes this change from those.

- new regression proven **red before the fix**, reproducing R45's diagnostic verbatim
- companion regression: a namespaced identifier mixed WITH a strippable one, so the all-or-nothing guard is exercised and a future widening cannot quietly relax it
- `artifact-gates.test.ts` 44 pass / 0 fail
- full monorepo suite **438 tests / 424 pass / 0 fail / 14 skipped**
- verified this rule has **one** implementation only, unlike #289 — no template twin (`grep` for `PROPERTY_REFERENCE_EXPECTATION_UNAUTHORIZED` hits only `artifact-gates.ts` and its test)

## Relationship to #285

#285 asks the broader question: the gate only authorises identifiers from a structured `ultrafuzz/reference-expectations@1` catalogue that these runs never supply, so in this configuration *every* citation is unauthorised. This PR does not answer that. It makes the strip-versus-fail decision turn on whether external authority is *asserted*, which is a property of the identifier rather than of its cosmetics. Supplying a real catalogue remains the better long-term answer and #285 stays open for it.

<h3>Greptile Summary</h3>

The PR replaces shape-based classification of unauthorized property-lens expectation identifiers with namespace-aware classification while retaining fail-closed handling for reserved authority claims.
- Strips plausible copied citations regardless of whether they use label, dotted, line-reference, or camelCase forms.
- Preserves and rejects identifiers claiming reserved benchmark authority.
- Adds regression coverage for the R45 failure, separator and case variants, malformed entries, and mixed citations.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| packages/runtime/src/artifact-gates.ts | Replaces shape matching with bounded citation recognition and a case-normalized reserved-authority namespace allowlist. |
| packages/runtime/test/artifact-gates.test.ts | Adds focused regressions covering camelCase citations, valid punctuation variants, malformed values, and reserved namespaces. |

</details>

```mermaid
flowchart TD
  A[Unauthorized reference expectation] --> B{Plausible copied citation?}
  B -->|No| C[Preserve artifact bytes]
  C --> D[Emit unauthorized error]
  B -->|Yes| E{Reserved authority namespace?}
  E -->|Yes| C
  E -->|No| F[Strip expectation identifier]
  F --> G[Refresh verification digest]
  G --> H[Emit sanitized warning]
```
</details>

<sub>Reviews (4): Last reviewed commit: ["fix(runtime): close keyword-internal sep..."](https://github.com/monad-developers/ultrafuzz/commit/99d871e27ca063b5bb39a64111745a6d7b2a365d) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=50915595)</sub>

**Context used:**


aviggiano:
## Both reviews rejected this; both were right. Reworked in `985667d`.

The reviewed predicate `!expectationId.includes(":")` was wrong in **both** directions, and each reviewer found a different half. Neither finding was arguable, so I verified both in the source rather than debating them.

### Half one: the fail-closed branch was dead code

Real catalogue identifiers are not colon-namespaced:

- `packages/artifacts/schema/reference-expectations.schema.json` — `id` is `{type: string, minLength: 1}`, **no pattern**
- the artifact contract's own canonical example is `expectation-example` (`artifact-contracts.ts`)
- the ScFuzzBench ground truth uses bare labels: `total-borrowed-v0`, `monotonic-index-price`
- `.ultrafuzz/references.yml` namespaces with **dots** (`properties.crytic`)
- `scfuzzbench:aave-v4:...` appears **only** in fixtures and docs

So for any real catalogue the branch never fired and the rule was "strip everything". One reviewer proved it with a probe harness: a near-miss citation (`expectation-exampl` against a supplied `expectation-example`) went from failing loudly on `main` to being silently stripped — including for runs that *do* supply a catalogue, which is the benchmark-scoring case this gate exists for.

### Half two: it recreated the same trap with new punctuation

A colon appears in perfectly ordinary citations — a source URL, `RoundingProps.sol:88`, `ERC4626-01: totalAssets never reverts`. Hard-failing an entire lens on one character of model-authored free text is precisely the failure mode this change set out to stop.

### What `985667d` does instead

1. **Bounds the strippable set positively** — `^[A-Za-z][A-Za-z0-9._-]{0,63}$`, a single-line token that could plausibly have been copied out of a reference document. The negative bound swallowed a sentence, a whitespace-only entry, a URL, and a JSON blob; all four now fail closed and are pinned by a new test. `testConvertToAssetsSharesDesirable` (#293) and `LEND_ACC_01` (#283) both match, so it stays exactly as permissive as it needs to be.
2. **Makes the authority check separator-agnostic** — `^(?:scfuzzbench|benchmark|ground-?truth)[:._/-]`, case-insensitive after `trim()`. A colon-only allowlist would have let `scfuzzbench_aave_v4_iSpoke_supply` through as an ordinary citation. Now pinned across `:`, `.`, `/`, `_`, `-` and mixed case.
3. **States the cost of stripping instead of asserting it away.** The earlier comment claimed "stripping only ever REMOVES a provenance claim, so it is the safe direction". Both reviewers showed that understates it: the sanitizer re-seals the verification marker via `refreshVerifiedArtifactDigest`; `reference_expectations` force-selects a property for implementation regardless of the priority threshold (`report-artifacts.ts:294-300`), so stripping can quietly drop a property from implementation; and fan-in's `PROPERTY_REFERENCE_EXPECTATION_DROPPED` checks skip rows with empty expectation lists. What stripping *cannot* do is manufacture credit — `reference_expectation_ids` is report-only and no grading path reads it. That asymmetry is the real justification and it is now written in the code.

### One suggestion declined, on measurement

"Never strip when a catalogue IS supplied" is a good instinct, but it fails `property lens gate strips unsupported external labels without weakening benchmark authority`, which deliberately pins partial retention with a catalogue present — an unauthorised label stripped while a namespaced one is preserved. I implemented it to check rather than reasoning about it: **45/46 with it, 46/46 without**. Changing that behaviour needs its own PR and its own argument about that test's intent.

The remaining suggestion — persisting `removed_reference_expectations` into a durable event payload so a sanitized run is auditable after the fact (one reviewer confirmed zero occurrences of `PROPERTY_REFERENCE_EXPECTATION_SANITIZED` anywhere in the run archive, so "stripped-and-forgotten" is literal) — needs a `run-state.schema.json` change and is filed separately rather than bundled.

### #293 will not be closed by this PR alone

One reviewer settled the open question by reading R45's untruncated `last_error` (2366 chars, ends cleanly): **19** distinct unauthorised IDs, all camelCase `testXxx`, **none containing a colon or any authority prefix**. So the node will strip all 19 and pass. That is strong evidence but still inference from a diagnostic rather than from the artifact, so this says `Refs #293`; R46's `runtime-verification` lens is the empirical test and #293 closes on that.

### Housekeeping

The previous push accidentally included **195 files** of generated test detritus under `packages/runtime/.ultrafuzz/` from a `git add -A`. Removed and force-pushed. The only credential-shaped value among them was a test placeholder `api_key`, checked before the rewrite; the files remain on disk untracked.

### Validation

- artifact-gates **46/46**, including the seven tests that pin byte-unchanged preservation of a forged benchmark mapping
- new tests cover: camelCase strips, dotted/mixed-separator citations strip, implausible entries fail closed, reserved-authority forgeries fail closed across five separators and mixed case
- typecheck, eslint, prettier clean; full monorepo suite running

One process note worth flagging: a reviewer found the shared checkout was being mutated concurrently by my own work, so the prescribed in-repo `tsc && node --test` compiled a different branch's source and produced a green run that proved nothing. Reviewing a branch in a live working tree is not sound; a worktree per reviewer is.

aviggiano:
Full monorepo suite green on `985667d`: **440 tests / 426 pass / 0 fail / 14 skipped**, EXIT=0.

aviggiano:
## Verification review passed; its two findings are fixed in `99d871e`

The verification pass confirmed both original objections are resolved (mutation testing: neutering either half of the predicate breaks exactly one test each, so both halves are live and pinned — the rejected `!id.includes(":")` was dead code). It then found two things worth fixing, both of which I reproduced end-to-end through the real gate before changing anything.

**A live bypass: `ground-?truth` only made the HYPHEN optional.** So `ground_truth.total-borrowed-v0`, `ground.truth.x` and `sc_fuzzbench.x` were silently stripped with a warning while `ground-truth.x` and `groundtruth.x` hard-failed — the exact "same forgery with different punctuation" this design claims to prevent. Now separator-agnostic *inside* the keyword too: `^(?:sc[._-]?fuzz[._-]?bench|benchmark|ground[-._]?truth)[:._/-]`. A separator is still required after the keyword, so `benchmarking` and `Benchmarks-are-fine` remain citations rather than authority claims.

**The advertised normalisation was not pinned.** Deleting `.toLowerCase()` left the suite green, because every mixed-case forged variant also contained a colon and so failed via the implausibility bound regardless. The forged loop now contains *only* entries that match `COPIED_REFERENCE_CITATION`, so each one pins the authority check itself. Measured: dropping `.toLowerCase()` now fails exactly one case (45/46).

**`.trim()` removed rather than left as unpinnable decoration.** The reviewer proved it unreachable — `isCopiedReferenceCitation` tests the raw string first, so whitespace fails outright and the authority check never sees an untrimmed value. Dropping it left the suite green, which is the definition of dead code. Removed, with the reasoning recorded in place of the code.

**Two shapes admitted that would otherwise have been the next trap:** a leading digit or underscore (`4626-01`, `_internal`) and an interior colon (`RoundingProps.sol:88`). All plausible things to copy out of a reference document, and hard-failing a lens on them would repeat #283 and #293 with new characters. The colon is safe because the authority check runs first and wins. This also corrects an overclaim in my previous message, which cited `RoundingProps.sol:88` as fixed while it was in fact still failing.

Free text, whitespace-only entries, URLs and JSON blobs continue to fail closed. Unicode lookalikes were confirmed to fail closed too.

### Also worth recording

The reviewer found the suggested review command produces a **false green** in a fresh worktree: with no `node_modules`, `npx tsc` fetches a decoy package that prints "This is not the tsc command you are looking for", exits 0, and compiles nothing. Anyone chaining `&&` gets a clean run with zero tests executed. Worth knowing for future reviews of this repo.

### Validation

- artifact-gates **46/46**; full monorepo suite **440 / 426 pass / 0 fail / 14 skipped**
- every `release-gates` step reproduced locally on `99d871e`: `format:check`, `lint`, `build`, `benchmark:check`, `validate:release` — all ok
- Greptile, semgrep and both Socket checks green

Merging with the same recorded deviation as #284 and #290: GitHub's hosted runners are still evicting queued jobs, so `release-gates` never ran on CI. It is not a required check on an unprotected `main`, and the job's steps are green locally on this exact tree.

PR #296, opened 2026-08-06
fix(modal): treat a shut-down sandbox as an absent optional file

### LINKED ISSUE #295: overseer crashes when a supervised sandbox has shut down, abandoning the run it exists to recover
Confirmed live in Aave v4 run R45. This one is costly out of proportion to its size: it silently removes supervision from a run, so the *next* transient failure becomes permanent.

## What happened

`2026-08-06T18:24:54Z` — the detached overseer for `aave-v4-v0012-main-issue281-r45-invariant-only` died with an unhandled rejection and exited 1:

```
NotFoundError: The Sandbox is unavailable. This Sandbox may have already shut down.
    at translateExecSandboxUnavailableError (modal/dist/index.js:62586:10)
    at translateExecErrors (modal/dist/index.js:62605:13)
    at async readOptionalModalSandboxText (packages/modal/dist/runner.js:2381:26)
```

R45 then ran with **no supervision at all**. It still made progress on its existing generation — it passed `property-specification-fanin`, the point where R43 stalled permanently, and generated all four Solidity harness files. Then at `18:56:35Z` `stateful-invariant-setup` failed with `provenance.failure.category: "agent-failure"` and no `last_error`, i.e. a transient agent failure of exactly the kind recovery exists for.

Nothing retried it. The furthest-progressed run in the whole investigation sat terminal for 13 minutes for no reason other than its supervisor having died half an hour earlier. Restarting the overseer immediately produced `action: "launch", reason: "owner-missing"` and it re-armed the run — so recovery was available the entire time and simply had no process to trigger it.

## Root cause

`readOptionalModalSandboxText` (`packages/modal/src/runner.ts:2957`) treats only one error as "absent":

```ts
} catch (error) {
  if (error instanceof SandboxFilesystemNotFoundError) return undefined;
  throw error;
}
```

When the sandbox is gone, the SDK raises `NotFoundError`, not `SandboxFilesystemNotFoundError`, so it propagates and kills the process.

For an **optional** read those two conditions are the same answer: the readiness marker cannot be read. A dead sandbox definitively has no readable marker.

The file already treats `NotFoundError` as absence in six other places, including the immediately adjacent sandbox lookup at `runner.ts:1858`:

```ts
} catch (error) {
  sandbox?.detach();
  if (error instanceof NotFoundError) return undefined;
  throw error;
}
```

`NotFoundError` is already imported at `runner.ts:10`. So this reads as an oversight rather than a policy choice.

## Why it matters more than a crash normally would

An overseer's entire purpose is to supervise runs whose sandboxes die. Dying *because* a sandbox died inverts that. And the failure is silent from the run's point of view: run state still says `running`, progress continues on the existing generation, and the loss only becomes visible at the next failure — by which time the log line that explains it is far in the past.

Also worth noting: the overseer runs **locally** from `packages/modal/dist/cli.js`, not inside the immutable benchmark image. A fix therefore protects runs that are already in flight, without rebuilding an image.

## Suggested fix

Treat `NotFoundError` as absence in `readOptionalModalSandboxText`, matching the sibling helper. A regression test can drive it with a filesystem stub whose `readText` rejects with `NotFoundError`.

Independently, consider whether the overseer's top-level poll loop should survive *any* unexpected error for one cycle rather than exiting, since exiting is always the worst available outcome for the thing it supervises. That is a larger design question and belongs in its own change.

## Mitigation in place

The ops wrapper for R45 now restarts the overseer on non-zero exit, logging every restart so a crash loop stays visible. That is a workaround, not a fix.

aviggiano:
## Summary

- fixes #295
- one-line behaviour change, but it cost the furthest-progressed run in the Aave v4 investigation its supervision
- the overseer runs **locally**, not inside the immutable benchmark image, so this protects runs that are already in flight

## What happened

R45's detached overseer died at `2026-08-06T18:24:54Z` and exited 1:

```
NotFoundError: The Sandbox is unavailable. This Sandbox may have already shut down.
    at translateExecSandboxUnavailableError (modal/dist/index.js:62586:10)
    at async readOptionalModalSandboxText (packages/modal/dist/runner.js:2381:26)
```

The run then ran unsupervised. It still progressed on its existing generation — it passed `property-specification-fanin`, the point where R43 stalled permanently, and generated all four Solidity harness files. Then at `18:56:35Z` `stateful-invariant-setup` failed with `provenance.failure.category: "agent-failure"` and no `last_error`: a transient agent failure, precisely what recovery exists for.

Nothing retried it. Restarting the overseer immediately produced `action: "launch", reason: "owner-missing"` and it re-armed the run — recovery was available the entire time and had no process to trigger it.

## Root cause

`packages/modal/src/runner.ts:2978` treated one error as absence:

```ts
if (error instanceof SandboxFilesystemNotFoundError) return undefined;
throw error;
```

For an **optional** read, "the file is not there" and "the sandbox holding it is gone" are the same answer. A shut-down sandbox definitively has no readable file.

Three things make this an oversight rather than a policy choice:

1. `runner.ts` already treats `NotFoundError` as absence in **six** other places.
2. The nearest sibling — the sandbox lookup at `runner.ts:1858` — does exactly `if (error instanceof NotFoundError) return undefined;`.
3. `NotFoundError` was already imported at `runner.ts:10`.

## Why the impact is out of proportion to the diff

An overseer's whole purpose is supervising runs whose sandboxes die. Dying *because* a sandbox died inverts that. And it fails silently from the run's point of view: run state still reads `running`, progress continues on the existing generation, and the loss only surfaces at the next failure — by which time the log line explaining it is half an hour in the past. That is what made R45 look like a plumbing failure when it was actually an unsupervised transient.

## Test note

The existing test was named *"treats **only** an explicit remote not-found as an absent persisted file"* — it deliberately pinned the narrower rule. Its name and intent are updated to the two-class rule, and the generic-`Error` rethrow assertion is kept so this cannot drift into a catch-all.

## Validation

- regression proven **red before the fix**, reproducing the production error verbatim: `promise rejected "NotFoundError: The Sandbox is unavailable…" instead of resolving`
- `packages/modal` **305 tests / 305 pass**
- full monorepo suite running

## Out of scope, deliberately

#295 also raises whether the overseer's top-level poll loop should survive *any* unexpected error for one cycle rather than exiting, since exiting is always the worst outcome for the thing it supervises. That is a design question about failure policy and belongs in its own change, not bundled with a not-found classification fix.

The ops wrappers for both live runs now restart the overseer on non-zero exit with every restart logged. That is a workaround, not a substitute for this.

<h3>Greptile Summary</h3>

The PR makes Modal benchmark supervision resilient to transient per-job polling failures while retaining bounded failure handling.

- Adds a configurable consecutive-failure threshold.
- Tracks failures independently per supervised job and abandons persistently failing jobs without immediately stopping healthy siblings.
- Keeps optional sandbox-file reads narrow by continuing to rethrow ambiguous Modal `NotFoundError` failures.
- Adds regression coverage for transient failures, partial completion, and abandonment behavior.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| packages/modal/src/defaults.ts | Adds the default threshold governing how many consecutive failed poll ticks a job may encounter before abandonment. |
| packages/modal/src/runner.ts | Introduces bounded per-job poll-error handling and preserves the narrow optional-file absence classification. |
| packages/modal/test/runner.test.ts | Covers transient polling failures, completion while a sibling fails, persistent per-job failure, and all-job failure behavior. |

</details>

```mermaid
flowchart TD
  P[Poll active jobs] --> J{Job poll succeeds?}
  J -->|Yes| R[Reset job failure count]
  J -->|No| I[Increment job failure count]
  I --> T{Threshold reached?}
  T -->|No| P
  T -->|Yes| A[Abandon failing job]
  R --> C{All active jobs complete?}
  A --> C
  C -->|No| P
  C -->|Yes, no abandoned jobs| S[Return snapshots]
  C -->|Yes, abandoned jobs exist| E[Exit with aggregated error]
```
</details>

<sub>Reviews (3): Last reviewed commit: ["fix(modal): count overseer tick failures..."](https://github.com/monad-developers/ultrafuzz/commit/3f29ffce9f3bf9279fa19bebac2fe822863f0040) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=50917664)</sub>

**Context used:**


aviggiano:
Full monorepo suite green on `5e2d701`: **436 tests / 422 pass / 0 fail / 14 skipped**, EXIT=0. `packages/modal` alone is 305/305.

Operational note while this is in review: both live runs' ops wrappers now restart the overseer on non-zero exit with each restart logged. Restarting R45's produced `action: "launch", reason: "owner-missing"` within seconds, and it is now on recovery generation 3 at `healthy / keep / canonical-progress` — which confirms the diagnosis from the other direction: recovery was available the whole time and simply had no live process to trigger it.

aviggiano:
## Both reviews rejected this; both were right. Reverted and reworked in `9850366`.

The one-line change was wrong on measurement in two independent ways, and I verified both in the source before acting.

### The premise was false

`modal@0.9.0` relabels **five** distinct gRPC conditions as the same error class (`index.js:62575-62589`):

```js
function isSandboxUnavailableError(err) {
  if (err instanceof NotFoundError) return true;
  if (err instanceof ClientError11) {
    return err.code === Status11.NOT_FOUND || err.code === Status11.CANCELLED
        || err.code === Status11.UNKNOWN || err.code === Status11.DEADLINE_EXCEEDED
        || err.code === Status11.UNAVAILABLE;
  }
  return false;
}
// → new NotFoundError("The Sandbox is unavailable. This Sandbox may have already shut down.")
```

So an ordinary `DEADLINE_EXCEEDED` or `UNAVAILABLE` blip against a perfectly healthy sandbox arrives with that exact "may have already shut down" text. My comment's claim — "a shut-down sandbox definitively has no readable file" — was a true statement about a false premise. Treating the class as "file absent" would let a blip **drive writes**:

- `finishReservedModalRecoveryWorker` (`runner.ts:1677-1700`): `published === undefined` defeats both the early return and the `"incompatible Modal recovery readiness marker"` guard, reaching `stageLaunchFiles`, which re-stages config, lineage and credential files **over a live worker**
- `collect` (`runner.ts:2327`): the retry shield `hasRetryableLiveStatusCollectionMismatch` requires `status !== undefined`, so it is bypassed in exactly the newly-created case and an **empty artifact bundle** gets persisted
- the overseer's volume inspection can read a finished run as not-complete and drive a spurious relaunch

The precedent I cited was also not analogous: the sibling `NotFoundError` catches are on `sandboxes.fromId/fromName` and `volumes.fromName` lookups, where the class genuinely means "does not exist" with no transient aliasing.

### It did not even achieve its goal

With the marker read returning `undefined`, control reaches `stageLaunchFiles` and then `publishLaunchReady`, both of which `exec` against the same dead sandbox and throw. Nothing between `overseeModalBenchmarkOnce` and `cli.ts` catches it, because `overseeModalBenchmarks` had no error boundary. **The crash was relocated, not removed** — and now with a chance of having re-staged files first. I verified the control flow at `runner.ts:1689-1717`.

### The actual defect, and the actual fix

`overseeModalBenchmarks` had no `try`/`catch` around its per-job call, so **any** throw from a poll tick took the whole detached process down — the sandbox-unavailable read was one of many. That is what happened to R45.

`9850366` reverts the helper change and fixes the loop:

- a failed tick is recorded as `failed_jobs` in the emitted snapshot and polling continues, so one wedged run cannot stop the others being supervised
- a persistently broken job still surfaces: the loop rethrows after `MODAL_OVERSEER_MAX_CONSECUTIVE_FAILURES` consecutive ticks in which **every** job threw
- the completion check now also requires `snapshots.length === jobs.length`. This one is subtle and worth naming: an all-failed tick leaves `snapshots` empty, and `[].every(...)` is `true`, so without the guard the overseer would have reported a wedged job as **complete** and exited — a second way to lose supervision, introduced by the very fix for losing supervision.

`readOptionalModalSandboxText` keeps its narrow behaviour, and its comment now records why widening it would be wrong so the next person does not retry this.

### Validation

- both new tests proven **red against the old loop** and green with the fix:
  ```
  × keeps polling after a job throws and reports the failure alongside healthy jobs
  × gives up loudly once every job has failed for the consecutive limit
  ```
- `packages/modal` **307 tests / 307 pass**

### Deliberately still out of scope

A reviewer suggested making "dead sandbox ⇒ absent" available as an opt-in per call site, plus a bounded retry to discriminate transient from terminal (since after translation the SDK gives no other way to tell). Both are reasonable and neither is needed for #295 now that the resilience sits in the loop, so they stay unbundled.

Also unchanged, and now clearly correct: guarding `stageLaunchFiles`/`publishLaunchReady` so they are never attempted against a sandbox that was not confirmed live after the read. That is a real gap, but it is a separate defect from the one this PR fixes.

aviggiano:
## Verification review: MERGE, and its three findings are fixed in `3f29ffc`

The reviewer confirmed both original objections are resolved — the revert is behaviourally identical to main (only comments differ), and both new tests were re-verified red against the pre-fix loop rather than taken on trust. They also checked something I had not: repeated absorbed failures leak nothing, because `withModalLaunchStateLock` releases before rethrowing and `modal.close()` runs in a `finally`, so a retried tick 60s later is equivalent to an operator restart — the exact path that recovered R45.

Three findings, all fixed:

**P2 — the shared counter only tripped when EVERY job failed on the same tick.** With two or more jobs a permanently broken one (incompatible config fingerprint, launch state naming a different run, a public config that forbids recovery — all thrown unconditionally) was absorbed forever, because a healthy sibling reset the count each tick. "Loud" silently became "never" precisely as job count grew. Failures are now counted **per job**.

**A wedged job is dropped from the poll set instead of killing the process.** Its healthy siblings keep being supervised — the whole point of the loop — and it is still never forgotten: once nothing remains to poll, or the remaining jobs complete, the loop ends non-zero naming every abandoned job. The snapshot now carries `abandoned_jobs` and a per-job `consecutive_failures` so it is visible from stdout as it happens.

**P3 — the give-up no longer rethrows verbatim.** The underlying error is usually the very symptom this loop exists to survive (`The Sandbox is unavailable...`), so rethrowing it would read to the next operator as this fix having regressed. It is now a descriptive Error naming the abandoned jobs and the limit, with the original attached as `cause`.

**P3 — a test comment overstated its own coverage, and the case it claimed to cover had no test.** The "keeps polling" comment said the tick "must not have been reported as complete even though its only successful snapshot said so", but on that tick the surviving snapshot has `complete: false`, so it never exercised the length guard. Comment corrected, and the realistic case is now pinned: a healthy job reporting complete on the same tick a sibling throws, where without the guard the loop returns a one-entry array for a two-job overseer and silently abandons the sibling.

### Noted, not fixed here

`probeModalSandbox` (`runner.ts:1191`) already maps `NotFoundError` → `{state: "missing"}`, which by this PR's own five-gRPC-codes argument means a transient `UNAVAILABLE` can make a live sandbox read as missing. Pre-existing and unchanged, but it is the same hazard class the new comment argues against, so the narrowness rationale is not yet applied uniformly. Worth its own issue rather than a bundled change.

Also unaddressed by design: no `--max-consecutive-failures` CLI flag, and `docs/how-to/run-evals-on-modal.md` does not yet describe `failed_jobs`/`abandoned_jobs`. Operator-relevant, but documentation and surface belong in a separate change from a supervision-correctness fix.

### Validation

- `packages/modal` **309 tests / 309 pass**; monorepo suite green
- both new tests re-verified red against the pre-fix loop; removing only the `snapshots.length === jobs.length` guard makes the give-up test fail, so that guard is genuinely pinned

Merging with the same recorded deviation as #284, #290 and #294: hosted runners are still evicting queued jobs, so `release-gates` never ran on CI; it is not a required check on an unprotected `main` and its steps are green locally.

Issue #292, opened 2026-08-06
A ledger with no entries and invented probe text satisfies both invariant evidence gates


aviggiano:
Surfaced independently by both reviews of #290. **Pre-existing on `main`, not introduced by that PR** — I am filing it so it stops being rediscovered.

This ledger passes the `project-discovery` gate:

```json
{
  "schema_version": "ultrafuzz.invariant-evidence-ledger.v1",
  "entries": [],
  "inventory_rows": [],
  "scan_probes": [{ "id": "p1", "source_path": ".", "query": "...", "result": "invented result text" }]
}
```

Verified by one reviewer against `origin/main` as well as the #290 branch, so the workspace-root allowance already granted exactly this. A probe naming a *nonexistent* path is likewise accepted, by design (`// Scan probes may intentionally target optional files. When a probe path is absent, its result text is the durable evidence of that absence.`) — and its `result` text is equally unchecked.

Two facts make this weaker than it looks, which is why it is worth reasoning about rather than panicking about:

1. Nothing anywhere reads `probe.result`. Every consumer is a presence check: `artifact-gates.ts` (`missing scan probe results`, `Property fan-in requires scan probe results from project discovery`) and `minItems` in `invariant-ledger.ts`. No code compares probe text to file bytes.
2. The schema forces `entries: []` and `inventory_rows: []` for the no-evidence shape, so nothing can be laundered *into* the invariant campaign. Ledger entries remain strictly byte-checked against the pinned commit.

So the exposure is that "the agent searched and found nothing" is unfalsifiable, and the fan-in gate's `scan_probes` requirement is satisfiable without work. Whether that matters is a policy question, same shape as #285:

- accept it — a probe is a record of intent, and the entries are where provenance is enforced;
- require at least one probe whose path exists and is snapshotted, so a run cannot claim a scan it did not do;
- keep accepting absent/directory probes but mark the derived proof as partial, so downstream consumers can tell.

Needs an owner with a view on the threat model. Not blocking any current run.

aviggiano:
**The hole is larger than filed.**

The fan-in gate returns at `packages/artifacts/src/artifact-gates.ts:417-437` for this shape and never calls `verifyInvariantProbePath` — so fan-in does not check probe containment **at all**, independently of the empty-ledger question. Rejecting empty ledgers will not fix that; it needs its own change.

[...]

Issue #291, opened 2026-08-06
project-discovery prompt never says a scan probe may name a directory


aviggiano:
Found by both independent reviews of #290, and it is arguably the root cause of the R45 failure that #289 describes.

`.ultrafuzz/prompts/setup/project-discovery.md` tells the agent to record "source path, query, and result" for each scan probe but never says what kind of path is acceptable. The "checked against the checked-out source" sentence applies to ledger *entries*, not probes. Meanwhile the prompt's own suggested probe command takes directory arguments:

```
grep -rIlE ... test contracts
```

So naming `tests` as a probe path is the prompt's natural output — which is exactly what R45's `project-discovery` did before it was rejected.

#290 makes the gate accept it, so nothing is broken. This issue is about closing the loop the other way: one sentence stating that a probe `source_path` may name a directory or an absent path, and that only ledger `entries[]` are byte-checked against the pinned commit.

Deliberately not bundled into #290. That PR is a gate fix on the path of a live investigation; changing a prompt changes model output and would confound the run-to-run comparison it is meant to unblock. Worth doing on its own, with the confound acknowledged.

Related: the same prompt-versus-gate mismatch pattern as #285, in a different gate.

PR #298, opened 2026-08-06
fix(runtime): require a ledger_ids field only where there are ledger IDs

### LINKED ISSUE #297: Property fan-in requires a ledger_ids field on every canonical property, but the schema and prompt both make it conditional
Confirmed from live artifacts in Aave v4 run R46, and it has now cost **two consecutive runs** a full `property-specification-fanin` attempt.

## The failure

Both R45 (18:46:01Z) and R46 (~20:05Z) failed fan-in with a long list of:

```
Properties Markdown must preserve canonical property "property-1" with its description, category,
priority, and sources (missing "ledger_ids")
```

Both then passed on a retry, so the pipeline survives — at the cost of a wasted node attempt each time, roughly ten minutes plus tokens, and a coin flip on a node every run must pass.

## Three components disagree, and the gate is the outlier

Read from R46's own fan-in artifacts on the volume:

- `properties.json` — **219** canonical properties, of which **78** have a non-empty `ledger_ids`
- `properties.md` — **219** blocks, and the string `ledger_ids` appears **exactly 78 times**

So the model rendered the field for precisely the properties that have one, and omitted it for the 141 that do not. That is exactly what it was told to do:

`.ultrafuzz/prompts/properties/property-specification-fanin.md:86`
> add a `ledger_ids` array to every canonical property **that represents one or more ledger entries**

And the schema agrees — `canonicalPropertySchema.ledger_ids` is `.optional()` (`packages/artifacts/src/property-provenance.ts:201-218`, and the TypeScript view at `:86` is `ledger_ids?: string[]`).

The gate does not (`packages/runtime/src/artifact-gates.ts:551`):

```ts
const ledgerFieldValues = block === undefined ? [] : markdownFieldValues(block, "ledger_ids");
const missingLedgerField = block !== undefined && ledgerFieldValues.length !== 1;
```

`property.ledger_ids` is never consulted. Every block must carry exactly one `ledger_ids` field, whether or not the property has any ledger mapping to render. A property derived purely from a reference lens — 141 of 219 here — cannot satisfy it without inventing a field the schema says is optional.

## Why it sometimes passes

Whether a retry survives depends on whether the model happens to emit a `ledger_ids` line for properties that have nothing to put in it. That is the same shape of luck as #283/#293: correctness by coincidence of formatting.

## Suggested fix

Require the field only when there is something to render:

```ts
const missingLedgerField =
  block !== undefined && (property.ledger_ids ?? []).length > 0 && ledgerFieldValues.length !== 1;
```

The parity checks that follow (`INVARIANT_LEDGER_MARKDOWN_MAPPING_MISSING`, the count comparison, the extra-ID check) already handle everything that matters once a property does have ledger IDs, so nothing is weakened: a property with ledger IDs must still render all of them, exactly once each, and must not add any.

Alternatively, make `ledger_ids` required with `min(0)` in the schema and say so in the prompt — but that means asking the model to emit `ledger_ids: ` with an empty value on two thirds of its rows, which is worse for no benefit.

## Severity

Not run-fatal, because retries have so far recovered. But it is on the critical path for every run, it wastes a full agentic node attempt when it fires, and it fires more often as the property count grows — R46 produced 219 properties, 141 of them exposed to it.

aviggiano:
## Summary

- fixes #297
- confirmed from live artifacts; it has cost **two consecutive Aave v4 runs** a full `property-specification-fanin` attempt
- three components disagreed about whether `ledger_ids` is conditional, and the gate was the outlier

## The failure

R45 (18:46:01Z) and R46 (~20:05Z) both failed fan-in with a long list of:

```
Properties Markdown must preserve canonical property "property-1" with its description, category,
priority, and sources (missing "ledger_ids")
```

Both passed on a retry, so the pipeline survives — at the cost of a wasted agentic node attempt each time and a coin flip on a node every run must pass.

## The measurement that settles it

Read from R46's own fan-in artifacts on the volume:

| | count |
| --- | --- |
| canonical properties in `properties.json` | **219** |
| of those, with a non-empty `ledger_ids` | **78** |
| blocks in `properties.md` | **219** |
| occurrences of `ledger_ids` in `properties.md` | **78** |

The model rendered the field for exactly the properties that have one. That is what it was told to do (`property-specification-fanin.md:86` — "add a `ledger_ids` array to every canonical property **that represents one or more ledger entries**") and what the schema allows (`canonicalPropertySchema.ledger_ids` is `.optional()`).

The gate never consulted `property.ledger_ids`:

```ts
const missingLedgerField = block !== undefined && ledgerFieldValues.length !== 1;
```

So all 141 properties with no ledger mapping were rejected for omitting a field the schema calls optional and the prompt does not ask for.

## The fix

```ts
const expectsLedgerField = (property.ledger_ids ?? []).length > 0;
const missingLedgerField = block !== undefined && expectsLedgerField && ledgerFieldValues.length !== 1;
```

Nothing is weakened. Once a property does have ledger IDs, the checks below still enforce that each one is rendered, the right number of times, with no extras — and a regression pins that half explicitly so this cannot be read as dropping ledger parity.

## Both my tests were green for the wrong reason first

Worth stating, because the exit status lied: the fixtures omitted the project-discovery evidence ledger, so the gate short-circuited on `INVARIANT_LEDGER_MISSING` before reaching any parity code, and the "does not demand" test passed against unfixed source. This is the same trap a reviewer caught in #290, and I only found it by reading the diagnostics rather than the pass count. With the ledger supplied, the red state reproduces R46's message verbatim.

## Validation

- new regression proven red before the fix, with the production message
- companion regression pins that a property WITH ledger IDs must still render the field
- `artifact-gates.test.ts` **48 pass / 0 fail**
- verified this rule has a **single** implementation — no generated-template twin, unlike #289
- typecheck, eslint, prettier, `git diff --check` clean; full monorepo suite running

## Why this keeps happening

Fifth defect of the same family: a gate enforcing something stricter than the prompt asks and the schema declares. #289 (directory scan probes), #291 (prompt silent on directory probes), #293 (identifier shape), #292 (unfalsifiable probe text), and now this. Individually small; together they are why no run has yet reached `report.md`. Worth someone auditing the gate/prompt/schema triple systematically rather than one live failure at a time.

<h3>Greptile Summary</h3>

The PR aligns the property fan-in artifact gate with the optional `ledger_ids` schema contract, requiring the Markdown field only for canonical properties that contain ledger IDs.
- Conditions the field-presence check on a non-empty canonical `ledger_ids` array.
- Adds regressions covering properties both without and with ledger IDs.

The PR appears safe to merge with no actionable correctness or security issues identified.

The conditional requirement matches the optional canonical field, while existing parity checks continue rejecting missing or extra ledger mappings for properties that contain ledger IDs.

| Filename | Overview |
|----------|----------|
| packages/runtime/src/artifact-gates.ts | Correctly narrows the ledger field-presence requirement while retaining existing missing, extra, and duplicate ID parity checks. |
| packages/runtime/test/artifact-gates.test.ts | Adds focused fixtures that reach the parity gate and verify both optional-field acceptance and required-field rejection. |

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(runtime): require a ledger\_ids field..."](https://github.com/monad-developers/ultrafuzz/commit/052f54e94a1f7c1b7922e946ba89f855e3ff2630) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=50943188)</sub>

**Context used:**


aviggiano:
Full monorepo suite green on `052f54e`: **442 tests / 428 pass / 0 fail / 14 skipped**, EXIT=0.

Deliberately **not** relaunching a run for this one. R47 launched at 20:06Z from `d78fe3a` and R46 is at `stateful-invariant-setup`; this defect costs a wasted fan-in attempt but retries have recovered from it in both runs that hit it, so a third ~$400 run to pick it up would buy very little. It lands for whatever comes next.

aviggiano:
## Both reviews: merge, no blocking findings

Both reviewers attacked the coverage question directly and both landed in the same place: **the "every ledger entry is mapped" guarantee never lived in the check this PR relaxes.** It lives at `artifact-gates.ts:477-488` (`INVARIANT_LEDGER_REFERENCE_MISSING`), a JSON-level check in the *entry* direction that runs before markdown is opened. The relaxed check is a markdown-level check in the *property* direction. Orthogonal.

The adversarial reviewer built seven fixtures and ran them against both `origin/main` and this head:

| fixture | scenario | result on this head |
| --- | --- | --- |
| ledger entries with **zero** properties referencing them, `ledger_ids` omitted everywhere | the feared regression | **rejected** — `INVARIANT_LEDGER_REFERENCE_MISSING` ×2 |
| 2 entries, only 1 mapped | partial coverage | **rejected** |
| property has `["ev-1","ev-2"]`, markdown renders only `ev-1` | under-render | **rejected** — `MAPPING_MISSING` |
| property has IDs, markdown has two `ledger_ids:` lines | arity | **rejected** |
| property has **no** IDs, markdown invents one | forged mapping | **rejected** — `MAPPING_EXTRA` |
| `"ledger_ids": []` in JSON | empty list | **rejected** — `PROPERTIES_SCHEMA_INVALID` |

**No fixture changed pass/fail between main and this head.** The unconditional demand was duplicate noise layered on checks that already fire.

That also answers the `[]` question: `minItems: 1` / `.min(1)` means an empty array cannot reach the code at all, so `(property.ledger_ids ?? []).length > 0` is exactly equivalent to `!== undefined` post-validation — defensive, not wrong.

**Mutation testing killed both mutants, one test each:** forcing `expectsLedgerField` true (≡ main) fails only the "does not demand" test; forcing it false fails only the "still requires" test. Neither survives, and no other test constrains the line in either direction. One reviewer noted the companion test adds coverage that *did not previously exist* — before this PR nothing pinned the "exactly one `ledger_ids` field" rule even for properties that do have IDs.

**On whether R46's 141 unmapped properties are themselves the defect:** no. R46 passed fan-in on retry, which means `INVARIANT_LEDGER_REFERENCE_MISSING` was satisfied — every ledger entry *was* covered by the 78 mapped properties. The other 141 are lens-derived properties with no doc-ledger backing, which is legitimate.

### The one finding, filed not bundled → #299

Both reviewers flagged that the prompt still reads unconditionally at `:93-94` ("include its complete `ledger_ids` list in that block"), and a fixture confirms a model obeying that literally is rejected with `MAPPING_EXTRA` — the mirror image of this bug, still live. It is a one-clause prompt change, but changing a prompt changes model output, so it belongs between runs rather than in a gate fix on the critical path of a live investigation. Filed as #299, with a suggestion to land it together with #291 as one prompt-clarity change.

Not acting on the `- ledger_ids: none` variant a reviewer raised as SUSPECTED: there is no evidence any model emits it, and adding speculative spellings to the filter is the whack-a-mole this series of fixes has been trying to stop.

### Validation

- artifact-gates **48/48**; monorepo **442 / 428 pass / 0 fail / 14 skipped**
- every `release-gates` step reproduced locally
- Greptile, semgrep and both Socket checks green

Merging with the same recorded deviation as #284, #290, #294 and #296: hosted runners are still evicting queued jobs, so `release-gates` never ran on CI.

One environment note for future reviewers of this repo, found the hard way twice: in a fresh worktree without `node_modules`, bare `npx tsc` fetches a decoy package that prints "This is not the tsc command you are looking for", **exits 0, and compiles nothing** — so `tsc && node --test` yields a green run with zero tests executed. And `artifact-gates.test.ts` test 1 reads `dist/topology.yml`, a build output, so it fails in any worktree where `pnpm build` has not run.

Issue #299, opened 2026-08-06
Fan-in prompt tells the model to render ledger_ids on every canonical row, which the gate rejects when the property has none


aviggiano:
Raised independently by both reviewers of #298, with the failure verified empirically. It is the mirror image of the bug that PR fixes, and it is still live.

## The mismatch

`.ultrafuzz/prompts/properties/property-specification-fanin.md:86` is conditional:

> add a `ledger_ids` array to every canonical property **that represents one or more ledger entries**

but `:93-94` reads unconditionally:

> render each canonical row in a delimited block … and **include its complete `ledger_ids` list in that block**

A model following the second sentence literally renders a `ledger_ids` line on a property whose JSON has none. A reviewer built that fixture and confirmed the gate rejects it with `INVARIANT_LEDGER_MARKDOWN_MAPPING_EXTRA` — costing the whole fan-in attempt, exactly as the omission did before #298.

So #298 removed one direction of this trap and left the other.

## Why the unconditional reading is unsatisfiable anyway

`ledger_ids` carries `minItems: 1` in `properties.schema.json:41-46` and `.min(1)` in `property-provenance.ts:201-203`, so an empty list cannot even be expressed — `"ledger_ids": []` fails with `PROPERTIES_SCHEMA_INVALID`. A property with no ledger mapping has no "complete `ledger_ids` list" to include. The sentence asks for something the schema forbids.

## Suggested fix

One clause on line 93-94:

> …and, **when the property has ledger IDs**, include its complete `ledger_ids` list in that block.

## Grouping suggestion

This is the second prompt/gate disagreement filed from live evidence (#291 is the first: the `project-discovery` prompt never says a scan probe may name a directory, while its own suggested `grep` takes directory arguments). Both are one-sentence clarifications to prompts, and both are the *documentation* half of a gate fix that already landed. Doing them together in a single prompt-clarity PR would be cheaper than two, and would let one person reason about the confound once: changing a prompt changes model output, so it should land between runs rather than during one.

I am not bundling it into #298, which is a gate fix on the critical path of a live investigation.

PR #300, opened 2026-08-06
docs(prompts): say that optional evidence fields are optional

### LINKED ISSUE #291: project-discovery prompt never says a scan probe may name a directory

aviggiano:
## Summary

- fixes #291 and #299
- the documentation half of two gate fixes that already landed (#290, #298)
- each of these prompts told the model to do something a gate then rejected, and each cost a live Aave v4 run a full agentic node attempt

## Fan-in `ledger_ids` — #299

`property-specification-fanin.md:86` was already conditional:

> add a `ledger_ids` array to every canonical property **that represents one or more ledger entries**

but `:92-95` then read unconditionally:

> render each canonical row in a delimited block … and **include its complete `ledger_ids` list in that block**

A model following the second sentence renders the field on a property whose JSON has none. A reviewer of #298 built that fixture and confirmed it is rejected with `INVARIANT_LEDGER_MARKDOWN_MAPPING_EXTRA` — the mirror image of the omission #298 fixed, and still live.

The unconditional reading is also unsatisfiable: `ledger_ids` carries `minItems: 1` in `properties.schema.json` and `.min(1)` in `property-provenance.ts`, so `"ledger_ids": []` is a schema error. A property with no mapping has no list to render.

## Project-discovery scan probes — #291

The prompt never said what kind of path a probe may name, while its own suggested command takes directory arguments:

```
grep -rIlE ... test contracts
```

R45 named `tests` and died on it twice. #290 made the gate accept a directory; this says so, and states the distinction that actually matters: only ledger `entries` are byte-checked against the pinned commit, because their bytes *are* the evidence. A probe records *where* the agent searched, so a directory or an absent path is fine.

## Why this is tested rather than editorial

Both clarifications are pinned by a new anchor in `packages/prompts/test/semantic-anchors.test.ts`, alongside the existing anchors for reference-expectation grounding and liveness requirements. Proven red against the unmodified prompts:

```
× prompt semantic anchors > states that optional evidence fields are optional
  Tests  1 failed | 23 passed (24)
```

A prompt that contradicts its gate is otherwise only discoverable by burning an agentic node — which is how every member of this family was found.

## Timing, deliberately

I filed both of these rather than bundling them into the gate PRs, because changing a prompt changes model output and would confound a run comparison. That constraint has now expired: R46 and R47 both have their prompts baked into immutable images, so neither can be affected. This is the right window.

## Scope

Text only, in two prompt files, plus one test. No gate, schema or runtime behaviour changes.

## The pattern worth naming

This is the fifth and sixth instance of one shape: a gate enforcing something stricter than the prompt asks and the schema declares. #289, #291, #293, #297, #299 — individually small, collectively the reason no run has yet reached `report.md`. The gates are now fixed and the prompts now agree, but nothing prevents the next divergence. A test that walks each artifact contract and asserts prompt/schema/gate agreement would catch the next one at CI time instead of costing a run.

<h3>Greptile Summary</h3>

The PR aligns two built-in prompts with their artifact contracts and adds semantic anchors to prevent those instructions from drifting again.
- Makes `ledger_ids` rendering conditional for canonical properties without ledger mappings.
- Documents valid directory, absent-path, regular-file, and symlink behavior for discovery scan probes.
- Adds whitespace-normalized tests that pin both prompt clarifications.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| .ultrafuzz/prompts/properties/property-specification-fanin.md | Clarifies that the optional, non-empty `ledger_ids` field must be omitted when a canonical property has no ledger mapping. |
| .ultrafuzz/prompts/setup/project-discovery.md | Documents accepted scan-probe paths and now unambiguously applies source line or symbol locations only to ledger entries. |
| packages/prompts/test/semantic-anchors.test.ts | Adds semantic anchors covering the optional ledger mapping and scan-probe path contracts. |

</details>

<sub>Reviews (3): Last reviewed commit: ["docs(prompts): apply every review findin..."](https://github.com/monad-developers/ultrafuzz/commit/66e042ef5f9ddfca36a17ecc2aefc6a4575e2788) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=50949790)</sub>

**Context used:**


aviggiano:
## I found a false sentence in my own prompt text — corrected in `8d2763c`

While the reviews were running I checked my own wording against the implementation, and one claim was wrong in exactly the way this PR exists to prevent.

**The false sentence:** "only ledger `entries` are checked byte-for-byte against the checked-out source."

**Why it is false:** a scan probe naming a regular file goes through the same `readInvariantSourceSnapshot` as an entry source. Under `usesPinnedSource` that runs:

```
git ls-files --error-unmatch -- <path>
git diff --quiet HEAD -- <path>
git show <pinnedSourceRef>:<path>   → byte comparison
```

and throws `invariant scan probe <path> is not pinned and unchanged` on any mismatch. Only **directory** and **absent** probes skip snapshotting — which is precisely the allowance #290 added, and nothing more.

**Why it was worse than inaccurate:** telling a model that probe paths are not checked invites exactly the carelessness #292 already flags about unfalsifiable probe evidence. I would have been fixing one prompt/gate divergence by writing another.

Corrected to state what is true and why it matters:

> A probe `source_path` may name a directory you searched or a path that turned out not to exist; either is accepted as a record of where you looked. When a probe names a regular file, that file must be tracked and unmodified in the checkout, exactly as a ledger entry's source must be. What a probe does NOT need is quoted text: an entry's `verbatim` is compared against its source, a probe's `result` is not.

Two further changes from the same pass:

- **The anchors now match whitespace-normalised text** instead of exact line breaks. Pinning `"when that property has ledger\nIDs"` would have turned any innocent reflow into a fake semantic regression — and I did reflow the paragraph after editing it, which would have broken it immediately.
- **One anchor asserts the false sentence is absent**, so it cannot come back by accident.

Still proven red against the unmodified prompts (1 failed / 23 passed) and 24/24 with them.

This does not change the reviews already in flight in any way that invalidates them — the fan-in half is untouched, and the discovery half is now more conservative than what was reviewed, not less. But reviewers should read `8d2763c` rather than `ca90e96` for the probe sentence.

aviggiano:
## Both reviews were NO-MERGE on the same P1, which was already fixed — and they found three more. All applied in `66e042e`.

The blocking finding — "only ledger `entries` are checked byte-for-byte" is false — I had already found and corrected in `8d2763c` by checking my own wording against the code before either review returned. Both reviewers independently confirmed it from the same evidence: `workflow.tsx:3467` sends a regular-file probe through `readInvariantSourceSnapshot`, which under `usesPinnedSource` runs `git ls-files --error-unmatch`, `git diff --quiet HEAD` and a `Buffer.equals` against the pinned ref. Two independent confirmations that the sentence was wrong, and that the harm was real: a model told probe paths are unchecked cites build output and the node dies with "is not pinned and unchanged".

### The three findings I had NOT already fixed

**The fan-in file still contained the unconditional twin of the sentence I was fixing.** Four lines below my new "omit it entirely" text, the same paragraph said "render the exact ledger IDs under a `ledger_ids` field" with no condition — while the *very next sentence* handles the identical situation correctly with "When present, render the exact `reference_expectations`". I fixed one sentence and left its contradiction in place. That is not harmless: an empty rendered field is tolerated, but a placeholder like `ledger_ids: none` is parsed as a rendered ID and trips `INVARIANT_LEDGER_MARKDOWN_MAPPING_EXTRA`. Now conditioned.

**"A directory you searched" was over-broad.** Both implementations gate on `isDirectory() && !isSymbolicLink()`, and `invariantPathParentsInsideWorkspace` rejects any path reached through a symlinked parent. A symlinked `lib/` is an ordinary thing for a search agent to name. The caveat is now stated explicitly and anchored.

**A pre-existing sentence now contradicted the new one.** "Keep every `source_path` target-relative and use a line range or symbol that can be checked against the checked-out source" sat immediately after text saying a probe path may be absent — and `scan_probes` records have no line-range field at all. Scoped to `entries`.

I also took a reviewer's suggestion to make the guidance *prescriptive* rather than merely permissive: prefer the directory you searched or the absent path itself over a generated or untracked file. That closes the gate mismatch without granting the licence to be careless that the earlier draft implied — which was a fair criticism, and connects directly to #292.

### One finding filed rather than fixed → #301

Both reviewers noticed that the runtime gate and the generated template enforce **different rules** for a file probe: the template requires tracked, unmodified and byte-identical to the pinned commit; the gate requires none of those. So a probe naming an untracked file passes the gate and then kills the node. Same split-brain shape as #289, except the two copies differ in content rather than one being stale. Filed as #301 — found by code reading rather than by a dead run, which is the cheap way to find this class.

### Validation

- anchors match whitespace-normalised text, so a reflow cannot masquerade as a semantic regression, and one asserts the false sentence stays absent
- proven red against unmodified prompts (1 failed / 23 passed), 24/24 with them
- prettier, eslint, `git diff --check` clean; full monorepo suite running
- paragraph rewrapped after every edit

Reviewers should read `66e042e`; `ca90e96` was the reviewed head and three of its sentences have since changed.

Issue #304, opened 2026-08-06
Workspace handoff stages ~200 MB of un-ignored Recon corpus, a likely cause of the sandbox deaths at stateful-invariant-setup


aviggiano:
Measured on R46's live volume while `stateful-invariant-setup` was running. This is the strongest lead yet on the one blocker that no merged fix has touched: three runs have lost their sandbox at this node with `exit_category: "sandbox-exited"` and no error text.

## The measurement

The `stateful-invariant-setup` workspace is **575 MB across 651 files**, and most of it is a handful of generated blobs:

| path | size |
| --- | --- |
| `out/build-info/0b7f82b309e3a0ac.json` | 155.2 MB |
| `recon-corpus/build-snapshot/0b7f82b309e3a0ac.json` | **155.2 MB** |
| `out/build-info/4d7e2dd8ed4e1a6f.json` | 33.0 MB |
| `recon-corpus/build-snapshot/4d7e2dd8ed4e1a6f.json` | **33.0 MB** |
| `recon-corpus/covered.*.html` | ~5 MB each, several |

`recon-corpus/build-snapshot/` is a byte-for-byte duplicate of Foundry's `out/build-info/`.

## Why that matters for the handoff

The Aave v4 target's `.gitignore` ignores `broadcast/`, `node_modules`, `out/` and friends — but **not `recon-corpus/`**. Verified by reading the workspace's own `.gitignore`.

`stageWorkspaceTree` (`packages/runtime/src/workspace-handoff.ts`) enumerates paths with:

```
git ls-files -z --cached --others --exclude-standard -- . :(exclude)<runtime roots>/**
```

`--exclude-standard` drops `out/`, so Foundry's copy is safe. `recon-corpus/` is not ignored and not in `WORKSPACE_RUNTIME_ROOTS` (`.ultrafuzz`, `.smithers`, `node_modules`, `artifacts`), so **every one of those files is listed and then `git add --force`-ed**, and the node's `workspace.patch` output — contract `ultrafuzz/text@1` — has to represent ~200 MB of untracked JSON and HTML as unified diff text.

A 155 MB minified JSON rendered as a diff is every line prefixed with `+`, held as a JavaScript string. V8's maximum string length is around 512 MB and heap pressure arrives well before that. A process killed by the OOM killer writes no JS error, which is exactly the signature observed: sandbox gone, `last_error` empty, `worker.log` showing only heartbeats.

## Confidence

[...]

PR #305, opened 2026-08-06
fix(runtime): keep harness-generated corpus out of the workspace patch

aviggiano:
## Summary

- refs #304
- the best-evidenced lead on the one blocker no merged fix has touched: **four sandbox deaths at `stateful-invariant-setup`** across R45 and R46, all `exit_category: "sandbox-exited"` with an empty `last_error`
- correct on its own merits regardless of whether it is the cause

## The measurement

Taken from R46's live volume while the node was running — a 575 MB workspace across 651 files:

| path | size |
| --- | --- |
| `out/build-info/0b7f82b309e3a0ac.json` | 155.2 MB |
| `recon-corpus/build-snapshot/0b7f82b309e3a0ac.json` | **155.2 MB** — byte-identical duplicate |
| `out/build-info/4d7e2dd8ed4e1a6f.json` | 33.0 MB |
| `recon-corpus/build-snapshot/4d7e2dd8ed4e1a6f.json` | **33.0 MB** — byte-identical duplicate |
| `recon-corpus/covered.*.html` | ~5 MB each, several |

Then the decisive check, made by reading the workspace's own `.gitignore`: it ignores `broadcast/`, `node_modules` and `out/` — but **not `recon-corpus/`**.

So `--exclude-standard` drops Foundry's copy and keeps Recon's, `recon-corpus` was not in `WORKSPACE_RUNTIME_ROOTS`, and `git add --force` stages roughly 200 MB of untracked JSON and HTML. `workspace.patch` carries contract `ultrafuzz/text@1`, so that has to be represented as unified diff **text** — a 155 MB minified JSON with every line prefixed `+`, held as a JavaScript string, against a V8 string ceiling near 512 MB with heap pressure long before it.

A process killed by the OOM killer writes no JS error. That is exactly the observed signature: sandbox gone, `last_error` empty, `worker.log` containing only heartbeats.

## Confidence, stated plainly

**Leading hypothesis with hard measurements, not a proven cause.** Measured: the sizes, the missing ignore rule, and that staging lists un-ignored untracked files. Not measured: peak RSS, or an exit code — none of the four deaths recorded one, for the reason #302 documents and #303 fixes going forward. If a later run records `137`, that closes the loop.

The exclusion stands on its own either way: `recon-corpus` is produced by the invariant campaign, not authored by the model, and no downstream node consumes it from a patch.

## A design point the tests forced out

The name-based skip below the pathspecs exists to catch a top-level **file** named like an excluded root, which `/**` cannot match. That is right for runtime roots, which are never authored content — but a file literally named `coverage` or `corpus` plausibly is, and generated corpus is always a directory. So generated roots are excluded by pathspec only, and a second regression pins that an authored top-level `coverage`/`corpus` file still reaches the patch. My first draft had them sharing one set, and the test caught it by failing with `EISDIR`.

## Validation

- new regression proven **red before the fix**
- companion regression for the authored-file case
- `workspace-handoff` 15/15, `artifact-gates` 48/48
- typecheck, eslint, prettier clean; full monorepo suite running

## Not included, deliberately

#304 also proposes a **size ceiling on staging with a loud failure**, and I think that is the more valuable half long-term: it would catch the next generated directory nobody thought to exclude, and turn a silent sandbox death into a diagnosable message. It is a behaviour change that can newly fail runs, so it deserves its own PR and its own argument about the threshold rather than riding along with a list of four directory names.

## Ownership note

This is the staging path I fixed in #281/#282. That change made staging work in the presence of ignored roots; it did not consider what staging would happily swallow once it worked.

<h3>Greptile Summary</h3>

The PR separates tracked and untracked workspace enumeration so tracked edits under fuzzing-output roots remain eligible while untracked harness corpus is excluded.
- Adds generated-root exclusions for `recon-corpus`, `echidna`, and `magic`.
- Adds regressions for generated corpus, tracked seed edits, and authored top-level files.

The PR is not yet safe to merge because newly authored untracked files beneath the designated generated-root directories can still be silently omitted from downstream workspace handoffs.

The generated-root pathspecs indiscriminately exclude all untracked descendants, so the previously reported authored-directory data-loss path remains for new files even though tracked edits and top-level authored files are now preserved.

**Files Needing Attention:** packages/runtime/src/workspace-handoff.ts, packages/runtime/test/workspace-handoff.test.ts

| Filename | Overview |
|----------|----------|
| packages/runtime/src/workspace-handoff.ts | Splits tracked and untracked staging enumeration and applies generated-root exclusions only to untracked paths. |
| packages/runtime/test/workspace-handoff.test.ts | Adds coverage for generated output exclusion, tracked corpus edits, and top-level authored files sharing generated-root names. |

</details>

<sub>Reviews (3): Last reviewed commit: ["fix(runtime): correct the mechanism, and..."](https://github.com/monad-developers/ultrafuzz/commit/acf3c739887e97004ffecfe4b165d65844ec0437) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=50964940)</sub>

**Context used:**


aviggiano:
## Both reviews rejected this, and the second one refuted my causal claim by running the code. All findings applied across `699a8f5` and `acf3c73`.

### The mechanism I gave was wrong

I claimed a 200 MB unified diff held as a JavaScript string against V8's ~512 MB ceiling. A reviewer ran the real path against a 40 MB untracked file and got:

```
P4 threw: ENOBUFS  spawnSync git ENOBUFS
```

`runGit` caps output at `MAX_PATCH_BYTES * 2` (32 MB) and `captureWorkspacePatch` refuses anything over 16 MB, so an oversized patch surfaces as a **loud, catchable error** — one this repo has already seen and fixed once (R31, #252). The four silent sandbox deaths cannot be that mechanism, and I had baked the false claim into a doc comment where it would have misled the next reader. That is precisely the prompt-versus-reality error I spent two PRs today fixing elsewhere.

What survives scrutiny is a different step: `git add` hashes and zlib-compresses every enumerated blob **inside the git subprocess, before any of those ceilings apply**. A cgroup OOM there is silent, which matches an empty `last_error` and a `worker.log` containing only heartbeats. The comment now says that, and says explicitly that the attribution is reasoned rather than measured — no exit code has been captured yet.

So: possibly right fix, definitely wrong reason, now stated correctly.

### The silent-data-loss path, found by probe

Applying the exclusions to `--cached` as well as `--others` meant a **tracked** file under an excluded root was dropped from the patch with no error:

```
P3 tracked corpus/seed.txt modified -> ["other.sol"]
```

Because staging runs after `read-tree <baseline>`, the index keeps the baseline blob, and `applyWorkspacePatch` verifies both trees under the same exclusions — so every check passes and the downstream node quietly sees stale content. A committed seed corpus is a real convention in fuzzing repos. Now two `ls-files` calls, unioned: `--cached` with runtime exclusions only, `--others --exclude-standard` with runtime plus generated. Generated corpus is untracked by definition, so nothing intended is lost. A regression pins that an edit to tracked `recon-corpus/seed.txt` survives while a generated blob beside it does not, proven red against the single-listing version.

### The names were guesses, and missed the one that matters most

Both reviewers found it independently. Every invariant prompt runs:

```
recon fuzz . --contract CryticTester ... --corpus-dir echidna --recon-corpus-dir recon-corpus
```

So the Echidna corpus is `echidna/`, written on **every** invariant node, and coverage artifacts land inside it. My first list said `corpus` and `coverage` — which this pipeline never produces — while omitting `echidna/`. `stateful-invariant-coverage` runs `recon fuzz` in a loop, so the fix would have missed the directory it most needed to catch.

Now `["recon-corpus", "echidna", "magic"]`, all evidenced in the prompts (`coverage.md` writes `recon-coverage.json` into `magic/` and runs `covg-eval magic/ echidna/`). I also dropped `crytic-export` and `medusa` from my second draft: I had no evidence either is produced here, and guessing names is exactly what put `corpus` and `coverage` in the first one.

### Confirmed clean, and worth recording

The corpus-loss attack **does not land**. One reviewer traced every consumer: cross-node handoffs are document artifacts, `materializeWorkspacePatchDependencies` is the only restore path, and the snapshot machinery only ever touches `src contracts test tests`. The one reuse mention in `coverage.md:198` is intra-node. Additionally, `MAX_PATCH_BYTES` means no corpus above 16 MB has *ever* survived a handoff — so excluding it converts a hard failure into a clean no-op.

### The deferred size ceiling is now clearly the durable fix

Both reviewers said so and the evidence backs them: this list already missed `echidna/` on its first outing, and the prompts explicitly invite relocating corpus directories, so a root-anchored name list cannot be complete. The ceiling also has to move earlier than `MAX_PATCH_BYTES`, which is enforced *after* the allocation that kills the sandbox. That is a behaviour change deserving its own PR and threshold argument, and #304 now carries the reasoning.

### Validation

- `workspace-handoff` 16/16, `artifact-gates` 48/48
- every mutation kills exactly one test: shrinking the list, reverting the pathspec union, moving generated roots into the name check, and reverting the tracked/untracked split
- typecheck, eslint, prettier clean; full monorepo suite running

PR #306, opened 2026-08-06
fix(runtime): refuse an oversized workspace before git add, not after

aviggiano:
## Summary

- refs #304 — the half of it I deliberately deferred from #305, which both reviewers of that PR argued was the durable fix
- turns a silent sandbox death into a sentence an operator can act on

## Why the existing ceiling is not a safety net

`MAX_PATCH_BYTES` (16 MB) is checked on the **finished patch**. By then `git add` has already hashed and zlib-compressed every enumerated blob inside the git subprocess. A cgroup OOM there kills the container without writing a JS error — which is exactly the signature four sandboxes produced at `stateful-invariant-setup` across two Aave v4 runs:

```
exit_category: "sandbox-exited"
last_error:    (empty)
worker.log:    heartbeats only
```

The workspace measured **575 MB** at the time. The ceiling that exists is enforced after the allocation that kills the process.

## Why the name list in #305 is not sufficient either

It already missed `echidna/` on its first outing — the directory every invariant prompt configures via `--corpus-dir echidna`. Beyond that, `setup.md:122` and `invariant-testing-campaign.md:50` explicitly tell the model to *"adapt corpus directories to existing local conventions"*, and the pathspecs are root-anchored, so a relocated or nested corpus escapes any list. A denylist cannot be complete here by construction.

## The change

`assertStageableWorkspaceSize` sums the enumerated paths and refuses above `MAX_STAGED_BYTES` (256 MB), with a message naming the five largest contributors and their sizes:

```
workspace staging exceeds 268435456 bytes across 651 paths (575 MB);
largest: recon-corpus/build-snapshot/0b7f82b3.json (155 MB), out/build-info/... (155 MB), ...
```

256 MB is deliberately far above any plausible authored source tree — the Aave v4 target's tracked sources are a few megabytes — so it fires only on generated output, and loudly when it does.

A path that vanishes between listing and stat is skipped rather than fatal, matching the race `stageWorkspaceTree` already retries around: agent subprocesses are still running during capture.

## On the test

It stages a 300 MB file under a directory that is **not** on any exclusion list, so only the ceiling can catch it. It also asserts the message names the offending path *with its size* — a refusal that does not say where to look would just be a different kind of dead end, and that is the whole point of preferring this to a crash. Proven red before the fix.

## What this does and does not claim

It does **not** claim to fix the sandbox deaths. My causal story for those was refuted during #305's review — I had claimed a V8 string ceiling, and a reviewer showed `runGit`'s `maxBuffer` makes that unreachable, surfacing as a loud `ENOBUFS` instead. The `git add` OOM reading that replaced it is reasoned, not measured; no exit code has been captured yet, for the reason #302 documents.

What this change does claim is narrower and does not depend on that being right: an oversized staging attempt should fail with a diagnosis rather than however the kernel happens to end the process. If the OOM reading is correct, this prevents the death. If it is wrong, this removes one plausible cause from the list and costs nothing.

## Validation

- workspace-handoff **17/17**, proven red before the fix
- full monorepo suite running

<h3>Greptile Summary</h3>

The PR adds a 256 MB preflight before workspace staging so oversized worktrees fail with diagnostics instead of reaching `git add`, together with a regression test for a 300 MB generated file.

- Enumerates and stats stageable paths before invoking `git add`.
- Reports the five largest contributors when the aggregate exceeds the ceiling.
- Adds coverage for rejecting an oversized, non-excluded workspace file.

<h3>Confidence Score: 3/5</h3>

The PR should not merge until the preflight measures the same path bytes and file contents that `git add` will consume.

Non-UTF-8 path decoding can omit a stageable blob entirely, and concurrent file replacement or growth can invalidate an otherwise successful size snapshot before Git reads the file.

**Files Needing Attention:** packages/runtime/src/workspace-handoff.ts

| Filename | Overview |
|----------|----------|
| packages/runtime/src/workspace-handoff.ts | Adds the pre-git-add staging ceiling, but lossy path decoding and concurrent file mutation can make the measured total differ from the bytes Git stages. |
| packages/runtime/test/workspace-handoff.test.ts | Adds focused oversized-workspace coverage and verifies that the diagnostic identifies the large file. |

</details>

```mermaid
flowchart TD
  A[Enumerate stageable paths] --> B[Stat paths and sum sizes]
  B --> C{Total over 256 MB?}
  C -- Yes --> D[Throw diagnostic with largest paths]
  C -- No --> E[Pass raw path buffers to git add]
  E --> F[Write tree and produce patch]
```
</details>

<a href="https://app.greptile.com/ide/claude-code?prompt=%23%23%23%20Issue%201%0Apackages%2Fruntime%2Fsrc%2Fworkspace-handoff.ts%3A314-319%0A**UTF-8%20decoding%20bypasses%20size%20ceiling**%0A%0AWhen%20a%20stageable%20file%20has%20a%20non-UTF-8%20filename%2C%20decoding%20its%20raw%20path%20inserts%20replacement%20characters%2C%20so%20%60lstatSync%60%20misses%20and%20silently%20excludes%20the%20file%20while%20%60git%20add%60%20receives%20the%20original%20bytes.%20An%20oversized%20blob%20therefore%20reaches%20staging%20without%20contributing%20to%20the%20ceiling%2C%20restoring%20the%20unbounded%20staging%20failure%20this%20check%20is%20intended%20to%20prevent.%0A%0A%23%23%23%20Issue%202%0Apackages%2Fruntime%2Fsrc%2Fworkspace-handoff.ts%3A317-322%0A**Concurrent%20growth%20invalidates%20size%20snapshot**%0A%0AIf%20a%20still-running%20workspace%20subprocess%20grows%20or%20replaces%20a%20file%20after%20%60lstatSync%60%2C%20the%20check%20approves%20the%20old%20size%20while%20%60git%20add%60%20reads%20the%20new%20contents.%20Because%20retries%20cover%20only%20vanished%20paths%2C%20an%20oversized%20file%20bypasses%20the%20ceiling%20and%20can%20cause%20the%20same%20silent%20staging%20termination%20instead%20of%20the%20promised%20diagnostic.%0A%0A---%0A%0AFor%20each%20issue%20above%2C%20determine%20whether%20it%20is%20valid%20and%20should%20be%20fixed.%20If%20so%2C%20fix%20it%20directly.&repo=monad-developers%2Fultrafuzz&pr=306&platform=github"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInClaudeDark.svg?v=6"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInClaude.svg?v=6">Fix All in Claude Code</picture></a>

`````markdown
### Issue 1
packages/runtime/src/workspace-handoff.ts:314-319
**UTF-8 decoding bypasses size ceiling**

When a stageable file has a non-UTF-8 filename, decoding its raw path inserts replacement characters, so `lstatSync` misses and silently excludes the file while `git add` receives the original bytes. An oversized blob therefore reaches staging without contributing to the ceiling, restoring the unbounded staging failure this check is intended to prevent.

### Issue 2
packages/runtime/src/workspace-handoff.ts:317-322
**Concurrent growth invalidates size snapshot**

If a still-running workspace subprocess grows or replaces a file after `lstatSync`, the check approves the old size while `git add` reads the new contents. Because retries cover only vanished paths, an oversized file bypasses the ceiling and can cause the same silent staging termination instead of the promised diagnostic.

---

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(runtime): refuse an oversized worksp..."](https://github.com/monad-developers/ultrafuzz/commit/9197351e3824965d1347eaace7f8edd85d7d0e91) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=50972310)</sub>

> Greptile also left **2 inline comments** on this PR.

**Context used:**

### CODE REVIEW

**Comment by aviggiano** [comment]
## Closing this. The premise is measurably false, and I verified it myself rather than taking the review's word for it.

Both reviewers said NO-MERGE, and the adversarial one refuted the causal claim with a measurement I should have taken before writing a line of this. I reproduced it:

```
$ head -c 155000000 /dev/zero > big{1,2,3,4}.bin   # 620 MB across four blobs
$ /usr/bin/time -v git add -A
        Maximum resident set size (kbytes): 155812
```

**`git add` peak RSS is per-file, not cumulative.** 620 MB across four blobs peaks at the same ~155 MB as one. A reviewer measured the same shape and additionally found a single 700 MB blob peaks at **4.6 MB**, because it streams above `core.bigFileThreshold`.

And the sandbox is not small: `packages/modal/src/defaults.ts:28-32` gives invariant nodes `memoryMiB: 32_768` / `memoryLimitMiB: 65_536`. `git add` on the observed 575 MB workspace peaks at roughly **0.5% of the memory request**. There is no `--max-old-space-size` anywhere to make Node the constrained party either.

So the `git add` OOM story is not merely unmeasured — it is false at these sizes. That is the second time my causal explanation for these four sandbox deaths has been refuted, and this time the disproof cost one `/usr/bin/time -v`. I reasoned where I should have measured.

## The rest of the review stands on its own

Even with a sound premise this PR should not merge as written:

- **Wrong metric.** Peak memory is `max(size)` clamped by `core.bigFileThreshold`, not `sum(size)`. The ceiling refuses 620 MB across four files (163 MB peak) and 700 MB in one file (4.6 MB peak), while allowing 250 MB in one file (250 MB peak) — the most expensive of the three.
- **It breaks working runs.** Tracked, *unmodified* content counts fully toward the ceiling while contributing zero patch bytes. Demonstrated with a fixture: `origin/main` produces a 140-byte patch, this branch throws.
- **The blast radius is the whole run, not one node.** The check lands in `captureWorkspaceTree`, which previously had no size ceiling at all and runs during task *preparation* for every node. One polluted workspace would fail every downstream node before its agent starts — and the agent that could delete the offending directory never runs. No env override, no escape hatch; recovery would need a source edit and an image rebuild.
- **It would not have fired on the incident it cites.** Post-#305, R46's enumerated set is ~179 MB, under the 256 MB ceiling.
- **My path handling violated the file's own documented invariant.** `stageableWorkspacePaths` says explicitly that paths stay as bytes because a non-UTF-8 filename decoded through a string matches nothing; my `entry.toString("utf8")` did exactly that, and `catch { continue }` swallowed the resulting `lstat` failure. A reviewer showed a 300 MB file with a `0xFF 0xFE` name undercounting by the full 300 MB.

## What I am doing instead

1. **Correcting the false claim already merged.** #305's doc comment attributes the deaths to a `git add` OOM. That is now disproven and sits in `main`, which is precisely the prompt-versus-reality defect I spent two PRs fixing today. A follow-up removes it and records the measurement.
2. **Updating #304** with the RSS numbers and the sandbox limits, so nobody re-derives this theory.
3. **Leaving the cause open.** Reviewer 1's point is the right one: a 575 MB workspace at a long-running invariant node is at least as plausibly a *symptom* of a long campaign as a cause of death, and the real killer — echidna workers, `forge build`, something else entirely — is untouched by anything here. R47's overseer is healthy at generation 0, so if its sandbox dies a replacement is reserved and #303 records the exit code. That measurement is worth more than another guess.

#305 still stands on its own merits: harness-generated corpus is not authored content, no downstream node consumes it from a patch, and `MAX_PATCH_BYTES` meant no corpus above 16 MB ever survived a handoff anyway. It just is not the fix for these deaths, and neither is this.

PR #314, opened 2026-08-07
fix(runtime): start dependency replay after the prefix the worktree already holds

### LINKED ISSUE #312: applyWorkspacePatch base tree mismatch blocks stateful-invariant-implement-properties, with maxAttempts=1 so it kills the run

aviggiano:
Fixes #312.

## What breaks

`prepare:stateful-invariant-implement-properties` fans in every dependency's workspace patch and replays them. That is correct on a **fresh** run: the task worktree starts at the pinned baseline, so each patch's declared `base_tree` is satisfied in turn down the chain. A **resumed** run breaks the precondition, because the worktree lives on a durable volume and still holds the previous attempt's state.

Two production runs died at this node with the same expected tree:

| run | image | error |
| --- | --- | --- |
| R48 | `62aaef1` | `expected 2dd4efef…, got bf324c39…` |
| R49 | `51e6a7b` | `expected 2dd4efef…, got b8d46f13…` |

R49 is the sharper evidence: it ran **24 nodes fresh without touching this bug**, its sandbox exited, the overseer recovered it with `action: launch`, a new worker came up 90 seconds later with `resume=true, restoredTaskCount=48`, and it was dead within 90 seconds of resuming. Its overseer then exhausted its whole recovery budget — three further resumes, three identical deaths.

R48's dependency manifests, dumped from its volume:

| node | base_tree | result_tree | files |
| --- | --- | --- | --- |
| `setup-foundry` | **2dd4efef** | 4fefa8df | 1 |
| `base-test-setup` | 4fefa8df | 4fefa8df | 0 |
| `stateful-invariant-setup` | 4fefa8df | fd4e3c2c | 7 |
| `stateful-invariant-handlers` | fd4e3c2c | bf324c39 | 7 |
| `stateful-invariant-coverage` | bf324c39 | bf324c39 | 0 |

The worktree sat at the **end** of that chain. Every dependency's content was already present, folded in by the later links; `applyWorkspacePatch` threw only because it compares the worktree against one patch's own `base_tree` in isolation. R49's manifests were never dumped (the volume walk timed out), so its worktree being at the end of *its* chain is a hypothesis, not a measurement.

## The rule

Replay starts after the prefix the worktree already holds. Two conditions are required, and **only the first is a hash identity** — an earlier revision of this PR claimed both were, and review falsified it:

1. The worktree's tree equals dependency *i*'s declared `result_tree`. A tree id is a content hash, so the workspace is identical to that dependency's output over the snapshot the hash covers. Content outside that snapshot is content no patch can carry either, because `captureWorkspacePatch` diffs the same staged index — so nothing patch-delivered is missed.
2. The dependencies before *i* chain into it, each one's `result_tree` being the next one's `base_tree`. Without this, "everything before *i* is materialized" is a claim about **topology**, not content.

### Why condition 2 is not theoretical

Review reproduced the corruption end to end against real git, on a **fresh** run — no resume required:

```
setup-foundry-0:  base 32263f5d -> result e1507a29   (writes the scaffold)
setup-foundry-1:  base 32263f5d -> result 32263f5d   (agent changed nothing)
worktree:         32263f5d                            (the pinned baseline)

origin/main : 2 patches applied, loud base-tree mismatch
first revision of this PR : 0 patches applied, NO error, empty workspace
```

A partial skip is self-validating — the next `applyWorkspacePatch` re-checks `base_tree` and throws. A **total** skip validates nothing, and a sibling fan-in produces exactly that. It is one line of topology away: the `strategies` group defaults to `loops: 3, loop_mode: parallel`, and every patch publisher is held to a single copy only by an explicit `loops: 1`, with nothing validating that a `workspace-patch@1` publisher may not fan out.

When the prefix is not a chain, replay starts at 0 and `applyWorkspacePatch` raises its mismatch exactly as today. Failing the way we already fail is the safe direction; silent garbage findings are far worse than a crash.

## Validation no longer depends on applying

`validateWorkspacePatchCapture` is split out of `applyWorkspacePatch` and called for **every** capture, skipped or not. The manifest schema, object ids, patch digest, symlink/submodule rejection, sensitive-path rejection and the base-commit check all used to live behind the apply, so a skipped patch went entirely unvalidated — while the skip decision read `result_tree` from a manifest nothing had checked was even well formed. A re-pinned target between attempts would also have slipped past a skip.

## Scope

Replay path only. Post-agent preparation (`replayWorkspacePatches: false`) keeps its own rule and is not second-guessed, with a test that discriminates the two.

## Tests

13 tests. Five cover the rule; the rest lift `materializeWorkspacePatchDependencies` out of the template and run it against stubbed collaborators, because the rule alone was not enough — deleting the wiring left the rule tests green.

The stubbed worktree tree **advances** as patches land, as it does on a real run. A frozen stub had hidden a surviving mutation (hoisting the per-iteration tree read out of the post-agent loop) and had also made me assert a post-agent result of two patches where the real path applies four.

Mutations, each failing at least one test: `replayFrom` forced to 0; the replay-path guard dropped; chain scanned forward; return matched index instead of the one after; compare `base_tree` instead of `result_tree`; the chain-linearity check removed; the validate-all loop removed; the dependency sort comparator neutered; the per-iteration tree read hoisted.

Full runtime suite: **371 tests, 356 pass, 0 fail.**

<h3>Greptile Summary</h3>

The PR makes workspace-patch replay resume after the validated linear dependency prefix already represented by a durable worktree.
- Extracts capture-only validation so skipped patches remain subject to manifest, digest, object-ID, path-safety, and base-commit checks.
- Preserves the separate post-agent materialization behavior.
- Adds focused tests for fresh, resumed, partial, repeated-tree, fan-in, validation, ordering, and post-agent scenarios.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| packages/runtime/src/templates/smithers/workflows/workflow.tsx | Adds validated, chain-aware replay-prefix selection while retaining the distinct post-agent preparation path. |
| packages/runtime/src/workspace-handoff.ts | Extracts capture-only validation and keeps worktree-dependent checks inside patch application. |
| packages/runtime/test/workspace-patch-replay.test.ts | Adds focused regression and wiring coverage for replay selection, fan-in safety, validation, ordering, and state advancement. |
| packages/runtime/test/generated-workflow-verifier.test.ts | Verifies that the generated workflow retains the replay guard, chain check, validation pass, and replay slicing. |

</details>

```mermaid
flowchart TD
  A[Read and sort dependency captures] --> B[Validate every capture]
  B --> C{Replay path?}
  C -- No --> D[Use post-agent materialization rule]
  C -- Yes --> E[Capture current worktree tree]
  E --> F[Scan matching result trees backward]
  F --> G{Preceding manifests form a linear chain?}
  G -- No --> H[Replay from dependency zero]
  G -- Yes --> I[Replay after matched prefix]
  H --> J[applyWorkspacePatch validates base tree]
  I --> J
```
</details>

<sub>Reviews (3): Last reviewed commit: ["docs(runtime): scope the skipped-capture..."](https://github.com/monad-developers/ultrafuzz/commit/823787ae524ceaf2c3f5aebbd685233386cad270) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=51073115)</sub>

**Context used:**


aviggiano:
## Independent verification of the fresh-run corruption fix

I rebuilt the reported reproduction from scratch — real git repo, real `captureWorkspacePatch` / `captureWorkspaceTree` / `applyWorkspacePatch`, and the production `materializeWorkspacePatchDependencies` lifted from each revision's own template — and ran the same input against three revisions.

The fixture is the reported one, built rather than asserted: sibling A writes `test/recon/Setup.sol`, sibling B's agent changes nothing so it publishes a zero-file patch, both declare the same base, and the worktree starts at the **pinned baseline** — a fresh run, no resume.

```
A: base cdad1e09 -> result d8be5348 (1 file)
B: base cdad1e09 -> result cdad1e09 (0 files)
```

| revision | error | `test/recon/Setup.sol` afterwards | verdict |
| --- | --- | --- | --- |
| `origin/main` | `workspace patch base tree mismatch` | present | safe |
| `ea0085c` (first revision) | **none** | **absent** | **silent corruption** |
| `dd9df13` (this head) | `workspace patch base tree mismatch` | present | safe |

Two things this establishes that a passing test alone would not:

1. **The harness can fail.** It reproduces the corruption on `ea0085c`, so the clean result on `dd9df13` is evidence rather than an absence of evidence.
2. **The fix restores `origin/main`'s behaviour exactly** on this input — sibling A's content is applied and sibling B raises the pre-existing loud mismatch. The change does not trade the crash for a quieter, worse failure; it leaves the non-chain case exactly as loud as it was.

Verification script kept at `/tmp/verify312.mjs`.

Issue #328, opened 2026-08-07
implemented-properties rejects an empty OPTIONAL array, and the contract error discards the field path that would have said so


aviggiano:
R51 (`aave-v4-v0012-main-issue315fix-r51-invariant-only`, image from merged main `fa47b07`) got **past** the prepare step that killed R48, R49 and R50 — #312 and #315 are doing their job — and failed inside `stateful-invariant-implement-properties` itself, **twice, deterministically**:

```
artifact-contract failure for implemented-properties.json (ultrafuzz/implemented-properties@2):
Too small: expected array to have >=1 items      [x88, no field path]
```

Attempt 1 ran 13:42:11 → 14:06:04 (24 min), attempt 2 ran 14:13:59 → 14:19:31. Identical error both times.

### 1. An OPTIONAL array whose schema forbids `[]`

`referenceExpectationIdsSchema` is `z.array(nonEmptyString).min(1)` (`packages/artifacts/src/property-provenance.ts:22-24`), used as `reference_expectations` — an **optional** field on `implementedPropertySchema` (`:257`). So omitting it is valid and emitting it empty is not:

```
reference_expectations OMITTED   -> VALID
reference_expectations: []       -> Too small: expected array to have >=1 items @ properties.0.reference_expectations
reference_expectations: ['x']    -> VALID
89 properties with []            -> 89 issues, all the same message
```

R51's document (173,510 bytes, recovered from the task workspace on the durable volume) has **89 properties, every one carrying `reference_expectations: []`** — which is exactly the 88 repetitions plus the header.

An empty optional list is a natural thing for a producer to emit and is semantically identical to omitting the field. The schema's own design supports that reading: `status` is `implemented | pending | deferred | blocked`, there is a `blocker` object, and the `superRefine` requires source paths only when `status === "implemented"`. A document reporting properties it has *not* implemented is clearly anticipated — but it cannot include the optional expectations list as empty while doing so.

### 2. The contract error discards the field path it already has

`validateWithZod` computes a path for every issue (`packages/artifacts/src/schema-validation.ts:29-31`, `path: zodPath(issue.path, ...)`). The failure message then drops it (`packages/runtime/src/templates/smithers/workflows/workflow.tsx:4068`):

```ts
`artifact-contract failure for ${output.path} (${output.contract}): ${validation.issues
  .map((issue) => issue.message)
  .join("; ")}`
```

[...]

PR #331, opened 2026-08-07
fix(artifacts): accept an empty reference-expectation list on an implementation record

### LINKED ISSUE #330: Implementation records must omit reference_expectations, but the prompt tells the model to always emit it
## Summary

`stateful-invariant-implement-properties` cannot succeed on a target where any selected property has no reference expectations. The prompt tells the model to preserve each selected property's `reference_expectations` array; the artifact contract rejects that array when it is empty. The model does what it is told, the gate refuses the result, and the node fails with `executor-error`.

This killed the R51 Aave v4 invariant-only run today. Four sandbox deaths on this one node, then the recovery budget ran out and the app dropped to zero tasks with 25 of 32 nodes already green.

## Evidence

The failure is recorded — `executor-error` comes from a `NodeFailed` workflow event, and that event carries an error payload that the worker's top-level catch never surfaces (#307). From `smithers/logs/stream.ndjson`:

```
artifact-contract failure for implemented-properties.json (ultrafuzz/implemented-properties@2):
Too small: expected array to have >=1 items; Too small: expected array to have >=1 items; …
```

The artifact the node produced is otherwise complete and reasonable — 161 KB, a valid `selection` block, 89 property records, 60 `implemented` / 25 `deferred` / 4 `blocked`. The census of empty arrays across those 89 records:

| field | empty on |
|---|---|
| `reference_expectations` | **89 of 89** |
| `test_paths` | 89 of 89 |
| `implementation_paths` | 29 of 89 |

Only the first of those is fatal.

## Root cause

`packages/artifacts/src/property-provenance.ts`:

```ts
const referenceExpectationIdsSchema = z.array(nonEmptyString).min(1)…   // :22-24
…
reference_expectations: referenceExpectationIdsSchema.optional(),      // implementedPropertySchema
```

`optional()` combined with `.min(1)` means **absent is valid and empty is invalid**. A property with no reference expectations must omit the key.

`.ultrafuzz/prompts/strategies/invariants/implement-properties.md` says the opposite, at `:50`:

> Preserve each selected property's complete `reference_expectations` array in the implementation summary and use it to explain any blocker.

unconditionally, and again at `:171`. Two sentences earlier the same paragraph establishes the opposite convention for the neighbouring fields:

> Include both path arrays on every record, using empty arrays when no path exists.

So the model emitted `reference_expectations: []` on every record that had no expectations, which is the natural reading of "preserve the complete array" next to an explicit empty-array instruction.

### Not the cause — recording so the next person does not chase it

All 89 records also carry `test_paths: []`, and 29 carry `implementation_paths: []`. Those are **fine**. `nonEmptyStringArray` is `z.array(nonEmptyString)` with no `.min(1)` — it means "array of non-empty strings", not "non-empty array". The name is misleading enough that it cost me a wrong first hypothesis.

## This is a known defect class that missed this node

#300 (`docs(prompts): say that optional evidence fields are optional`, fixing #291/#299, alongside gate fixes #290/#298) states it exactly:

> each of these prompts told the model to do something a gate then rejected, and each cost a live Aave v4 run a full agentic node attempt

That work covered fan-in `ledger_ids`. `stateful-invariant-implement-properties` and `reference_expectations` are the same defect, one node over, and it has now cost a full run rather than a single attempt.

## Why retrying cannot clear it

It is deterministic, not flaky. Four consecutive attempts, identical contract failure:

| attempt | window | outcome |
|---|---|---|
| 1 | 13:42:11 → 14:06:04 | `failed` / `executor-error` |
| 2 | 14:13:59 → 14:19:31 | `failed` / `executor-error` |
| 3 | 14:35 → 14:49 | `failed` / `executor-error` |
| 4 | → 15:10:41 | worker terminated, exit 1 |

The prompt does not change between attempts, so neither does the model's choice. Recovery budget is spent relaunching into the same wall.

## Fix

Two halves, mirroring how #298 and #300 split the equivalent `ledger_ids` fix:

1. **Contract** — accept an empty `reference_expectations` array on an implementation record and treat it as "none", keeping the duplicate-ID check and leaving the catalog-side `.min(1)` contract untouched. An implementation record for a property with no expectations has nothing to carry; failing a node over absent-versus-empty is not a distinction worth a sandbox.
2. **Prompt** — make `:50` conditional, say explicitly that a property with no reference expectations omits the field, and scope the neighbouring empty-array convention to the two path arrays so the two instructions stop reading as one rule.

Fixing only the prompt would leave every already-built image failing the same way; fixing only the contract would leave the prompt still saying something the schema merely tolerates. Both.

## Note for #307

The error payload for this class **is** recoverable from `smithers/logs/stream.ndjson`, even though `worker.log` and the app log show only `worker terminated` / exit 1. That does not reduce the case for threading the cause — it took reading the classifier source to learn that `executor-error` implies a `NodeFailed` payload exists, and 16 MB of NDJSON to find it — but it is a cheaper diagnostic path than the issue currently assumes, and worth recording there.

aviggiano:
## Summary

- fixes #330
- accept an empty `reference_expectations` array on an implementation record and treat it as "none"
- make the implement-properties prompt say to omit the field rather than emit an empty array

This is the same split as #298 (gate) and #300 (prompt) applied to the equivalent `ledger_ids` defect, one node over.

## What broke

`stateful-invariant-implement-properties` failed four consecutive attempts on the R51 Aave v4 invariant-only run today, each with `executor-error`, and the run died with 25 of 32 nodes green when the recovery budget ran out.

The node's output was complete and reasonable — 89 property records, 60 `implemented` / 25 `deferred` / 4 `blocked`, valid `selection` block. Every one of the 89 carried `reference_expectations: []`, and the contract rejected all 89:

```
artifact-contract failure for implemented-properties.json (ultrafuzz/implemented-properties@2):
Too small: expected array to have >=1 items; …
```

`reference_expectations` was `referenceExpectationIdsSchema.optional()` where that schema carries `.min(1)` — absent valid, empty invalid — while the prompt said, unconditionally:

> Preserve each selected property's complete `reference_expectations` array

two sentences after establishing the opposite convention for the fields beside it:

> Include both path arrays on every record, using empty arrays when no path exists.

## What this changes

`implementedReferenceExpectationIdsSchema` drops the non-empty bound for implementation records only, keeping the duplicate-ID refinement. The catalog-side `referenceExpectationIdsSchema` is untouched, so `lensPropertySchema` and the reference-expectations artifact keep requiring a non-empty list — a catalog entry that declares expectations should still declare at least one.

The refinement body is duplicated rather than shared so neither schema can later constrain the other by accident.

Prompt: `:50` becomes conditional and says to omit the field; the paragraph at `:171` scopes the empty-array convention explicitly to the two path arrays.

Fixing only the prompt would leave every already-built image failing the same way. Fixing only the contract would leave the prompt asserting something the schema merely tolerates.

## Tests

Added `implementation records accept an absent or empty reference expectation list` to `packages/artifacts/test/schema.test.ts`, covering absent, empty, populated, and duplicate-rejected.

**Not run locally.** The machine I worked from has no Node, pnpm or bun, so I could not execute vitest or the build. CI is the first execution of both the new test and the changed schema — please treat the green tick as the actual verification rather than my say-so.

## Follow-up not in this PR

The R51 run is dead and needs relaunching as R52 from a build that includes this commit; the fix does not resume it.

<h3>Greptile Summary</h3>

This PR allows implementation records to represent no reference expectations as either an omitted field or an empty list while retaining duplicate-ID validation.
- Keeps canonical property catalogs constrained to non-empty expectation lists when the field is present.
- Updates the implementation prompt to prefer omission when no expectations exist.
- Adds coverage for absent, empty, populated, and duplicate expectation lists.

The PR appears safe to merge, with the schema, prompt, runtime semantics, and tests aligned around treating absent and empty implementation lists equivalently.

The relaxed schema is limited to implementation records, duplicate detection remains intact, canonical catalog constraints are unchanged, and downstream validation already normalizes absent and empty lists to the same value.

| Filename | Overview |
|----------|----------|
| .ultrafuzz/prompts/strategies/invariants/implement-properties.md | Clarifies that only path fields use empty-array placeholders and that reference_expectations should be omitted when absent. |
| packages/artifacts/src/property-provenance.ts | Adds an implementation-specific reference-expectation schema that accepts empty arrays without weakening canonical catalog validation or duplicate rejection. |
| packages/artifacts/test/schema.test.ts | Covers absent, empty, populated, and duplicate reference-expectation lists on implementation records. |

</details>

<sub>Reviews (1): Last reviewed commit: ["style(artifacts): collapse the two-call ..."](https://github.com/monad-developers/ultrafuzz/commit/0137ac0d5671dab5a1cda1e272e122cd7f8a8fe2) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=51207312)</sub>

**Context used:**


aviggiano:
Superseded by #329, which landed the same diagnosis and a strictly better fix while this was open.

#329 normalises `reference_expectations` at **all three** sites via `optionalReferenceExpectationIds`; this PR only relaxed the implemented-properties usage, so it would have left the same trap in the other two. #329 also names the failing field in the contract error, which is the part that made this cost hours to find.

The one piece not covered there is the prompt half: `.ultrafuzz/prompts/strategies/invariants/implement-properties.md:50` still says to preserve each selected property's complete `reference_expectations` array unconditionally, two sentences after establishing an explicit empty-array convention for the neighbouring path arrays. That is no longer fatal now that the contract accepts `[]`, but it is the same prompt-and-gate divergence #300 set out to remove. Noted on #328 rather than carried here.

Closing rather than rebasing — the schema change is redundant and conflicts.

PR #358, opened 2026-08-07
Let a re-executed node replace its own superseded workspace patch

### LINKED ISSUE #357: A re-executed node's own prior-generation workspace.patch is rejected as tampering
## Symptom

R52 (`aave-v4-v0012-main-issue328fix-r52-invariant-only`, invariant-only, n=1) went terminal with
`terminal-run-non-resumable`, nodes succeeded=25 failed=2 remaining=5. The whole invariant chain
succeeded through `stateful-invariant-setup`, `-handlers`, `-implement-properties`, `-coverage`. The
single node blocking `triage` -> `dedupe-findings` -> `severity-classification` -> `final-report` is:

```
stateful-invariant-campaign  failed  retry_count=1
last_error: artifact-contract failure: workspace patch artifact was modified workspace.patch
```

## Mechanism, from mtimes on the volume

The campaign node ran twice, in two recovery generations:

- generation 0: 18:37:07Z -> 18:52:34Z
- generation 1: 19:00:54Z -> 19:15:48Z (state.json `started_at`/`finished_at`)

`artifacts/stateful-invariant-campaign/` on the volume:

```
    26001  2026-08-07T19:15:46Z  invariant-suite-baseline.json   <- generation 1 wrote this
     1773  2026-08-07T19:15:46Z  invariant-suite-manifest.json   <- generation 1
      161  2026-08-07T18:37:07Z  workspace-patch-baseline.json   <- generation 0 preparation
      167  2026-08-07T18:37:07Z  workspace-patch-preparation.json<- generation 0 preparation
      435  2026-08-07T18:52:34Z  workspace-patch.json            <- generation 0 output
     1141  2026-08-07T18:52:34Z  workspace.patch                 <- generation 0 output
```

Generation 1 re-executed the node, wrote its fresh `invariant-suite*` artifacts at 19:15:46, then
`materializeWorkspacePatch` -> `writeWorkspacePatchArtifact` found generation 0's own `workspace.patch`
from 18:52:34 still on the volume with different bytes and threw:

`packages/runtime/src/templates/smithers/workflows/workflow.tsx:817-827`

```ts
if (existsSync(target)) {
  const existing = resolveRegularArtifactFile(...);
  const existingContents = readFileSync(existing, "utf8");
  if (existingContents !== "" && existingContents !== "\n") {
    if (existingContents === contents) return;
    throw new Error(`artifact-contract failure: workspace patch artifact was modified ${relativePath}`);
  }
}
// These paths are runtime-owned. Replace only an empty runtime placeholder;
// reject any non-empty agent-authored or tampered patch above.
```

The guard cannot distinguish **"my own output from a previous generation of this same node"** from
**"an agent tampered with this file"**. Both present as non-empty-and-different. Because the node is
re-executed rather than retried in place, the artifact root is not emptied, so the stale file survives.
`retry_count=1` shows the in-generation retry hit the identical wall.

This is the same class as #312 and #316: a legitimately re-executed producer's earlier artifact read as
drift/conflict. Notably `captureInvariantSuiteBaseline` in the same file already reasons about exactly
this ("a legitimately re-executed ancestor ... as tampering"), which is why `invariant-suite-baseline.json`
refreshed cleanly at 19:15:46 while `workspace.patch` did not. The staleness concept exists in the file;
`writeWorkspacePatchArtifact` just does not participate in it.

Secondary observation from the same mtimes: `workspace-patch-baseline.json` is also generation 0's
(18:37:07), so generation 1's capture is taken against a generation-0 baseline. That is worth deciding
on explicitly rather than inheriting by accident.

## Why this is not "just widen the check"

The guard is load-bearing — it is what stops an agent from authoring or editing a runtime-owned patch.
Any fix has to keep rejecting agent-authored content while accepting the node's own superseded output.
The discriminator must be provenance (which generation/attempt wrote it), not content equality.

## Candidate fixes, not presuming which

1. **Treat the node's own prior-generation artifact as stale and overwrite it**, keyed off the recorded
   producing attempt/generation the way the invariant-suite baseline already is. Fixes the root cause;
   requires that the producing identity actually be recorded where this code can read it.
2. **Empty the runtime-owned patch artifacts when a node is re-executed in a new generation**, so the
   placeholder path is taken. Smaller blast radius, but it moves the fix away from the guard and into
   re-execution setup, where a future caller can forget it.

Option 1 keeps the invariant local to the guard that enforces it, but only if provenance is genuinely
available at that point — I have not yet confirmed it is, so I am not committing to it in this issue.

## Repro

Re-execute a `taskPublishesWorkspacePatch` node in a new recovery generation, against an artifact root
that still holds a differing `workspace.patch` from the prior generation.

aviggiano:
Closes #357.

## What R52 hit

`stateful-invariant-campaign` failed with

```
artifact-contract failure: workspace patch artifact was modified workspace.patch
```

and it was the **only** thing between that run and a `report.md`. Node counts at terminal were
succeeded=25 failed=2 remaining=5; the invariant chain had already succeeded through
`stateful-invariant-setup`, `-handlers`, `-implement-properties` and `-coverage`, so this failure alone
left `triage` -> `dedupe-findings` -> `severity-classification` -> `final-report` unreachable. (The other
failed node, `property-specification-fanin`, is #285 and only blocks `aggregate-test-files`.)

## Mechanism

The node ran in two recovery generations — 18:37:07Z→18:52:34Z, then 19:00:54Z→19:15:48Z. The mtimes in
its artifact directory on the volume name the mechanism exactly:

```
26001  19:15:46Z  invariant-suite-baseline.json      <- generation 1 wrote this
 1773  19:15:46Z  invariant-suite-manifest.json      <- generation 1
  161  18:37:07Z  workspace-patch-baseline.json      <- generation 0 preparation
  435  18:52:34Z  workspace-patch.json               <- generation 0 OUTPUT
 1141  18:52:34Z  workspace.patch                    <- generation 0 OUTPUT
```

Generation 1 re-ran the agent, wrote its own `invariant-suite*` artifacts, captured a patch that
legitimately differed from generation 0's, and `writeWorkspacePatchArtifact` found the survivor
non-empty and unequal and threw. There is no second chance: the run log records `attempt=1
maxAttempts=1`, so `resetTaskArtifactsForRetry` never runs — and the pair is not in the set that reset
preserves anyway.

Same class as #312 and #316: a legitimately re-executed producer's earlier artifact read as drift.
`captureInvariantSuiteBaseline` in this same file already reasons about exactly this, which is why
`invariant-suite-baseline.json` refreshed cleanly while `workspace.patch` did not.

## The fix, and what it deliberately does not do

The guard is load-bearing — it is what keeps agent-authored content out of a runtime-owned artifact —
so it stays. What changes is the discriminator. Content equality cannot work, because it is precisely
what conflates "my own superseded output" with "someone edited this".

The manifest can: `captureWorkspacePatch` records `patch_sha256` over the patch it was taken with, so
only a pair the runtime published is self-consistent. `discardSupersededWorkspacePatchArtifacts`
removes such a pair before publishing, and leaves every other shape in place for the existing
rejection: a lone patch, a mismatched digest, an unreadable manifest, a foreign schema version.

An agent that forged a self-consistent pair would have it discarded and overwritten by this node's own
capture, so forging one gains nothing — the runtime's bytes still win. That is the honest statement of
the trade-off: the change converts one narrow case (agent writes a *valid* runtime-shaped pair) from
"fail the run" to "overwrite it", and every other tampering shape still fails loudly.

The discard happens before either half of the new pair is written, because once the patch is replaced
the manifest beside it no longer describes it and the pair could not be recognised.

## Evidence

The R52 shape was **first reproduced against unmodified main**, failing with the production error from
the same `writeWorkspacePatchArtifact` → `materializeWorkspacePatch` stack as the run log:

```
✖ R52 reproduction: generation 1 cannot publish over generation 0's own patch
  Error: artifact-contract failure: workspace patch artifact was modified workspace.patch
      at writeWorkspacePatchArtifact (...)
      at materializeWorkspacePatch (...)
```

Tests drive the real template functions over a real temp directory rather than stubs, and lift
`materializeWorkspacePatch` itself so the **call** is pinned — the sibling #312 suite records that
dropping such a call left all of its rule-level tests green.

All four guards are mutation-tested; each mutation was verified to actually apply before trusting its
result:

| mutation | suite |
|---|---|
| baseline | 7 pass, 0 fail |
| drop the `patch_sha256` match | 6 pass, **1 fail** |
| drop the `schema_version` check | 6 pass, **1 fail** |
| drop the manifest-required check | 0 pass, **7 fail** |
| drop the `discardSuperseded...` call | 6 pass, **1 fail** |

- `pnpm --filter @ultrafuzz/runtime test`: 426 tests, 411 pass, 0 fail, 15 skipped; bun lane 14 pass, 0 fail
- `pnpm --filter @ultrafuzz/runtime typecheck`: clean
- `pnpm lint`: clean

<h3>Greptile Summary</h3>

The PR allows a re-executed workflow node to replace its own valid, superseded workspace-patch pair while retaining rejection for malformed, mismatched, foreign, or incomplete pairs.
- Classifies an existing pair using schema, baseline identity, commit identity, and patch digest validation before replacement.
- Applies classification independently to every artifact root.
- Adds integration-style coverage for replacement, tampering guards, multiple roots, unsafe files, and durable publication.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| packages/runtime/src/templates/smithers/workflows/workflow.tsx | Adds validation-based classification and replacement of a node's superseded workspace-patch artifacts while preserving existing rejection paths. |
| packages/runtime/test/generated-workflow-verifier.test.ts | Updates template-level assertions to pin the replacement flag's reject-by-default behavior. |
| packages/runtime/test/workspace-patch-supersede.test.ts | Adds comprehensive real-filesystem and Git-backed coverage for superseded-pair replacement and tamper rejection. |

</details>

```mermaid
flowchart TD
  A[Re-executed node captures fresh workspace patch] --> B[Inspect surviving patch and manifest]
  B --> C{Both files exist?}
  C -- No --> R[Retain existing rejection behavior]
  C -- Yes --> D{Manifest uses current baseline and validates against patch}
  D -- No --> R
  D -- Yes --> E[Classify pair as superseded output]
  E --> F[Durably replace workspace.patch]
  F --> G[Durably replace workspace-patch.json]
  G --> H[Publish fresh runtime capture]
```
</details>

<sub>Reviews (3): Last reviewed commit: ["fix(runtime): classify the superseded pa..."](https://github.com/monad-developers/ultrafuzz/commit/50573dd8d0ec4740891d77c043a3b60174e3008c) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=51284702)</sub>

**Context used:**


aviggiano:
Correcting an overclaim in my own PR description before review acts on it.

I wrote that this failure "was the **only** thing between that run and a `report.md`". That is not
supported by the evidence, and I should not have stated it.

The campaign node's declared outputs are `campaign-plan.json` (primary, `ultrafuzz/json-object@1`) and
`campaign-report.md` (`ultrafuzz/nonempty-markdown@1`). **Neither is in its artifact directory**, after
two full agent runs:

```
invariant-suite/                   19:15:46Z
invariant-suite-baseline.json      19:15:46Z
invariant-suite-manifest.json      19:15:46Z
prompt.rendered.md                 15:37:08Z
workspace-patch-baseline.json      18:37:07Z
workspace-patch-preparation.json   18:37:07Z
workspace-patch.json               18:52:34Z
workspace.patch                    18:52:34Z
```

For comparison, `stateful-invariant-coverage`, which succeeded, does hold its declared outputs in the
same canonical location, so absence here is meaningful rather than an artefact of where I looked. I also
checked for a mirrored artifact root under `workspace/target/artifacts` and it does not exist.

What the mtimes actually support is a more interesting story than the one I told:

- **Generation 0** got *past* this code — `workspace.patch` is dated 18:52:34, so
  `materializeWorkspacePatch` succeeded — and then failed later, at output verification, with its
  required outputs missing.
- **Generation 1** re-executed the node, ran the agent for ~15 minutes, and threw in
  `materializeWorkspacePatch` on generation 0's surviving patch, which is *earlier* in
  `artifactAwareAgent.generate` than `verifyArtifacts`.

So this defect **masks** the node's original failure rather than being the failure. That does not make
the fix less necessary — while it stands, the node dies before its outputs are ever verified, and no
generation can ever reveal or repair the real problem. But the honest claim is narrower than the one I
made: fixing #357 is required for the node to get a fair verification, and whether it then succeeds is
unknown and is separate work.

Everything else in the description — the mechanism, the reproduction against unmodified main, the
mutation results, the suite numbers — stands as written. I have not changed the diff.

aviggiano:
Both reviews are in, both asked for the same two substantive changes, and both were right. Pushed as
`50573dd`. Correcting my evidence first, because one row of my mutation table was wrong.

## The mutation table row was bogus, and here is why

I reported "drop the manifest-required check → **0 pass, 7 fail**". Both reviewers independently got
**5 pass / 2 fail** and flagged it. They are correct. I re-ran it and got 5/2.

The cause: my mutation loop parsed each spec with `IFS='|' read`, and the pattern I was substituting
contained `||`. The separator collided with the code under test, so the string that actually got
replaced was a truncated fragment — the edit applied, but it was not the edit I intended, and "0 pass,
7 fail" is the signature of a broken harness rather than a pinned guard.

That is the same lesson as before with a new mechanism, and it is worth stating plainly: I verified
that each mutation *applied*, which is what I said I had learned to do, but I did not verify that what
applied was what I *meant*. Every mutation result in this comment was produced by a Python harness with
no shell field splitting, and I checked one suspicious result — a 0-pass run — and found it was another
harness failure (unbalanced braces), not a caught mutation. I reformulated it and report the real number.

## Change 1: classify, do not delete

The discard removed the pair and rewrote it, turning one atomic `writeFileDurable` rename into a
four-step sequence with the durable pair absent in the middle. That is the worst available intermediate
state, because `materializeWorkspacePatchDependencies` **silently skips** a dependency missing both
halves. One reviewer injected a failure into the second write and demonstrated the artifact dir ending
as `['workspace.patch']`, then ran two further generations against that state and got this issue's exact
error, permanently — the classifier requires both halves, so it declines to recognise the wedge it
created.

`holdsSupersededWorkspacePatchPair` now only decides; `writeWorkspacePatchArtifact` takes
`replaceSuperseded`. Each file is replaced by rename, and a failed write leaves the old pair intact.

## Change 2: check the baseline, not just self-consistency

As written, any well-formed capture was accepted — **including a dependency's own pair copied into this
node's artifact dir**, which is reachable because agents are handed every dependency artifact dir. That
shape failed loudly before this PR and would have become silent. The classifier now requires `base_tree`
to equal this node's baseline, and delegates the rest to the already-imported
`validateWorkspacePatchCapture` (schema, object ids, digest, patch size, symlink/submodule entries, and
`base_commit` against `git rev-parse HEAD`). Strictly stronger, shorter, and it removed two dead
`isStrictlyInsideDirectory` guards, an unkillable digest-format regex, and a redundant `typeof` check
that reviewers identified separately.

## Honesty fixes in the code and tests

- The docstring called this **provenance**. It is not: `patch_sha256` digests a file the agent can read,
  in a directory reachable from the model-controlled workspace. It now says this is a self-consistency
  heuristic and a tripwire, acceptable only because the runtime's capture overwrites the survivor either
  way, and names `invariantSuiteProtectedBaselinePath` as what real provenance would look like.
- The `maxAttempts` reasoning now gives the structural reason rather than a fact about today's topology.
- The test header claimed "everything that touches the filesystem is the real thing" while stubbing
  `writeFileDurable` and `assertRegularFileInside` — the very function the atomicity argument turns on.
  Rather than soften the comment, the suite now uses the **real** `writeFileDurable`,
  `assertRegularFileInside` and `validateWorkspacePatchCapture` over a **real git repository**.
- The two path-resolution failures now name `workspace.patch` / `workspace-patch.json` instead of an
  unnamed "artifact", matching #329.
- The split-brain source-text verifier needed updating; it was extended rather than unbroken, and now
  pins that the rejection stays gated on the flag, that the flag defaults to rejecting, and exercises the
  new arm.

## Mutation results, corrected harness

| mutation | suite | verdict |
|---|---|---|
| baseline | 12 pass, 0 fail | — |
| drop the classify call | 8 pass, **4 fail** | caught |
| classify only the canonical root, not the mirror | 11 pass, **1 fail** | caught |
| drop the `base_tree` identity check | 11 pass, **1 fail** | caught |
| drop the capture validation | 9 pass, **3 fail** | caught |
| swallow path-safety failures in the classifier | 11 pass, **1 fail** | caught |
| drop the both-halves requirement | 10 pass, **2 fail** | caught |

The first four correspond to mutations a reviewer found **surviving** the previous suite (7 pass / 0
fail each); two of those four are now impossible rather than merely tested, since the dead guard is gone
and the digest comparison moved into the production validator.

## Gates

- Runtime suite: 431 tests, 416 pass, 0 fail, 15 skipped
- bun lane: 14 pass, 0 fail
- typecheck clean; `pnpm lint` clean; repo-wide `prettier --check .` clean

Reviewers also resolved the open question I left in #357: inheriting generation 0's
`workspace-patch-baseline.json` is deliberate and required, and re-deriving it would be actively wrong.
Recorded on the issue and stated in the code.

aviggiano:
I checked this against the R52 volume as it stands right now, because the workspace is being resumed and will re-enter that node.

**The resume is live and runs a pre-fix image.** Container `ta-01KZEWX79TV9X3S9E166S48BES`, `/run/ultrafuzz-config/lineage.json`:

```
"logical_run_id": "aave-v4-v0012-main-issue328fix-r52-invariant-only",
"generation": 1, "attempt": 4, "workspace_mode": "resume"
```

`benchmark.json` names `ultrafuzz-security-runner:v0.0.12-r52-main-3190f90-invariant-only` — built at `3190f90` (#329), 17 commits behind `origin/main` and predating this branch. So generation 2 will re-trip #357 unless the image is rebuilt or the pair is removed from the volume by hand.

**This PR's discriminator does fire on the surviving pair.** The two files still on disk, untouched since 18:52:

```
artifacts/stateful-invariant-campaign/workspace-patch-baseline.json  (18:37)
  {"schema_version":"ultrafuzz.workspace-patch-baseline.v1",
   "attempt_id":"stateful-invariant-campaign",
   "baseline_tree":"b92f9cf321dd71d62ac64cafb0611f1eb50613ae"}

artifacts/stateful-invariant-campaign/workspace-patch.json  (18:52)
  {"base_tree":"b92f9cf321dd71d62ac64cafb0611f1eb50613ae",
   "base_commit":"6959e3219b5506bf2acae18551cbb2a68a5b8fba",
   "result_tree":"00d4d27499671615f4d0fd2871709a4dfdf880a4",
   "patch_sha256":"a631d29d849efabba201a8260228969dab41e72cb78ce42438ba90afff83c996",
   "files":[{"path":"tests/foundry/stateful-invariant-campaign/ReconFailureRepros.t.sol"}]}
```

`baseline_tree` persists across generations, so the re-executed node's `baselineTree` is `b92f9cf3…`, which is exactly the survivor's `base_tree` — `holdsSupersededWorkspacePatchPair` returns true and the pair is replaced instead of throwing. Verified on the volume, not inferred.

**I duplicated this.** I diagnosed the same failure independently and filed #359 / PR #360 before finding #357. Yours is first and strictly better — mine clears the pair at baseline time, which throws away the cross-generation tamper signal that your `base_tree` check preserves. I've closed #360 as superseded and marked #359 a duplicate. Sorry for the noise.

Current node counts for the record: succeeded=25, failed=2 (`stateful-invariant-campaign`, `property-specification-fanin` #285), pending=5. `dedupe-findings` → `triage` → `severity-classification` → `aggregate-test-files` → `final-report` are all downstream of the campaign node, so this PR is the whole remaining tail of the run.

PR #360, opened 2026-08-07
fix(runtime): clear the previous attempt's workspace patch before retrying a node

### LINKED ISSUE #359: A node that publishes a workspace patch can never succeed on retry
## Summary

A node that publishes a workspace patch and then fails **after** the patch is written can never succeed on retry. The runtime's own artifact from the previous attempt trips the anti-tampering guard that exists to reject agent-authored patches, and the guard cannot tell the two apart.

This ended the R52 Aave v4 invariant-only run with 25 of 32 nodes green, after the campaign had already completed and produced findings — twice.

## Reproduction from R52

`stateful-invariant-campaign`, two attempts, both `executor-error`:

| attempt | window | cause |
|---|---|---|
| 1 | 18:37:09 → 18:52:35 | `artifact-contract failure for findings.json (ultrafuzz/findings@1): Invalid input: expected "1.0" at findings.json[0].schema_version, findings.json[1].schema_version, findings.json[2].schema_version` |
| 2 | 19:00:54 → 19:15:48 | `artifact-contract failure: workspace patch artifact was modified workspace.patch` |

Attempt 1 ran the fuzzing campaign to completion, found three property violations, and generated a reproducer. It failed only on the findings schema version. Before it failed, the runtime had already written `workspace.patch` and `workspace-patch.json`; the surviving manifest describes exactly one file:

```json
{
  "schema_version": "ultrafuzz.workspace-patch.v1",
  "base_commit": "6959e3219b5506bf2acae18551cbb2a68a5b8fba",
  "result_tree": "00d4d27499671615f4d0fd2871709a4dfdf880a4",
  "files": [{ "path": "tests/foundry/stateful-invariant-campaign/ReconFailureRepros.t.sol" }]
}
```

Attempt 2 re-ran the campaign and **the model had self-corrected** — the `findings.json` now in the task workspace carries `schema_version: "1.0"` on all three entries. It cleared the findings contract and died on the stale patch instead.

## Cause

`packages/runtime/src/templates/smithers/workflows/workflow.tsx:804-824`:

```ts
function writeWorkspacePatchArtifact(root: string, relativePath: string, contents: string): void {
  …
  if (existsSync(target)) {
    const existingContents = readFileSync(existing, "utf8");
    if (existingContents !== "" && existingContents !== "\n") {
      if (existingContents === contents) return;
      throw new Error(`artifact-contract failure: workspace patch artifact was modified ${relativePath}`);
    }
  }
  // These paths are runtime-owned. Replace only an empty runtime placeholder;
  // reject any non-empty agent-authored or tampered patch above.
  writeFileDurable(target, contents);
}
```

The comment states the assumption exactly: at write time the artifact is expected to be absent or an empty placeholder. A retry violates that, because the previous attempt's runtime write is still there. The captured patch necessarily differs between attempts — a fuzzing campaign produces different corpus and reproducer content each run — so the equality short-circuit never saves it.

The model is not implicated. `workspace.patch` appears in the node's declared outputs in `topology.yml`, but the rendered prompt for `stateful-invariant-campaign` never mentions it; I checked the 17 KB `prompt.rendered.md` from the run.

## Why it matters

- **It converts recoverable failures into terminal ones.** Attempt 1's failure was a model mistake the model then fixed by itself. Without this guard firing, R52 would have completed the campaign on attempt 2.
- **It is deterministic, not flaky.** Every subsequent attempt hits it. Retry and recovery budget are spent re-running a 15-minute campaign into a guaranteed failure.
- **It affects every workspace-patch-publishing node**, not just the campaign — `topology.yml` declares `workspace.patch` on several.

## Suggested fix

Clear the runtime-owned `workspace.patch` and `workspace-patch.json` from the artifact roots at the start of each attempt, where the workspace patch baseline is captured for that attempt. The guard then keeps its original meaning — reject anything non-empty written *during* the attempt — without treating the runtime's own prior output as tampering.

The alternative, stamping the attempt id into the manifest and letting the guard allow overwrite when the existing artifact belongs to a prior attempt, is more faithful but larger: the manifest has no attempt field today and it would need one on both files, including the plain-text `workspace.patch` that has no place to put it.

## Related

- The findings `schema_version` divergence in attempt 1 is a separate producer/contract issue in the same family as #328. It self-corrected here, so it is a quality problem rather than a blocker, but the campaign prompt documents `schema_version` for `recon-fuzzer-results.json` and says nothing about it for `findings.json`.
- The campaign's three findings are worth their own look; one (`property_catalogAddedSharePriceNonDecreasing`, violated by borrow → accrue → supply) is sequence-dependent, and two look like missing rounding tolerance rather than protocol defects.

aviggiano:
## Summary

Fixes #359. A node that publishes a workspace patch and fails **after** the patch is written can never succeed on retry, because the runtime's own artifact from the previous attempt trips the guard meant to catch agent tampering.

## The guard and the assumption it makes

`writeWorkspacePatchArtifact` (`workflow.tsx`) refuses to overwrite a non-empty `workspace.patch` whose contents differ, and its comment states the assumption plainly:

> These paths are runtime-owned. Replace only an empty runtime placeholder; reject any non-empty agent-authored or tampered patch above.

That holds within an attempt. Across attempts it does not: `materializeWorkspacePatch` writes the artifacts at the *end* of an attempt, the artifact directory outlives the attempt, and a later attempt captures different contents — a fuzzing campaign alone produces new corpus and reproducer content every run — so the equality short-circuit never saves it.

## What it cost

R52 ended with 25 of 32 nodes green:

| attempt | window | cause |
|---|---|---|
| 1 | 18:37 → 18:52 | `findings.json` schema_version mistake — but the campaign had **already run**, found three violations, and written `tests/foundry/stateful-invariant-campaign/ReconFailureRepros.t.sol` |
| 2 | 19:00 → 19:15 | `artifact-contract failure: workspace patch artifact was modified workspace.patch` |

The model self-corrected between attempts — the `findings.json` in the task workspace now carries `schema_version: "1.0"` on all three entries — so attempt 2 cleared the findings contract and died on attempt 1's leftover patch instead. A recoverable, self-healing failure became a terminal run failure.

It is deterministic: every further attempt hits it, spending a 15-minute campaign each time.

## The change

Clear the two runtime-owned artifacts once the attempt's baseline is settled, inside the block that establishes `workspacePatchBaselineTrees` for the attempt. The guard then means what its comment claims — nothing non-empty was written *during this attempt*.

Placement notes:

- The clear runs **after** `captureWorkspaceTree`, so it cannot perturb the baseline the patch is diffed against. One of the artifact roots is a mirror inside the workspace, so ordering matters here.
- It is keyed on the in-memory `workspacePatchBaselineTrees` miss rather than on `persistedBaseline === undefined`. The baseline file lives in the artifact directory and therefore survives into the next attempt, so gating on the persisted value would have skipped exactly the R52 case.
- Anything that is not a regular file is skipped rather than removed, leaving the write path to reject it with its own error.

## Alternative considered

Stamping the attempt id into the manifest and letting the guard permit an overwrite when the existing artifact belongs to a prior attempt is more faithful to intent, but larger: the manifest has no attempt field today, and `workspace.patch` is plain text with nowhere to put one.

## Tests

`packages/runtime` full suite — `node scripts/run-tests.mjs` plus the bun-run generated-adapter tests — passes: **14 pass, 0 fail** on the bun leg, with the node leg green ahead of it. `prettier --check` clean.

Note this required installing `bun`, which the suite needs and which was absent from my environment; that is a property of my box, not of the change.

## Not addressed here

The `findings.json` schema_version divergence from attempt 1. It self-corrected, so it is a quality issue rather than a blocker, but the campaign prompt documents `schema_version` for `recon-fuzzer-results.json` and says nothing about it for `findings.json` — same family as #328.

<h3>Greptile Summary</h3>

The PR clears stale runtime-owned workspace-patch artifacts after establishing an attempt’s baseline, preventing previous-attempt output from triggering the artifact-tampering guard on retry.
- Removes `workspace.patch` and `workspace-patch.json` from canonical and mirrored artifact roots before agent execution.
- Preserves baseline correctness by performing cleanup only after baseline capture.
- Leaves non-regular filesystem entries intact so the existing write-path validation reports them.

The PR appears safe to merge, with no actionable correctness or security failures identified.

The cleanup runs after baseline capture, targets only the two runtime-owned patch files, and does not affect dependency handoffs or workspace diffs because canonical artifacts are republished and mirrored artifacts live under an excluded runtime root.

| Filename | Overview |
|----------|----------|
| packages/runtime/src/templates/smithers/workflows/workflow.tsx | Clears stale workspace-patch outputs after baseline establishment while retaining existing containment and file-type validation behavior. |

</details>

```mermaid
sequenceDiagram
    participant Runtime
    participant Baseline as Baseline cache
    participant Artifacts as Artifact roots
    participant Agent
    Runtime->>Baseline: Load or capture workspace baseline
    Runtime->>Artifacts: Remove stale patch and manifest
    Runtime->>Agent: Execute attempt
    Agent-->>Runtime: Return workspace changes
    Runtime->>Artifacts: Write current patch and manifest
```
</details>

<sub>Reviews (1): Last reviewed commit: ["fix(runtime): clear the previous attempt..."](https://github.com/monad-developers/ultrafuzz/commit/144c4903a3b08b04680ec23b4149791a24676279) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=51292347)</sub>


aviggiano:
The R52 workspace is being resumed right now, and this resume will re-trip the bug.

Live container `ta-01KZEWX79TV9X3S9E166S48BES`, `/run/ultrafuzz-config/lineage.json`:

```
"logical_run_id": "aave-v4-v0012-main-issue328fix-r52-invariant-only",
"generation": 1, "attempt": 4, "workspace_mode": "resume"
```

`benchmark.json` names the image `ultrafuzz-security-runner:v0.0.12-r52-main-3190f90-invariant-only`, i.e. built at `3190f90` (#329). That is 17 commits behind `origin/main` and, more to the point, predates this branch — so the resume runs the pre-fix `materializeWorkspacePatchDependencies`.

The stale artifact this PR clears is still on the volume, untouched since R52 died:

```
artifacts/stateful-invariant-campaign/
  workspace-patch-baseline.json     Aug  7 18:37   # attempt 1's baseline
  workspace-patch.json              Aug  7 18:52   # attempt 2's output
  workspace.patch            1141B  Aug  7 18:52
```

`state.json` still has `stateful-invariant-campaign  failed  retry_count=2  last_error="artifact-contract failure: workspace patch artifact was modified workspace.patch"`, so the node re-runs on resume. Because `workspace-patch-baseline.json` persists, the attempt reuses attempt 1's baseline and skips straight to the write path with the 18:52 `workspace.patch` still present; a re-run fuzzing campaign cannot reproduce those bytes, so `writeWorkspacePatchArtifact` will read them as tampering again.

Rest of the run for context — 25 succeeded, 2 failed, 5 pending; the only other failure is `property-specification-fanin` (#285). `dedupe-findings`, `triage`, `severity-classification`, `aggregate-test-files` and `final-report` are all downstream of the campaign node, so this is the whole remaining tail of the run.

Two ways out, neither of which I can take (read-only Modal access, and merges aren't mine): merge this and rebuild the image, or delete `artifacts/stateful-invariant-campaign/workspace.patch` and `workspace-patch.json` on the volume before the resume reaches that node.

aviggiano:
Superseded by #358, which fixes the same bug from #357 — filed first, and with the better mechanism.

Both PRs stop `writeWorkspacePatchArtifact` reading a re-executed node's own prior-generation patch as tampering. Mine deletes the pair once the attempt's baseline is settled; #358 classifies the survivor via `base_tree` identity against the persisted baseline plus `validateWorkspacePatchCapture`, then overwrites it. The difference matters: deleting the pair discards the tamper signal across the generation boundary — after my clear, an agent-authored patch written early in the new attempt is indistinguishable from an absent one. #358 keeps that check intact and rejects a lone patch, a mismatched digest, or a foreign manifest exactly as before.

I verified #358's discriminator fires on the actual R52 artifacts still on the volume (details in https://github.com/monad-developers/ultrafuzz/pull/358#issuecomment-5221529904): the persisted `baseline_tree` and the survivor's `base_tree` are both `b92f9cf321dd71d62ac64cafb0611f1eb50613ae`.

Closing. #359 is a duplicate of #357.

PR #372, opened 2026-08-07
Exclude oversized untracked roots by size, and record what was left out

### LINKED ISSUE #368: Generated-corpus exclusion is an exact-name list, so an agent-chosen variant like recon-corpus-deep blows the capture buffer again
## What happened

R53 (`aave-v4-v0012-main-issue357fix-r53-invariant-only`, image from `2d47e73`) failed
`stateful-invariant-handlers` after ~15 minutes with:

```
git diff produced more than the 33554432-byte capture buffer (a handed-off patch must also stay under
16777216 bytes); the workspace is too large to hand off. Largest contributors within the first 33902592
bytes git wrote; git emits in path order, so anything past that cutoff is not visible here:
```

Node ran 22:14:47Z -> 22:29:41Z, `retry_count 0`. The run went to recovery generation 1 at 22:30:58Z.

## Why this is a NEW defect and not a regression of #305

#305 fixed exactly this class by excluding generated corpus directories from the workspace capture:

```ts
const WORKSPACE_GENERATED_ROOTS = ["recon-corpus", "echidna", "magic"] as const;
```

That is an **exact-name** list. The agent did not use those names — it created **`recon-corpus-deep`**
and **`echidna-deep`**, which are not in the list, so they were captured. One file in
`recon-corpus-deep` alone contributed ≥33.8 MB, over the 32 MiB buffer by itself.

Grepping `.ultrafuzz/prompts/strategies/invariants/` for `-deep` returns **nothing**, so these names are
not prompt-specified. The agent chose them, presumably for a deeper fuzzing pass alongside the
prompt-specified `--corpus-dir echidna --recon-corpus-dir recon-corpus`. Any other variant
(`recon-corpus-2`, `corpus-deep`, …) would escape the same way.

So the exclusion mechanism assumes it can enumerate every generated directory by name, and nothing
constrains the agent to those names.

## The #311 fix is what made this diagnosable

Worth noting because it validates that change: the message names the actual contributors with byte
counts and states its own visibility limit ("anything past that cutoff is not visible here"). Before
#311 this was an anonymous ENOBUFS. The diagnosis took one read.

## Candidate fixes, not presuming which

1. **Prefix/glob matching** on the existing roots (`recon-corpus*`, `echidna*`). Smallest change, but it
   is still a name-based guess and `corpus-deep` would escape it.
2. **Exclude by what the directory IS rather than what it is called** — e.g. any root the target
   gitignores, or any untracked root above a size threshold. Robust against naming, but changes what a
   legitimate large artifact means and needs care so real work is never silently dropped.
3. **Constrain the agent** to write corpora only under a single reserved root, and treat anything else
   as a contract violation. Makes the invariant explicit rather than inferred, but it is a prompt change
   and only as reliable as the agent's compliance.

Option 2 is the only one that does not depend on predicting a name, but it is also the one that could
silently exclude something a node legitimately needs to hand off, so it needs the most care. I am not
choosing here.

## Acceptance criteria

- [ ] A generated corpus directory with an agent-chosen name does not enter the workspace capture.
- [ ] A regression test covers a variant name, not just the three known ones.
- [ ] The failure mode, if it recurs, still names the contributing paths (do not regress #311).

aviggiano:
Closes #368.

## What R53 hit

`stateful-invariant-handlers` failed on an image that **already carried the #305 corpus exclusion**:

```
git diff produced more than the 33554432-byte capture buffer ... Largest contributors ...:
```

#305 excludes generated corpora by **exact name** — `recon-corpus`, `echidna`, `magic`. The agent wrote
its deep fuzzing pass to `recon-corpus-deep/` and `echidna-deep/`. Grepping the invariant prompts for
`-deep` returns nothing, so those names are the agent's own invention; no name list could have predicted
them. One file under `recon-corpus-deep` was ≥33.8 MB by itself, over the entire 32 MiB buffer before
anything else counted.

The #305 comment called this exactly: *"a gap a size ceiling would close and a name list cannot."*

## The rule, and the hazard it creates

**8 MiB per untracked top-level directory.**

A size rule has one serious failure mode, and it is the same one the surrounding comments warn about for
tracked paths: it turns a **loud** failure into **silent data loss**. `applyWorkspacePatch` verifies both
trees under the same exclusions, so an omission passes every check and the dependent simply sees content
that was never delivered.

Two things prevent that:

1. **The manifest records every exclusion** — `excluded_roots: {path, bytes}[]`, optional and additive.
   `validateManifest` validates it when present, and it is absent when nothing was excluded, so its
   presence always means something really was left out.
2. **The rule is narrow** — untracked only, directories only, above a ceiling no authored source in this
   pipeline approaches.

Capture/apply symmetry holds despite the rule reading the filesystem: the dependent worktree never
received an excluded root, and excluding an absent root is a no-op. The two sides agree on the staged set
without needing to agree on the decision.

## Two #311 regressions were re-aimed, not deleted

A single oversized untracked root no longer overflows, so both tests had to change. Weakening them would
have discarded the diagnostic that made this issue diagnosable in one read:

- **attribution test** now builds **aggregate** overflow from six 6 MiB roots, each *under* the ceiling —
  a case the ceiling deliberately does not catch, so contributor ranking is still exercised.
- **byte-ranking test** now modifies **tracked** files, which the untracked-only ceiling does not touch.

Both still assert exactly what they were written to assert.

## Mutation results

| mutation | result |
|---|---|
| baseline | 45 pass, 0 fail |
| drop the ceiling exclusion | **caught** |
| stop recording exclusions in the manifest | **caught** |
| raise the ceiling far above any real corpus | **caught** |
| widen the rule to top-level FILES | *survived — equivalent mutant* |

The survivor is genuinely equivalent: `oversizedUntrackedRoots` skips entries with no separator, so a
top-level file's name never enters the set and changing only the consumer has no effect. Mutating **that**
guard instead **is** caught — by a test I added after mutation testing revealed the "directories only"
narrowing was asserted in a comment and enforced by nothing.

## Gates

- Runtime suite: 438 tests, 421 pass, 15 skipped
- **The 2 failures are pre-existing on `origin/main`**, verified by stashing this change and re-running:
  identical failures either way (417/2 of 434 without, 421/2 of 438 with)
- prettier clean on both files

<h3>Greptile Summary</h3>

The PR excludes untracked top-level directories larger than 8 MiB from workspace handoffs and records exclusions in the patch manifest.
- Adds per-root size measurement and exclusion during staging.
- Extends runtime manifest validation with optional exclusion records.
- Reworks overflow diagnostics tests and adds coverage for oversized directories, ordinary directories, and large top-level files.

<h3>Confidence Score: 2/5</h3>

The PR does not appear safe to merge because generated exclusion manifests can be rejected and the ceiling remains bypassable through raw-path decoding and concurrent growth.

The runtime emits a field absent from the strict shared artifact contract, while size calculation can still miss oversized content when filenames contain invalid UTF-8 or corpus files grow between measurement and staging.

**Files Needing Attention:** packages/runtime/src/workspace-handoff.ts; packages/artifacts/src/workspace-patch.ts; packages/artifacts/schema/workspace-patch.schema.json

| Filename | Overview |
|----------|----------|
| packages/runtime/src/workspace-handoff.ts | Adds oversized untracked-root filtering, exclusion metadata, and runtime manifest validation. |
| packages/runtime/test/workspace-handoff.test.ts | Re-aims overflow diagnostics and adds focused coverage for the intended size-ceiling boundaries. |

</details>

```mermaid
flowchart TD
  A[Enumerate tracked and untracked paths] --> B[Measure untracked directory roots]
  B --> C{Root exceeds 8 MiB?}
  C -- Yes --> D[Exclude root and record path plus bytes]
  C -- No --> E[Stage raw pathspec]
  D --> F[Write result tree and patch]
  E --> F
  F --> G[Return manifest with optional excluded_roots]
```
</details>

<sub>Reviews (2): Last reviewed commit: ["test(runtime): drop the baseline orphane..."](https://github.com/monad-developers/ultrafuzz/commit/5672748a0e8115e78d03afb443e162db8481605c) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=51345144)</sub>

**Context used:**


aviggiano:
Reviewed this against the live R53 volume while it was still the blocking failure. The rule and the
manifest record both look right to me, and re-aiming the #311 tests rather than deleting them is the
call I'd have made. One blocking defect, reproduced end to end — it's Greptile's Issue 1, but the
consequence is worse than "schema-validating consumers reject it", so here is the measurement.

## The manifest this writes is rejected by the artifact gate, so the run still dies at the same node

`workspacePatchSchema` is a `z.strictObject`, and `validateArtifactContract` routes
`ultrafuzz/workspace-patch@1` straight into it (`packages/artifacts/src/artifact-contracts.ts:311`).
`artifact-gates.ts:1462` calls that on every declared node output, and `stateful-invariant-handlers`
declares `workspace-patch.json` with exactly that contract. So the moment the ceiling fires, the
manifest fails its own output contract.

Run on `19132a9` with the R53 shape — one agent-named untracked root over the ceiling, one authored
file alongside it:

```
capture succeeded; manifest keys: schema_version, base_commit, base_tree, result_tree, patch_sha256, files, excluded_roots
excluded_roots: [{"path":"recon-corpus-deep","bytes":35651584}]
artifact contract ok: false
issues: [
  {
    "path": "artifacts/x/workspace-patch.json",
    "code": "WORKSPACE_PATCH_SCHEMA_INVALID",
    "message": "Unrecognized key: \"excluded_roots\""
  }
]
```

The capture now succeeds, which is the fix working. But the node then fails its artifact contract, so
R53 would still have died at `stateful-invariant-handlers` — ENOBUFS traded for
`WORKSPACE_PATCH_SCHEMA_INVALID`, and only on the exact runs where the ceiling did its job. The runtime's
own `validateManifest` does ignore unknown fields, which is why the runtime suite stays green and this
does not show up until a real gate runs.

Worth saying plainly: the exclusion record is what makes the ceiling safe rather than silent, so the
field has to stay. It's the schema that has to move.

## The patch

Three files. `workspacePatchPath` already accepts a single segment and already rejects `.git`,
`.ultrafuzz`, `node_modules`, `artifacts`, `.env*`, so reusing it is stricter than the hand-rolled
`!path.includes("/")` check in `validateManifest` rather than weaker.

`packages/artifacts/src/workspace-patch.ts`:

```ts
export const workspacePatchExcludedRootSchema = z.strictObject({
  path: workspacePatchPath,
  bytes: z.number().int().nonnegative()
});

export const workspacePatchSchema = z
  .strictObject({
    // ...
    files: z.array(workspacePatchFileSchema),
    excluded_roots: z.array(workspacePatchExcludedRootSchema).optional()
  })
```

`workspacePatchJsonSchema` in the same file needs the matching optional property (`excluded_roots`
array of `{path, bytes}`, `additionalProperties: false`, both required), and
`packages/artifacts/schema/workspace-patch.schema.json` needs the identical block or
`artifact schema snapshots are present and aligned with exported schema constants` in
`packages/artifacts/test/schema.test.ts` fails on the mismatch — that snapshot test is what catches this
class, and it's the reason to make the change in the schema package rather than loosening the strict
object.

`contract_digest` is unaffected: it is a hash of the `artifact-contracts.ts` definition
(`id`/`format`/`description`), not of the zod schema, so an optional field does not re-pin anything.

I confirmed the whole set green locally on an independent implementation of the same rule: artifacts 73
pass / 0 fail, runtime 437 tests / 422 pass / 0 fail, prettier clean.

## Second, non-blocking: the size probe stats a decoded path

`oversizedUntrackedRoots` keys on `entry.subarray(0, separator).toString("latin1")` — deliberately
byte-exact, and the comment says why — but then stats `path.resolve(workspaceRoot, entry.toString("utf8"))`.
For a non-UTF-8 filename that resolves to a replacement-character path, `lstatSync` throws ENOENT, the
`catch` counts it as zero, and the root escapes the ceiling. `captures a filename that is not valid UTF-8`
already establishes such names reach this code.

`lstatSync` takes a Buffer path, so this stays byte-exact:

```ts
const stat = lstatSync(Buffer.concat([Buffer.from(`${path.resolve(workspaceRoot)}${path.sep}`), entry]));
```

Greptile's Issue 3 (a file growing between the size pass and `git add`) I'd leave alone — the sizes can
only ever be a sample, `MAX_PATCH_BYTES` still backstops it, and the contributor ranking still fires,
which is the outcome the ceiling is explicitly documented as not neutralising.

I have the schema half written and tested if it's useful, but it belongs on this branch rather than in a
competing PR — say the word and I'll hand over the diff rather than push anything here.

aviggiano:
The blocking defect I measured above is now fixed in **#374**, which is artifacts-only — **zero file
overlap with this PR**, so nothing here needs to change and the two merge in either order. A run is only
unblocked once both are in.

Verified end to end by cherry-picking this branch (`5672748`) onto #374 and running the R53 shape — a
34 MB agent-named corpus root plus authored work beside it:

```
manifest keys: schema_version, base_commit, base_tree, result_tree, patch_sha256, files, excluded_roots
excluded_roots: [{"path":"recon-corpus-deep","bytes":35651584}]
files: [{"path":"test/Handlers.t.sol"}]
artifact contract ok: true          <- was false: WORKSPACE_PATCH_SCHEMA_INVALID
```

Same capture against `origin/main`'s artifacts package still returns `Unrecognized key: "excluded_roots"`,
so the red/green is on the schema change alone rather than on anything about the capture.

One thing worth flagging for this PR specifically: #374 validates `excluded_roots[].path` deliberately
loosely — no charset pattern — because this branch renders roots with `toString("latin1")` to match a
non-UTF-8 directory name exactly. Applying the `files` charset there would have failed the gate on exactly
the oversized corpus the field exists to report. If a later change to `oversizedUntrackedRoots` ever
normalises those names, the schema is looser than it needs to be, which is the safe direction.

aviggiano:
The design review is right and I am closing this. I verified both blocking findings myself rather than
taking them on trust, and both reproduce.

**1. It silently destroys the deliverable of the very node #368 is about.** The exclusion is keyed on the
FIRST path segment, and in a Foundry target the node's authored Solidity lands under `test/`. My own
probe, tracked `test/` plus authored files beside a 9 MiB corpus under `test/recon/corpus-deep/`:

```
files handed off : []
excluded_roots   : [{"path":"test","bytes":9437224}]
```

Both `Handlers.t.sol` and `TargetFunctions.sol` gone. This is `stateful-invariant-handlers`, whose entire
output is Solidity under `test/`. I wrote a fix whose stated purpose was to avoid silent data loss and it
causes exactly that, on the node in the issue.

**2. It does not even unblock the run.** `workspacePatchSchema` in
`packages/artifacts/src/workspace-patch.ts:66` is a `z.strictObject` with no `excluded_roots`, and
`artifact-contracts.ts:311` routes `ultrafuzz/workspace-patch@1` into it. So the moment the ceiling fires,
the node fails its own output contract — ENOBUFS traded for `WORKSPACE_PATCH_SCHEMA_INVALID`.

I missed this because I checked `validateManifest` inside `workspace-handoff.ts`, saw it tolerates unknown
keys, and concluded the field was additive. I verified one validator and generalised to "the manifest
accepts it". That is the same error I have been correcting elsewhere in this investigation — concluding
absence from a probe that could not have found the thing — applied to my own change.

**3. My "equivalent mutant" claim was wrong in both directions**, and the reviewer's reading is correct:
dropping `separator >= 0 &&` yields `entry.subarray(0, -1)`, the name minus its last byte, so a top-level
file named `<oversized-root>`+1 byte WOULD be dropped — a non-equivalent surviving mutant. And the test I
added kills neither single mutation, only both together. I overstated what the table pinned.

**4. The metric is wrong, and this repo already rejected it once.** PR #306 was closed for measuring the
workspace rather than the diff. I scoped to untracked, which fixes #306's largest defect, but kept
on-disk bytes as the proxy. Measured by the reviewer: 7 MiB of NULs deflates to ~41 KB of patch, so the
rule over-excludes compressible corpora by ~200x, while five 7 MiB incompressible roots still overflow.
My own regression test uses 9 MiB of ASCII `"0"`, which compresses — so it proves an exclusion happened,
never that an overflow was prevented. That is precisely the criticism I would have made of someone else's
test.

**The design I should have built**, which the review sets out and I agree with: capture, and only on a
MEASURED overflow, retry with the largest contributors excluded — ranked by real diff bytes, which
`git-capture-diagnostics.ts` already computes for #311. It keys on the metric that actually breaks,
fires only when something would genuinely fail, covers the aggregate case this ceiling admits it cannot,
and needs no magic constant. Recorded on #368 for reimplementation.

Closing rather than patching: the granularity defect is not a tweak to this approach, it is a consequence
of choosing path-prefix as the exclusion unit.

PR #366, opened 2026-08-07
v0.1.0 release (open source candidate)

aviggiano:
- Do not merge until we have evals showing this is much better!

PR #374, opened 2026-08-07
fix(artifacts): let a workspace-patch manifest record what the capture left out

aviggiano:
Refs #368. **The other half of #372** — that PR cannot unblock a run without this one.

## Why this exists

R53 (`aave-v4-v0012-main-issue357fix-r53-invariant-only`) died terminally at
`stateful-invariant-handlers` at 22:59:59Z, `run: failed`, 23 succeeded / 8 pending / 1 failed:

```
git diff produced more than the 33554432-byte capture buffer ... Largest contributors ...:
```

#372 fixes the capture: it drops an oversized untracked root by size and records it in the manifest as
`excluded_roots`, because a silent omission is worse than the overflow — `applyWorkspacePatch` verifies
both trees under the same exclusions, so every check passes and the dependent just sees content that was
never delivered.

**That record has nowhere legal to live today.** `workspacePatchSchema` is a `z.strictObject`, and
`validateArtifactContract` routes `ultrafuzz/workspace-patch@1` straight into it
(`packages/artifacts/src/artifact-contracts.ts:311`). `artifact-gates.ts:1462` calls that on every declared
node output, and `stateful-invariant-handlers` declares `workspace-patch.json` with exactly that contract.
So #372 alone trades an ENOBUFS at that node for a contract failure at the same node.

This is measured, not predicted — same capture, the two artifact packages:

| gate | result |
|---|---|
| `origin/main` artifacts | `artifact contract ok: false` — `WORKSPACE_PATCH_SCHEMA_INVALID: Unrecognized key: "excluded_roots"` |
| this branch | `artifact contract ok: true` |

Run end to end on `#372 (5672748)` cherry-picked onto this commit, with the R53 shape — a 34 MB
agent-named corpus root plus authored work beside it:

```
manifest keys: schema_version, base_commit, base_tree, result_tree, patch_sha256, files, excluded_roots
excluded_roots: [{"path":"recon-corpus-deep","bytes":35651584}]
files: [{"path":"test/Handlers.t.sol"}]
artifact contract ok: true
```

The capture succeeds, the corpus is excluded and recorded, and the authored test still hands off.

## The trap this closes

A manifest field has to be added in **three** places or it fails only on Modal, never in the suite. The
runtime's own `validateManifest` ignores unknown keys, so a capture-side change stays green through the
entire runtime suite (#372: 438 tests) and dies at the gate on a live run. All three are covered here:

1. `workspacePatchSchema` (the `z.strictObject`)
2. `workspacePatchJsonSchema`
3. the checked-in snapshot `packages/artifacts/schema/workspace-patch.schema.json`

Optional, and absent rather than empty when nothing was excluded, so **presence always means something
really was left out**. Every pre-#368 manifest stays valid and the schema version does not move.
`contract_digest` hashes the `artifact-contracts.ts` definition (id/format/description), not the zod
schema, so an optional field does not re-pin it.

## One deliberate looseness, pinned by a test

`excluded_roots[].path` is validated far more loosely than `files[].path`, and the difference is not an
oversight. A `files` path is **applied** — `applyWorkspacePatch` writes it into the dependent worktree —
so every segment must survive `normalizeWorkspacePatchPath`. An excluded root names something that was
**never delivered** and is never opened.

Concretely: #372 keys these roots on raw bytes and renders them with `toString("latin1")`, precisely so a
directory whose name is not valid UTF-8 is matched exactly rather than through a lossy decode that could
collide two distinct roots. Run `WORKSPACE_PATCH_SEGMENT` over that and a corpus directory with one
non-ASCII byte fails the gate — reintroducing this exact defect one layer down.

What *is* enforced: one segment, non-empty, not `.` or `..`, non-negative safe-integer bytes, no duplicate
roots. All unreachable from today's producer (git never lists a `.` component; the root is by construction
the text before the first `/`), so they guard a future or hand-written producer without risking a real
capture. A test asserts the JSON Schema carries **no** charset `pattern`, so tightening it is a decision
rather than an accident.

## Gates

- `packages/artifacts`: **76 pass, 0 fail** (5 new assertions across 3 new tests)
- `prettier --check` clean on all three files

Mutation testing, since a schema change that is not load-bearing is indistinguishable from no change:

| mutation | result |
|---|---|
| drop `excluded_roots` from the zod object | **caught** (2 tests) |
| leave the checked-in snapshot at `origin/main` | **caught** (1 test) |
| add the `files` charset pattern to the JSON Schema | **caught** (2 tests) |

## Coordination

- Touches only `packages/artifacts/{src,test}/workspace-patch.ts` and its snapshot. **Zero file overlap
  with #372**, which is runtime-only — the two merge in either order, though a run is only unblocked once
  both are in.
- #366 and #165 both touch `packages/artifacts/`, but neither touches any of these three files.
- Does **not** touch `workflow.tsx`, `workflow-sync.ts`, `smithers.ts`, `node-provider.ts` or
  `public-worker.ts`, so merging this invalidates no long-lived branch rebase.

<h3>Greptile Summary</h3>

This PR extends the workspace-patch artifact contract to record size-excluded roots while retaining compatibility with older manifests.

- Adds optional `excluded_roots` metadata containing a root path and measured byte count.
- Updates the exported and checked-in JSON Schemas.
- Adds contract, path-shape, numeric-bound, duplicate-root, and schema-parity tests.

The PR appears safe to merge, with one non-blocking contract-semantics inconsistency around accepting an empty `excluded_roots` array.

The runtime gate correctly accepts and validates populated exclusion records, but the schema and test allow field presence without any excluded content, contrary to the newly documented invariant.

**Files Needing Attention:** packages/artifacts/src/workspace-patch.ts, packages/artifacts/schema/workspace-patch.schema.json, packages/artifacts/test/workspace-patch.test.ts

| Filename | Overview |
|----------|----------|
| packages/artifacts/src/workspace-patch.ts | Adds the Zod and JSON Schema definitions for excluded roots, but permits an empty present array despite documenting presence as meaningful. |
| packages/artifacts/schema/workspace-patch.schema.json | Updates the checked-in schema snapshot with the optional excluded-root records and matching bounds. |
| packages/artifacts/test/workspace-patch.test.ts | Adds broad validation coverage, including an assertion that currently conflicts with the documented absent-versus-empty invariant. |

</details>

<a href="https://app.greptile.com/ide/claude-code?prompt=%23%23%23%20Issue%201%0Apackages%2Fartifacts%2Fsrc%2Fworkspace-patch.ts%3A111%0A**Empty%20exclusions%20retain%20presence**%0A%0AThe%20optional%20array%20accepts%20%60excluded_roots%3A%20%5B%5D%60%20even%20though%20the%20adjacent%20contract%20says%20the%20field's%20presence%20always%20means%20content%20was%20omitted.%20This%20prevents%20consumers%20from%20safely%20using%20presence%20to%20distinguish%20incomplete%20captures%3B%20enforce%20a%20non-empty%20array%20here%20and%20in%20the%20corresponding%20JSON%20Schema.%0A%0A---%0A%0AFor%20each%20issue%20above%2C%20determine%20whether%20it%20is%20valid%20and%20should%20be%20fixed.%20If%20so%2C%20fix%20it%20directly.&repo=monad-developers%2Fultrafuzz&pr=374&platform=github"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInClaudeDark.svg?v=6"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInClaude.svg?v=6">Fix All in Claude Code</picture></a>

`````markdown
### Issue 1
packages/artifacts/src/workspace-patch.ts:111
**Empty exclusions retain presence**

The optional array accepts `excluded_roots: []` even though the adjacent contract says the field's presence always means content was omitted. This prevents consumers from safely using presence to distinguish incomplete captures; enforce a non-empty array here and in the corresponding JSON Schema.

---

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(artifacts): let a workspace-patch ma..."](https://github.com/monad-developers/ultrafuzz/commit/0107ee79c47dce674fa4f1ef2a26a6237af23b4c) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=51351188)</sub>

**Context used:**


aviggiano:
Converting to draft. Not mergeable as written, for a reason I should have caught before opening it.

**Its producer no longer exists.** This PR is the consumer half of #372, and #372 was closed unmerged after
review found three fatal defects in it. #368 was instead fixed by #375 (merged as `0088f80`), which matches
generated corpus roots by prefix and does not record exclusions in the manifest at all. So nothing writes
`excluded_roots`, and this change is an optional field with no producer.

**Worse, its main doc comment is false about `main`.** `workspacePatchExcludedRootSchema` justifies its
deliberately loose validation like this:

> the producer keys these roots on their raw bytes and renders them with `toString("latin1")`
> (`packages/runtime/src/workspace-handoff.ts`), precisely so that a directory whose name is not valid
> UTF-8 is matched exactly instead of through a lossy decode

Neither `latin1` nor `excluded_roots` appears anywhere in `packages/runtime/src/workspace-handoff.ts` on
`main`. That behaviour only ever existed in the closed #372. The comment's entire argument for why the
charset rules are relaxed rests on a producer that was never merged, so merging this would land a
justification a reader cannot verify and would reasonably act on.

That is the same defect class I have now flagged four times on #368 — a comment asserting a property the
code does not have. It is worth noting that it survived 8 green checks, because no test can catch it: the
claim is about a *different package* than the one under test.

**What is still worth keeping.** The distinction it draws is real and was not obvious: a `files` path is
applied and must survive `normalizeWorkspacePatchPath`, whereas an excluded root is by definition never
opened, so imposing the same charset rules would let a corpus directory with one non-UTF-8 byte in its
name convert an oversize report into a contract failure at the same node. That reasoning should be reused
when the measurement-based fix recorded on #368 is actually built — but it must ship *with* its producer,
in one PR, so the comment describes real code and a test can exercise the round trip.

Reopening this as non-draft requires that producer. Related: #377, filed tonight for a different
non-hermetic-dependency defect found on R54.

aviggiano:
Closing as dead work per #413. Its only producer, #372, was closed unmerged; `excluded_roots` is absent from `main`, and this schema-only branch requires a `bytes` value that current main never measures. There is therefore no valid producer payload for this PR to admit.

PR #375, opened 2026-08-07
Match generated corpus roots by prefix, not exact name

aviggiano:
Refs #368. Interim fix, deliberately minimal, after PR #372 was closed unmerged.

## What R53 hit

`stateful-invariant-handlers` died on an image that **already carried** the #305 exclusion:

```
```

#305 excludes by **exact name**. The agent wrote its deep fuzzing pass to `recon-corpus-deep/` and
`echidna-deep/` — names that appear nowhere in the invariant prompts, so it invented them.

## The change

One glob character on the untracked pathspec: `:(exclude)${root}*/**`.

This is the **same mechanism** as the existing exact-name exclusion, so it inherits its properties rather
than introducing new ones:

- a pathspec on the **untracked listing only**, so a tracked edit can never be dropped
- **root-anchored**, so an authored `test/` tree is untouchable even with a corpus nested inside it
- a **fixed string**, so capture and apply always agree without consulting the filesystem

Those three properties are precisely what the size-ceiling attempt in #372 broke: it dropped tracked
files, it destroyed an entire authored suite when a corpus sat under the same first path segment, and its
new manifest field was rejected by the artifacts strict schema. Two independent reviews closed it. Here
each property has a **test**, not a comment claiming it.

## It is not durable, and that is stated in the code

`corpus-deep/` or a nested `test/recon-corpus/` still escape — exactly as the #305 comment says a name
list must. The durable fix is to **retry on measured overflow using the diff-byte ranking**
`git-capture-diagnostics.ts` already computes for #311, which keys on the metric that actually breaks
rather than on file size. That design, and the four requirements it must satisfy, are recorded on #368.
This change buys the runs that fix needs, at near-zero risk.

## Tests

| test | pins |
|---|---|
| excludes agent-chosen variants of the generated corpus roots | the reported defect |
| keeps authored trees the corpus prefix must not reach | a nested corpus does **not** take the authored suite with it (#372's fatal defect) |
| keeps a tracked file under a corpus-prefixed path | tracked content is never dropped (#372's other fatal defect) |

## Mutation results

Each verified to apply before its result was trusted:

| mutation | result |
|---|---|
| revert the prefix glob to the exact-name form | **caught** (1 fail) |
| drop root anchoring (`**/${root}*/**`) | **caught** (4 fail) |
| apply the generated exclusion to the tracked listing too | **caught** (2 fail) |

## Gates

- Runtime suite: 437 tests, 420 pass, 15 skipped
- The 2 failures are **pre-existing on `origin/main`** — the same two by name that the #372 review
  independently confirmed against a reverted tree
- prettier and eslint clean on both files

<h3>Greptile Summary</h3>

The PR broadens untracked generated-corpus exclusions from exact root names to prefixes while retaining exact matching for `magic`.
- Applies prefix pathspecs only to the untracked workspace listing.
- Keeps exclusions root-anchored and preserves tracked edits.
- Adds regression coverage for renamed corpus roots, nested authored trees, tracked files, and exact-only `magic` handling.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| packages/runtime/src/workspace-handoff.ts | Changes generated-corpus pathspec construction to prefix-match the two CLI-configurable corpus roots while preserving runtime-root and `magic` exact exclusions. |
| packages/runtime/test/workspace-handoff.test.ts | Adds focused workspace-handoff tests covering prefix exclusions, root anchoring, tracked-file preservation, and exact matching for `magic`. |

</details>

<sub>Reviews (2): Last reviewed commit: ["fix(runtime): narrow the prefix to the t..."](https://github.com/monad-developers/ultrafuzz/commit/675ed71015664e3baad0fcf2f6404383ca15d3db) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=51352356)</sub>

**Context used:**


aviggiano:
Both reviews are in and both raised the same substantive objection. Addressed in `675ed71`.

**What they agreed on, and I accepted.** The three "inherited safety properties" were stated in a way
that concealed the property the prefix *widens*: a NEW (untracked) authored file under a prefix-named
top-level directory is dropped silently, with a validating manifest and a matching result tree — even
inside a directory git already tracks. One reviewer demonstrated it across a real node boundary
(consumer applied the patch cleanly and never received the file); the other showed
`echidna-handlers/`, `magic-numbers/` and `magically/deep/nest/` all vanishing. I reproduced all of it
before changing anything.

**`magic` is back to an exact match.** R53 produced `recon-corpus-deep/` and `echidna-deep/` and nothing
else. Those two are the roots the prompts expose as CLI flags — the lever the agent actually pulled.
`magic/` is a fixed destination the prompts name literally, no variant has ever been observed, and
`magic` is a common enough word that widening it costs authored directories to protect against nothing.
Measured before and after: `magic-numbers/` and `magically/deep/nest/` were being dropped, and now
survive. Adding a prefix without an observation is how a name list acquires collateral damage.

**The cost is now a test rather than a sentence.** `documents the prefix's cost` asserts the loss
directly — same directory, three edits, two outcomes, decided purely by tracked-vs-untracked. If the
durable measurement-based fix lands, that test should start failing and then be deleted; that is the
intended signal, not a regression.

**The comment now states what the code does.** It previously claimed "an authored directory that merely
starts with a generated root's name is not corpus", which the code contradicts. That was the fourth time
in this issue a comment of mine described behaviour the code did not have, in the PR written to avoid
exactly that pattern. It now records the widening, that a pathspec without `:(glob)` lets `*` cross `/`
so the drop reaches arbitrarily deep, and the three escapes the earlier list omitted: top-level generated
FILES, case variants (`Echidna-deep/` is kept), and aggregate overflow.

**Two metric slips corrected.** ">=33.8 MB by itself" is >=33.8 MB of DIFF — on-disk size was never
measured, and conflating those two was the central lesson of the #372 postmortem. And a fixture labelled
"exactly as on the real target" contained a `magic-2` variant that was never observed anywhere.

**Test 43 relaxed** on a reviewer's point: it asserted the nested corpus *is* present, which pins an
acknowledged leak as expected behaviour and would fail a future rule that legitimately excludes it. It
now pins only that the authored file survives.

**One process note I owe the reviewers.** I ran my mutation matrix inside `/tmp/uf368b` while a review
was reading it, so one reviewer's first probe pass silently measured a mutated tree. They caught it,
re-ran in an isolated `git archive` copy, and said so. That was my error and it could have invalidated a
review without either of us noticing; mutation runs belong in a throwaway copy.

Now 46 tests, 46 pass. Mutations still caught: reverting the prefixes (1 fail), widening `magic` (1 fail),
dropping root anchoring (4 fail).

PR #387, opened 2026-08-08
fix(runtime): let a deduplicated campaign finding explain every failure it merged

aviggiano:
Fixes #385.

## The contradiction

`.ultrafuzz/prompts/strategies/invariants/invariant-testing-campaign.md` asks the
campaign agent for two things:

- step 4 — *"Finalize the backend record before deduplicating failures. Preserve
  the originating backend and raw record reference on every pre-deduplication
  failure…"* → `recon-fuzzer-results.json` holds every raw failure.
- step 5 — *"For each **unique** failure, write one finding object in
  `findings.json`."* → `findings.json` holds one entry per deduplication group.

`campaignFindingReferenceDiagnostics` enforced a third rule that contradicts
both: it joined failure → finding on ID equality alone, so it required one
finding per *pre-deduplication* failure.

R55 (`aave-v4-…-issue378fix-r55-invariant-only`) died on this after a 66-minute
campaign. Recon found 29 reproduced failures collapsing into 2 groups (27 on
`property-1`, 2 on `property-3`). The agent wrote exactly the 2 findings step 5
asks for, each reusing a contributing failure's ID and listing every contributing
raw ID in `contributing_backend_failures`. The gate raised
`PROPERTY_FINDING_REFERENCE_MISSING` for the other 27 and failed the node
terminally — `events.jsonl` shows `node-artifacts-missing` immediately followed
by `findings-normalized {"count": 2}` and `artifact-manifest-written`, so every
artifact was present and valid and only the cross-artifact join rejected them.

A retry could never have helped: any correct deduplication fails the same way,
and the only behaviour that passes is not deduplicating at all, which pushes 29
near-identical entries into `dedupe-findings`, `triage` and `final-report`.

## The fix

- Join a failure to its finding by **ID equality or** membership in that
  finding's `contributing_backend_failures` array.
- A deduplicated finding must carry every property ID its contributing failures
  carry (subset, because one merged finding may span several properties). The
  plain one-to-one join keeps its existing strict `property_ids` set equality.
- Two findings claiming the same raw failure are still
  `PROPERTY_FINDING_REFERENCE_AMBIGUOUS`; a raw failure no finding claims is
  still `PROPERTY_FINDING_REFERENCE_MISSING`; the dangling-finding check now also
  accepts a finding explained through its contributing list.
- The prompt now names `contributing_backend_failures` and states the join that
  actually holds after deduplication, replacing *"A failure and its final
  deduplicated finding must share the same ID."*

[...]

Issue #385, opened 2026-08-08
Deduplicating campaign failures is unsatisfiable: the finding-join gate demands one finding per pre-deduplication failure


aviggiano:
## What happened

R55 (`aave-v4-v0012-main-issue378fix-r55-invariant-only`, trial-1) failed
`stateful-invariant-campaign` at 2026-08-08T05:23:06Z with
`provenance.failure.category = "artifact-contract"`, after a 66-minute agent
attempt that produced a complete, reproduced campaign:

```
"Property-derived campaign failure \"failure-2\" has no resulting finding with the same ID;
 ... failure-3 ... failure-29 (27 of them)"
```

`provenance.findings_count` is 2, and the node is terminal (`retry_count: 0`,
`status: "failed"`). The run is still `running` only because `dedupe-findings`
had already been unblocked by the normalized findings.

## Root cause: the gate forbids the deduplication the prompt requires

`recon-fuzzer-results.json` has **29 pre-deduplication failures** collapsing into
**2 deduplication groups**:

| deduplication group | failures |
| --- | --- |
| `property-1` | 27 (`failure-1`…`failure-24`, `failure-26`…`failure-28`) |
| `property-3` | 2 (`failure-25`, `failure-29`) |

`findings.json` has exactly 2 entries — `failure-1` ("Hub total owed exceeded Hub
added assets") and `failure-25` ("Hub added assets exceeded summed Spoke
assets") — each reusing one contributing failure's ID and listing every
contributing pre-deduplication failure ID:

```json
{ "id": "failure-1", "property_ids": ["property-1"],
  "contributing_backend_failures": ["failure-1", "failure-2", ..., "failure-28"] }
```

That is exactly what `.ultrafuzz/prompts/strategies/invariants/invariant-testing-campaign.md`
asks for:

- step 4 — *"Finalize the backend record before deduplicating failures. Preserve
  the originating backend and raw record reference on every pre-deduplication
  failure and preserve all contributing backend provenance on the final
  deduplicated finding."* → the backend record holds all 29.
- step 5 — *"For each **unique** failure, write one finding object in
  `findings.json`."* → 2 findings.

But the same prompt also says *"A failure and its final deduplicated finding must
share the same ID"*, and `campaignFindingReferenceDiagnostics`
(`packages/runtime/src/artifact-gates.ts:2801-2855`) enforces that literally: it
joins failure → finding on **ID equality only**, so it demands one finding per
[...]

PR #388, opened 2026-08-08
Let a campaign deduplicate counterexamples into one finding (#386)

### LINKED ISSUE #386: Campaign gate demands one finding per raw fuzzer counterexample, so correct deduplication fails the node
Found in R55 (`aave-v4-v0012-main-issue378fix-r55-invariant-only`, image from `6fd2362`). This is the first
run in this investigation whose fuzzing campaign actually produced invariant violations, and the gate rejected
the result.

## What happened

`stateful-invariant-campaign` failed with `retry_count: 0` and 27 diagnostics of the form:

```
Property-derived campaign failure "failure-2" has no resulting finding with the same ID
... "failure-3" ... through "failure-29"
```

## The evidence, read off the run volume

```
artifacts/stateful-invariant-campaign/recon-fuzzer-results.json : 29 failures
    property_ids distribution                                   : {property-1: 27, property-3: 2}
artifacts/stateful-invariant-campaign/findings.json             : 2 findings
    failure-1  property_ids=[property-1]  "Hub total owed exceeded Hub added assets"
    failure-25 property_ids=[property-3]  "Hub added assets exceeded summed Spoke assets"
artifacts/stateful-invariant-campaign/campaign-summary.json     : 0 failures
```

29 raw counterexamples represent **two** distinct property violations. The agent deduplicated them into two
findings, one per violated property, and named them after the property each came from.

## Why this is a gate defect and not a producer defect

`campaignFindingReferenceDiagnostics` (`packages/runtime/src/artifact-gates.ts:2801`) reads failures from
`campaignResultArtifactNames` (`:1569`):

```ts
const campaignResultArtifactNames = ["recon-fuzzer-results.json", "echidna-results.json", "medusa-results.json"];
```

Those are **raw fuzzer outputs** — one entry per counterexample. The check then requires, for every failure
carrying `property_ids`, a finding with the *same `id`*:

```ts
if (failurePropertyIds.length > 0 && matchingFinding === undefined) {
  // PROPERTY_FINDING_REFERENCE_MISSING
}
```

A fuzzer that finds one broken invariant 27 different ways emits 27 failures. Satisfying this contract
requires 27 near-duplicate findings — which is precisely what `dedupe-findings` exists to undo, and which
would inflate the benchmark's finding count by the fuzzer's shrink behaviour rather than by the number of
bugs. The contract cannot currently express "many counterexamples, one bug".

Worth stating plainly, because the opposite call was correct once already: in #285 the fan-in fabricated
reference-expectation identifiers and the gate was right to reject them. This is not that. Here the producer's
output is coherent and the evidence is the `property_ids` distribution — 27 of 29 failures share one property.

## Impact

This is on the findings path, so it decides the benchmark's headline number. A run that genuinely violated two
invariants 29 times fails its node, and the two real findings do not get the clean provenance the contract was
meant to guarantee. The run continues to `final-report`, so a `report.md` can still be produced — but with a
failed node, which under our own counting rule makes the run blocked rather than measured.

## Suggested direction (not prescriptive)

The relation between failures and findings is many-to-one, not one-to-one. Options worth weighing:

- Require every property-derived failure to be *represented* by some finding covering the same
  `property_ids`, rather than by a finding sharing its `id`.
- Or have findings carry an explicit `failure_ids: string[]` and require that the union covers every
  property-derived failure — which keeps provenance auditable while allowing deduplication.

Either way the ambiguity check above it (`PROPERTY_FINDING_REFERENCE_AMBIGUOUS`, which fires when one id maps
to several findings) needs to stay meaningful under the new relation.

## Acceptance criteria

- [ ] A regression test that fails without the fix: N counterexamples sharing one `property_id`, deduplicated
      into one finding, passes the gate.
- [ ] A test that a property-derived failure whose property is covered by NO finding still fails — the gate
      must not become vacuous.
- [ ] A test that the ambiguity/mismatch checks still fire under the new relation.
- [ ] The R55 artifact shape above is used as a fixture, so the exact real-world case is pinned.

aviggiano:
Fixes #386.

## The defect

`campaignFindingReferenceDiagnostics` required every property-derived campaign failure to have a resulting finding **sharing its ID**. A fuzzer emits one failure record per *counterexample*, so a campaign that correctly deduplicates many counterexamples of one property into a single finding fails the gate.

R55's `stateful-invariant-campaign` produced exactly this shape, read from the run's durable volume:

```
recon-fuzzer-results.json : 29 failures, property_ids {property-1: 27, property-3: 2}
findings.json             : 2 — failure-1  (property-1) "Hub total owed exceeded Hub added assets"
                                failure-25 (property-3) "Hub added assets exceeded summed Spoke assets"
```

27 spurious `PROPERTY_FINDING_REFERENCE_MISSING` diagnostics failed the node. Unlike #285, where the gate was right and the model was wrong, here the gate is wrong — the `property_ids` distribution is the evidence that the deduplication was correct.

## The fix

Require that every property a failure exercises is **covered by some finding's `property_ids`**, rather than that some finding shares the failure's ID.

The gate does not become vacuous:

- A property-derived failure whose property no finding covers is still an error, and the diagnostic now names the uncovered property IDs instead of only the failure ID.
- A finding that *does* claim a failure's ID must still carry that failure's exact property set, so `PROPERTY_FINDING_REFERENCE_MISMATCH` and `PROPERTY_FINDING_REFERENCE_AMBIGUOUS` are untouched.

## Tests

Before this PR, **no test anywhere covered `PROPERTY_FINDING_REFERENCE_MISSING` or `PROPERTY_FINDING_REFERENCE_AMBIGUOUS`** — the one nearby test used a single failure with a single same-ID finding, which is why the 1:1 assumption survived. Five tests added; the first two fail on `main`:

1. Three counterexamples of one property deduplicated into one finding passes the gate.
2. **The R55 artifact shape as a fixture** — 29 failures over two properties, two findings — passes the gate.
3. A property-derived failure whose property no finding covers still fails (non-vacuity).
4. A partially covered multi-property failure fails and the message names the uncovered property.
5. Ambiguous same-ID findings and mismatched property sets still fire.

`packages/runtime` suite: 429 passed, 0 failed, 15 skipped. `typecheck` clean; `eslint` clean on both changed files.

## Scope note

This changes only the failure→finding direction. The dual check, `danglingCampaignFindingDiagnostics` (a property-derived finding must share an ID with some failure), is left as-is: R55's findings carry real failure IDs, so no evidence shows that direction misfiring, and loosening it without evidence would weaken a gate that catches invented findings.

<h3>Greptile Summary</h3>

The PR changes campaign artifact validation to support deduplicating multiple property counterexamples into one finding while retaining property-provenance checks.
- Replaces per-failure same-ID coverage with per-finding property-set coverage.
- Rejects findings that claim properties absent from all campaign failures.
- Clarifies campaign prompt requirements for deduplicated findings and multi-property observations.
- Adds regression coverage for deduplication, partial coverage, property combinations, mismatches, ambiguity, and invented properties.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| packages/runtime/src/artifact-gates.ts | Changes campaign failure-to-finding validation to property-set coverage and adds a complementary diagnostic for unobserved finding properties. |
| packages/runtime/test/artifact-gates.test.ts | Adds comprehensive regression cases for deduplicated counterexamples and retained provenance constraints. |
| .ultrafuzz/prompts/strategies/invariants/invariant-testing-campaign.md | Clarifies that findings represent unique root causes, reuse one contributing failure ID, and preserve all contributing properties. |
| packages/prompts/test/semantic-anchors.test.ts | Adds semantic anchors preventing the campaign prompt from drifting back to one finding per counterexample. |

</details>

```mermaid
flowchart LR
  F[Campaign failures] --> P[Collect observed property IDs]
  F --> C[Check each failure's property set]
  G[Deduplicated findings] --> C
  G --> U[Reject unobserved property claims]
  P --> U
  C --> D[Property provenance diagnostics]
  U --> D
```
</details>

<sub>Reviews (4): Last reviewed commit: ["Ground a finding&#39;s property claims in th..."](https://github.com/monad-developers/ultrafuzz/commit/224c2aed2030068d0079b12de237c0e149b1009a) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=51398817)</sub>

**Context used:**


aviggiano:
Production confirmation, read live from R55's run state rather than reconstructed:

```
FAILED stateful-invariant-campaign | Property-derived campaign failure "failure-2" has no resulting finding
with the same ID; Property-derived campaign failure "failure-3" has no resulting finding with the same ID;
Property-derived campaign failure "failure-4" has no resulting finding with the same ID; ...
```

That is the exact diagnostic this PR removes for the deduplicated case, and it is the only reason the node failed. R55's run: `aave-v4-v0012-main-issue378fix-r55-invariant-only`, image `im-01KZFF97HB1BGSFJ7134HXYPS4`, source revision `6fd2362`.

Worth recording because it narrows the blast radius: the failure does **not** stall the DAG. R55 continued — `dedupe-findings` succeeded at 05:28Z and `triage` is running — so the cost of this defect is that the campaign node is marked failed and its `findings.json` is the artifact of a failed node, not that the pipeline halts.

aviggiano:
Two independent reviews landed and converged. Both confirmed the diagnosis against the producer contract rather than taking it on trust: `invariant-testing-campaign.md` says "**Finalize the backend record before deduplicating failures**" and "For each **unique** failure, write one finding object in `findings.json`". So the backend record is pre-deduplication by construction and the 1:1 gate encoded a reading the prompt never asked for. The alternative hypothesis — that the producer should have emitted one finding per counterexample — is refuted.

Both also found the same two gaps in my first commit. Fixed in `eb39119`, each pinned by a test that was red first:

**1. Coverage was judged against the union of all findings.** A counterexample that broke two invariants at once passed as long as two separate single-property findings existed, even though no finding claimed the combination — the most interesting counterexample in the file, silently dropped. Coverage is now judged **per finding**: a finding covers a failure only when it claims every property that failure broke.

**2. `MISMATCH` still enforced the 1:1 assumption.** A deduplicated finding reuses one of its failures' IDs, so it legitimately carries more properties than that one failure. Exact set equality meant the node failed again for the same underlying reason #386 describes, just under a different code. The anchoring finding must now **cover** the failure's properties, not equal them; dropping one is still an error.

**3. The prompt sentence the gate was written from.** `"A failure and its final deduplicated finding must share the same ID"` reads distributively as 1:1. Relaxing only the gate would leave *both* shapes acceptable — so a future run could emit 29 findings for 29 counterexamples and inflate the finding count with the fuzzer's shrink behaviour, which is the outcome this issue argues against. The prompt now says to reuse the ID of one covered failure, carry every property those failures reported, and not emit one finding per counterexample.

Deliberately **not** fixed here, filed instead:

- #389 — campaign backend provenance is keyed by failure ID, which deduplication makes unsatisfiable. Unreachable under the shipped contract ("recon-fuzzer is the single final bug-finding backend"), so widening the report join on speculation would be worse than recording it. Also captures two pre-existing join bugs the review turned up.
- #390 — `campaign-summary.json` reported 0 failures next to a record of 29 in R55, and no gate reads that file at all.

Runtime suite: 431 passed, 0 failed, 15 skipped (7 new tests). prettier and eslint clean on all three changed files.

aviggiano:
### These two PRs fix the same defect in incompatible ways, and they conflict

R55's `stateful-invariant-campaign` failure got filed twice — #385 and #386 — and each has its own
open, `MERGEABLE` fix: #387 (`fix/campaign-dedupe-finding-join`) and #388 (`fix/386-campaign-finding-dedup`).
Both are mergeable *against `main`*, which says nothing about each other. Test-merged locally:

```
$ git merge --no-commit --no-ff pr388   # onto pr387
CONFLICT (content): Merge conflict in .ultrafuzz/prompts/strategies/invariants/invariant-testing-campaign.md
CONFLICT (content): Merge conflict in packages/runtime/src/artifact-gates.ts
Automatic merge failed
```

All three shared files collide (`artifact-gates.test.ts` auto-merged but stacks two different
test suites for the same function). Whichever lands first breaks the other.

### They are not the same fix

Both relax `campaignFindingReferenceDiagnostics`, but on different joins:

| | #387 | #388 |
| --- | --- | --- |
| join key | explicit — `finding.contributing_backend_failures` must list the failure ID | implicit — some finding's `property_ids` must cover the failure's |
| `contributing_backend_failures` | read, and becomes the contract | still unread anywhere in the repo |
| a failure matched by an unrelated finding | rejected — no ID link | accepted — any finding carrying `property-1` satisfies any `property-1` failure |
| `danglingCampaignFindingDiagnostics` | carve-out added for merged findings | untouched |

R55's producer *did* emit `contributing_backend_failures` on both findings, listing all 27 and 2
contributing failures respectively. Under #387 that field is what makes the artifact valid; under #388
the artifact passes without it ever being read, so the provenance stays inferred — which is the same
inference #389 identifies as unsatisfiable one join downstream.

Not arguing for either here — flagging that this needs one decision, not two merges. Suggest closing the
losing issue as a duplicate at the same time.

Cross-ref: #385, #386, #387, #388, #389.

aviggiano:
Coordination note from the R55 volume, not a review objection — **#388 and #387 fix the same defect and cannot both land.**

#385 and #386 are the same defect filed twice, and this PR and #387 are two independent fixes for it. Both are `MERGEABLE` against `main` individually — mergeability is measured against `main`, not against each other — but they conflict directly:

```
$ git merge-tree --write-tree pr/387 pr/388   # merge base 6fd2362
CONFLICT (content): Merge conflict in .ultrafuzz/prompts/strategies/invariants/invariant-testing-campaign.md
CONFLICT (content): Merge conflict in packages/runtime/src/artifact-gates.ts
Auto-merging packages/runtime/test/artifact-gates.test.ts
```

Whichever merges first, the other needs a rebase or a close.

## Both accept the real R55 artifact

I replayed each PR's join rule over the actual files on the run volume — `artifacts/stateful-invariant-campaign/{recon-fuzzer-results.json,findings.json}`, 29 failures and 2 findings:

| | missing | ambiguous | mismatch | unobserved |
| --- | --- | --- | --- | --- |
| #388 (property-coverage join) | 0 | n/a | 0 | 0 |
| #387 (`contributing_backend_failures` join) | 0 | 0 | 0 | n/a |

Two details specific to this PR, both clean on the real data: `unobservedFindingPropertyDiagnostics` fires nothing, because the findings claim exactly `property-1` and `property-3` and both were reported by failures; and `danglingCampaignFindingDiagnostics` — which this PR leaves demanding ID equality — also fires nothing, because both findings are named after a contributing failure (`failure-1`, `failure-25`). That constraint is documented in this PR's prompt edit ("Give the finding that deduplicates a group of failures the ID of one of the failures in that group"), so the asymmetry between the relaxed forward join and the unchanged reverse join is intentional rather than a gap. Worth stating explicitly, since it is the one rule a producer could still trip on that the gate's error message would attribute to the finding rather than to the naming rule.

## The difference that decides it

- This PR needs **no new producer field** and stays non-vacuous via the unobserved-property check. The cost is that the failure→finding link is inferred from property sets rather than recorded, so you can no longer ask which of the 29 raw counterexamples a finding came from.
- #387 joins on `contributing_backend_failures`, keeping that link auditable, at the cost of requiring the producer to emit the field — a campaign that deduplicates without it fails exactly as R55 did.

Both prompts document their own rule, so neither is unsound. It is a provenance-strictness call, not a correctness one.

## Sequencing

R55 is terminally over — run `status: "failed"` at `2026-08-08T06:08:18.364Z`, controller lease expired after 11 recovery attempts, no live container. It hit **two** gate defects back to back: this one at `05:23:06` on `stateful-invariant-campaign`, then #392 at `05:54:09` on `final-report`. A run needs both this and #393 to land before it can reach a report.

PR #393, opened 2026-08-08
Say that report blocker_summaries are strings (#392)

### LINKED ISSUE #392: final-report emits typed blocker_summaries objects but the report contract requires strings, failing the run
R55 (`aave-v4-v0012-main-issue378fix-r55-invariant-only`, image `im-01KZFF97HB1BGSFJ7134HXYPS4`, source revision `6fd2362`) reached `final-report` and **failed terminally**:

```
final-report | artifact-contract failure for report.json (ultrafuzz/report@1):
               Invalid input at report.json#property_implementation_coverage
```

`report.json` and `report.md` were both written in the node workspace and then discarded, so the run produced no report at all. This is the last node in the pipeline, so this defect alone converts a complete run into a failed one.

## Cause

The model emitted `blocker_summaries` as **typed objects**:

```json
{
  "property_id": "property-10",
  "status": "deferred",
  "code": "transition-oracle-deferred",
  "summary": "This property requires exact before/after transition accounting ...",
  "next_action": "Add property-scoped snapshots around the corresponding handler ..."
}
```

34 of them, one per deferred property. The contract requires strings:

```ts
// packages/artifacts/src/artifact-contracts.ts
blocker_summaries: z.array(z.string().min(1)).optional()
```

and the runtime gate's own canonical form is the string `` `${propertyId}: ${record.blocker.summary}` `` (`packages/runtime/src/artifact-gates.ts`, in the coverage expectation block).

## Why the producer's reading is defensible — this is the #386 class of defect

`.ultrafuzz/prompts/review/final-report.md` gives the shape as an example with an **empty** array, which carries no element-type information:

```json
"blocker_summaries": []
```

and then says:

> Include `blocker_summaries` for selected records whose status is `blocked`, `pending`, or `deferred`, **using each typed blocker summary**.

The upstream handoff record has a *typed blocker object* with `code`, `summary`, and `next_action`. "Using each typed blocker summary" reads naturally as "use the typed blocker-summary object", which is exactly what the model did. Nothing in the prompt states the required `"<property-id>: <summary>"` string form.

So the prompt and the contract disagree, the model followed the prompt, and the gate killed the run — the same failure mode as #386, in a different artifact.

## Suggested fix

- State the exact element format in `final-report.md` and show a non-empty example, so the prompt and the contract encode one contract.
- Regression tests: the contract rejects object elements with a diagnostic naming `blocker_summaries`; the documented string form passes; and a semantic anchor pins the format sentence so a future prompt edit cannot silently drop it.

## Related observation

The gate computes `expected.blocker_summaries` in the canonical string form but only ever checks `Array.isArray(coverage.blocker_summaries)` — the computed value is never compared to what the report carries. The canonical format therefore exists only in code that does not enforce it, which is why the prompt was the only place the model could have learned it. Tightening that comparison is only safe once the prompt states the format exactly; enforcing a format the prompt never states is what produced #386.

aviggiano:
Fixes #392.

## What happened

R55 (`aave-v4-v0012-main-issue378fix-r55-invariant-only`, image `im-01KZFF97HB1BGSFJ7134HXYPS4`, source revision `6fd2362`) failed terminally at `final-report`, the last node in the pipeline:

```
final-report | artifact-contract failure for report.json (ultrafuzz/report@1):
               Invalid input at report.json#property_implementation_coverage
```

30 nodes had succeeded. `report.json` and `report.md` were both written inside the node workspace and then discarded, so the run produced no report at all.

## Cause

The model emitted `blocker_summaries` as 34 typed objects:

```json
{
  "property_id": "property-10",
  "status": "deferred",
  "code": "transition-oracle-deferred",
  "summary": "This property requires exact before/after transition accounting ...",
  "next_action": "Add property-scoped snapshots around the corresponding handler ..."
}
```

The contract takes `z.array(z.string().min(1))`, and the runtime gate's own canonical form is the string `` `${propertyId}: ${record.blocker.summary}` ``.

## Why this is a plumbing defect and not a model error

The prompt showed the field only as an **empty array** — carrying no element-type information — and then said to include the summaries *"using each typed blocker summary"*. The upstream handoff record holds a *typed blocker object*, so that phrase reads as "use the typed blocker-summary object", which is what the model did. Nothing in the prompt stated the string form.

Same shape as #386: the prompt and the artifact contract encode different contracts, the model follows the prompt, and the gate kills the run.

## The fix

State the element format in the prompt, show it non-empty, and say explicitly that the typed blocker's other fields stay in the handoff record.

## Tests

- **The red one:** a semantic anchor pinning the format sentence. Verified red by stashing only the prompt edit and rerunning — `requires final reports to preserve benchmark expectation coverage` fails without it.
- Contract assertions for both sides: the exact object shape R55 emitted is rejected, the documented string form is accepted. `blocker_summaries` had **no test coverage at all** before this, in either direction.

`@ultrafuzz/prompts` 46 passed; `@ultrafuzz/artifacts` 73 passed; eslint and prettier clean.

## Scope note

The gate computes `expected.blocker_summaries` in the canonical string form but only ever checks `Array.isArray` — the computed value is never compared against what the report carries, which is why the prompt was the only place the model could have learned the format. Tightening that comparison is deliberately **not** done here: enforcing a format the prompt never stated is exactly how #386 happened, and this PR makes the prompt state it first.

<h3>Greptile Summary</h3>

The PR aligns final-report instructions with the report artifact contract and improves schema diagnostics for malformed union fields.
- Documents the canonical string representation and ordering of blocker summaries.
- Adds exact JSON and Markdown coverage examples with semantic and runtime tests.
- Expands nested union validation issues to report more specific paths.
- Allows blocker-summary Markdown in escaped or verbatim form while still rejecting reworded summaries.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| .ultrafuzz/prompts/review/final-report.md | Documents the exact JSON and Markdown representation required for property implementation coverage and blocker summaries. |
| packages/artifacts/src/artifact-contracts.ts | Expands Zod union issues to surface the most specific available schema-validation paths. |
| packages/artifacts/test/schema.test.ts | Verifies typed blocker objects are rejected, canonical strings are accepted, and diagnostics identify the affected field. |
| packages/prompts/test/semantic-anchors.test.ts | Pins the prompt language that communicates blocker-summary and Markdown coverage contracts. |
| packages/runtime/src/artifact-gates.ts | Accepts blocker summaries rendered either through public-prose escaping or verbatim with normalized whitespace. |
| packages/runtime/test/artifact-gates.test.ts | Exercises the prompt examples through the runtime gate and verifies escaped, verbatim, and reworded blocker behavior. |

</details>

<sub>Reviews (3): Last reviewed commit: ["Stop asking a model to reproduce a JS es..."](https://github.com/monad-developers/ultrafuzz/commit/5945d8caf3378713f2c67c491309f6f6950f3abc) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=51402107)</sub>

</details>


aviggiano:
Two independent reviews landed. Both confirmed strings are the right choice — widening the contract to accept objects would be actively harmful, since `stringArray` filters non-strings, so object elements would silently become an empty list and fail `sameStringSequence` anyway, and the CLI markdown renderer would drop them too. No consumer anywhere reads `code` or `next_action` from the report.

One review corrected a factual error in the issue I filed, and it changed the fix. I wrote that the gate "only ever checks `Array.isArray`". It does not — it compares `blocker_summaries` byte-for-byte via `sameStringSequence` against `` `${propertyId}: ${record.blocker.summary}` ``. Saying the elements are strings was necessary but not sufficient: a paraphrase fails the gate exactly as an object does. Fixed in `bdaafca` — the prompt now says to copy the summary verbatim, and the example is a realistic full sentence rather than a short lowercase fragment that invites paraphrase. I've posted the correction on #392 as well.

The bigger gap was the **Markdown** half of the same block. It is compared line by line against the JSON, and no prompt in the tree described any of it — not the bullet labels, not the backticks, not the `<br>` join for included priorities, not the bare `Blocker summaries:` line, not the fact that each blocker bullet is escaped public prose. Grepping the prompts for those literals returned nothing. R55 died at the JSON contract before reaching that comparison, so the next run would have been the first to meet it, and a blocker summary naming a function with an underscore would have been enough to fail the node. A separate sweep for this defect class independently ranked that format as the single highest-risk unpinned coupling in the repo. Now documented exactly, with anchors.

Also in `bdaafca`: union failures are flattened to the branch the value came closest to matching. R55's entire run ended on `Invalid input at report.json#property_implementation_coverage`, naming neither the field nor the reason, because the union hid the object branch's issues. The new assertion pins that the diagnostic now reaches `property_implementation_coverage.blocker_summaries`. This helps the second union on `property_provenance` too.

**Correction to my PR description:** I wrote that `blocker_summaries` "had no test coverage at all before this, in either direction". That is inaccurate — `packages/runtime/test/artifact-gates.test.ts` already covered the canonical string on both the JSON and Markdown sides, including the escaping rule. It had no *contract-schema* coverage. The format was pinned in code and tests; only the prompt lagged.

Filed rather than fixed here, from the sweep: #394 (`strategy_provenance` shape required by the canonical renderer, which throws), #395 (`severity_guess` vocabulary unconstrained upstream but enforced on non-production records), #396 (`fuzzer_backend: "unavailable"` rejected by the backend join).

`@ultrafuzz/artifacts` 73 passed, `@ultrafuzz/prompts` 46 passed, `@ultrafuzz/runtime` 424 passed / 0 failed. typecheck, eslint, prettier clean.

aviggiano:
Two more reviews, one of which built the report my prompt described and ran the real gate against it rather than reading along. That was the right call, and it changed the fix.

**The prose rules passed — and were wrong in three ways I could not see by reading the code.** `reportPublicProse` escapes the brackets it injects, so the placeholder is `\[redacted-path\]`, not `[redacted-path]`. It redacts the *relative* prefixes `.ultrafuzz/`, `artifacts/`, `workspaces/`, `generated-tests/`, while *not* redacting absolute paths outside a fixed allowlist — so "any absolute filesystem path" was wrong in both directions. And it runs a secret-redaction pass first that my paragraph never mentioned.

Three attempts to describe that function produced three wrong descriptions. I take that as the answer rather than as a reason to try a fourth time: it is not a describable contract, and a model has less of the code in front of it than I did. **The gate now accepts the blocker bullet either as the escaped rendering or as the summary text with whitespace collapsed**, and the prompt says exactly that. The join the check exists to protect is untouched — a bullet reporting a *different* blocker than the JSON still fails, asserted in the same test as both accepted forms.

**The examples were mutually contradictory and described an unreachable state**: one included priority in JSON against two in Markdown, a blocker summary beside zero blocked/pending/deferred properties, and reference expectation IDs without the property IDs that produce them. A model copying them could not have passed. They are now one coherent example — and a new test feeds those exact blocks *from the prompt* through the real gate instead of restating them, so they cannot drift apart again. That test failed first, on a fixture defect it correctly caught.

Also fixed: on a depth tie, union expansion now reports every branch instead of the first. A number under `property_implementation_coverage` was reported as `expected "unavailable"` — advice that would pass the contract and then fail the gate requiring an object.

Anchors now pin all eight bullet labels rather than two, match on whitespace instead of hard-wrap positions (the previous ones would have broken on an innocent reflow and survived a gutted meaning), and pin the laxer escaping promise so the gate and prompt cannot separate again.

Rebased onto `main` now that #388 is merged. `@ultrafuzz/runtime` 435 passed / 0 failed, `@ultrafuzz/artifacts` 73 passed, `@ultrafuzz/prompts` 46 passed; prettier and eslint clean.

Issue #391, opened 2026-08-08
Nothing relates the number of campaign findings to the number of distinct failures observed


aviggiano:
Surfaced by two independent adversarial reviews of #388, which fixed the opposite defect (#386: the gate demanded one finding per counterexample). Filing rather than fixing inside #388 because closing it properly needs a schema field that does not exist.

## The gap

After #388 the campaign gate guarantees:

> every property ID any failure reported — and every co-occurring *combination* of property IDs — appears in some finding's `property_ids`; no finding anchored to a failure drops that failure's properties; and no finding names a property no failure reported.

It does **not** guarantee any relationship between how many distinct failures the fuzzer observed and how many findings the campaign reports. The predicate is set-based and cardinality-free, so for any campaign record a single finding whose `property_ids` is the union of every failure's properties, anchored to any one failure's ID, passes cleanly:

```jsonc
// failures: f-1 [property-1], f-2 [property-2], f-3 [property-3], f-4 [property-4, property-5]
// findings:
[{ "id": "f-1", "property_ids": ["property-1","property-2","property-3","property-4","property-5"], ... }]
```

Five distinct violations, one finding, zero diagnostics. On `main` before #388 this produced four errors — but only as a side effect of the 1:1 rule that made correct deduplication impossible, so it was never a check we could keep.

This matters concretely for the benchmark: `packages/cli/src/benchmark-analysis/lib/analysis.ts` maps one finding instance to at most one matched candidate and one root-cause cluster. Two ground-truth bugs collapsed into one finding can earn at most one cluster's credit, so over-collapse shows up as a **false negative** in ground-truth scoring — the failure mode this pipeline exists to avoid.

## Why it cannot be closed today

The fuzzer's own output cannot distinguish "29 shrink duplicates of one bug" (the real R55 case) from "29 genuinely distinct bugs that all break the accounting invariant". Both are the same JSON. `propertyCampaignFailureSchema` (`packages/artifacts/src/property-provenance.ts`) carries only `id`, `status`, `property_ids`, and `findingSchema` (`packages/artifacts/src/findings-schema.ts`) has no back-reference to the failures a finding deduplicates. The join is inferred, so the gate has nothing to verify against.

## Suggested direction

[...]

Issue #390, opened 2026-08-08
campaign-summary.json failure counts are never validated against the backend record


aviggiano:
Observed in R55 (`aave-v4-v0012-main-issue378fix-r55-invariant-only`, image `im-01KZFF97HB1BGSFJ7134HXYPS4`, source revision `6fd2362`), in the same evidence block that produced #386:

```
recon-fuzzer-results.json : 29 failures, property_ids {property-1: 27, property-3: 2}
findings.json             : 2 findings
campaign-summary.json     : 0 failures
```

`invariant-testing-campaign.md` requires the summary to carry "pre- and post-deduplication failure counts". R55's summary reported **0** while the backend record held 29 observed failures and `findings.json` held 2. Both correct counts were available in sibling artifacts in the same node.

Nothing gates this: grepping `packages/**/src` for `campaign-summary` returns no hits, so no artifact gate reads the file at all. A summary that says zero next to a record of 29 is exactly the shape that makes a run look clean when it is not, and it is the artifact a human reads first.

Suggested fix: gate `campaign-summary.json` against the sibling backend records — pre-deduplication count must equal the total failures across backend records, post-deduplication count must equal the number of findings. Add a regression test using the R55 shape above.

Distinct from #386 (which was the failure→finding join) and from #389 (backend provenance keying).

aviggiano:
**The counts this issue wants gated already exist per deduplication group.** R55's campaign `findings.json` carries, on each finding:

```jsonc
"deduplication": { "key": "property-1", "pre_dedup_count": 27, "representative_failure_id": "failure-1", "basis": "..." }
```

27 and 2 across the two findings, summing to exactly the 29 failures in `recon-fuzzer-results.json`. So the pre- and post-deduplication accounting the prompt requires is present and correct in the artifact the gate does read — it is only `campaign-summary.json`, the file nothing reads, that reported 0.

That makes the suggested gate straightforward and independent of the summary file: require the sum of `deduplication.pre_dedup_count` across findings to equal the number of property-derived failures, and the number of findings to equal the number of distinct deduplication keys. Both are checkable against artifacts already on disk, and both would have caught the 0-versus-29 discrepancy that motivated this issue.

Related corrections on the same evidence: #389, #391.

aviggiano:
[...]

Issue #389, opened 2026-08-08
Campaign backend provenance is keyed by failure ID, which deduplication makes unsatisfiable


aviggiano:
Found while reviewing #388. **Not reachable in the shipped invariant strategy** — `invariant-testing-campaign.md` says "recon-fuzzer is the single final bug-finding backend. Do not run Echidna, Medusa, or any other fuzzer as an additional final backend." Filing so the inconsistency is recorded rather than fixed opportunistically inside an unrelated PR.

`readCampaignFuzzerBackends` (`packages/runtime/src/artifact-gates.ts`) builds a map keyed by **`failure.id`** while naming it `backendsByFinding`, and the final-report join requires `property_provenance[].fuzzer_backends` to equal that map's entry for the report's `finding_id`.

Deduplication makes the two disagree. A deduplicated finding reuses the ID of exactly one of the failures it covers, but the same prompt requires it to "preserve all contributing backend provenance". So for a finding covering `recon:r-1` and `medusa:m-1` and anchored to `r-1`:

- the report must list `["medusa", "recon"]` (prompt), but
- `readCampaignFuzzerBackends` yields `["recon"]` (only `r-1` carries that ID),
- so `PROPERTY_REPORT_FUZZER_BACKEND_MISMATCH` fires, and the only way to satisfy the gate is to under-report backend coverage.

Before #388 this shape died earlier, at the campaign gate, so the contradiction was unobservable. #388 deliberately did not widen the report join, so the shape still cannot arise under the shipped single-backend contract.

Two related observations from the same review, both pre-existing:

1. Failure IDs are unique only *within* one campaign artifact (`packages/artifacts/src/property-provenance.ts`), but `readCampaignFuzzerBackends` unions across artifacts. A coincidental `failure-1` in both `echidna-results.json` and `recon-fuzzer-results.json` attributes both backends to one finding.
2. `readCampaignBackends` in `packages/cli/src/report-artifacts.ts` pairs `recon-fuzzer-results.json` only with logical node `stateful-invariant-recon-campaign`, but the shipped topology writes it under `stateful-invariant-campaign`. For the default topology the CLI therefore resolves zero backends and strips `fuzzer_backend` from the published report, while the runtime gate requires the pre-publish `report.json` to carry `"recon"`. The two joins disagree for the topology we actually ship.

[...]

PR #397, opened 2026-08-08
Read the contributing-failure links a campaign already writes (#391)

aviggiano:
Addresses #391.

## The gap

After #388 the campaign gate related findings to failures by **property coverage**, which is cardinality-free. A single finding carrying the union of every property passes for any campaign, however many distinct counterexamples the fuzzer actually found. Two independent reviews flagged this, and both they and I concluded the exact link "does not exist in the schema" and would need a schema addition.

**That was wrong.** Reading R55's real `findings.json` off the durable volume — rather than reasoning from the schema — shows the producer already volunteers it:

```jsonc
{
  "id": "failure-1",
  "contributing_backend_failures": ["failure-1", "failure-2", ..., "failure-28"],  // 27 entries
  "deduplication": { "key": "property-1", "pre_dedup_count": 27,
                     "representative_failure_id": "failure-1", "basis": "..." },
  "backend_provenance": [{ "fuzzer_backend": "recon", "failure_ids": [...] }]
}
```

The two findings' `contributing_backend_failures` union to **exactly** the 29 failures in the backend record, with no overlap, and the two `pre_dedup_count` values are 27 and 2. `findingSchema` is a loose object, so all of this validated and none of it was ever read.

## What this adds

Three checks over the union of campaign records:

- a finding lists a contributing failure **no campaign record contains**
- **two findings claim the same counterexample** — contradictory provenance
- property-derived failures **claimed by nobody** — the over-collapse signal

Over-collapse matters concretely: `benchmark-analysis` credits at most one ground-truth cluster per finding instance, so silently absorbing distinct bugs into one finding shows up as a **false negative** against ground truth.

## Why these are warnings, not errors

Deliberately, and this is the load-bearing decision in the PR.

The field is *volunteered by the producer*; the prompt does not mandate it. Every defect in this sequence — #386, #392, and the latent Markdown-format one — came from a gate demanding something the prompt never promised, each costing a full run. I have exactly one sample of this field's shape. Adding an error-severity check on that basis would be repeating the mistake with better intentions.

`ok` is `diagnostics.every((d) => d.severity !== "error")`, so these cannot fail a node. They make the disagreement visible in artifacts that are already collected. Promoting them to errors is a decision for after the prompt mandates the field and more than one run has been observed — noted in #391.

## Verification

- New test asserts `ok === true` in every case, that all three codes fire, and that each is `severity: "warning"`; plus a clean case producing no `PROPERTY_CAMPAIGN_*` diagnostics at all.
- **Run against R55's real artifacts** (96 KB `recon-fuzzer-results.json` with all 29 failures, plus the genuine 2-finding `findings.json`): `ok: true`, **0 diagnostics of any severity** — the check does not misfire on production data.
- `@ultrafuzz/runtime`: 436 passed, 0 failed, 15 skipped. eslint and prettier clean.

## Not included

`deduplication.pre_dedup_count` (#390) and `backend_provenance` (#389) are equally unread and equally useful; I left them to their own issues rather than widening this one.

<h3>Greptile Summary</h3>

This PR adds warning-only campaign artifact diagnostics that compare findings’ contributing failure IDs with the union of backend campaign records.

- Warns when findings reference unknown campaign failures.
- Warns when multiple findings claim the same failure ID.
- Warns when property-derived failures are unclaimed.
- Adds coverage for inconsistent and clean provenance inputs.

The cross-backend failure-ID collision needs to be fixed before merging because it can suppress or fabricate the provenance warnings this change is intended to provide.

Campaign failure IDs are only artifact-local, but the new diagnostics flatten all backend records and compare claims using unqualified IDs, causing distinct counterexamples with the same ID to be treated as one.

**Files Needing Attention:** packages/runtime/src/artifact-gates.ts and packages/runtime/test/artifact-gates.test.ts

| Filename | Overview |
|----------|----------|
| packages/runtime/src/artifact-gates.ts | Adds contributing-failure diagnostics, but flattening campaign records loses backend identity and conflates duplicate IDs across backends. |
| packages/runtime/test/artifact-gates.test.ts | Covers unknown, ambiguous, unclaimed, and clean single-backend cases, but omits the already-supported cross-backend duplicate-ID shape. |

</details>

```mermaid
flowchart LR
  R[Recon failures] --> F[Flatten failures]
  E[Echidna failures] --> F
  M[Medusa failures] --> F
  D[Finding contribution IDs] --> C[Compare by failure ID]
  F --> C
  C --> W[Warning diagnostics]
```
</details>

<a href="https://app.greptile.com/ide/claude-code?prompt=%23%23%23%20Issue%201%0Apackages%2Fruntime%2Fsrc%2Fartifact-gates.ts%3A2349-2353%0A**Backend-local%20failure%20IDs%20collide**%0A%0AWhen%20two%20campaign%20artifacts%20use%20the%20same%20failure%20ID%2C%20flattening%20their%20failures%20discards%20backend%20identity%20and%20treats%20both%20counterexamples%20as%20one.%20A%20claim%20for%20either%20failure%20then%20suppresses%20the%20other%20backend's%20unclaimed%20warning%2C%20while%20separate%20legitimate%20claims%20can%20produce%20a%20false%20ambiguous%20warning.%0A%0A---%0A%0AFor%20each%20issue%20above%2C%20determine%20whether%20it%20is%20valid%20and%20should%20be%20fixed.%20If%20so%2C%20fix%20it%20directly.&repo=monad-developers%2Fultrafuzz&pr=397&platform=github"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInClaudeDark.svg?v=6"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInClaude.svg?v=6">Fix All in Claude Code</picture></a>

`````markdown
### Issue 1
packages/runtime/src/artifact-gates.ts:2349-2353
**Backend-local failure IDs collide**

When two campaign artifacts use the same failure ID, flattening their failures discards backend identity and treats both counterexamples as one. A claim for either failure then suppresses the other backend's unclaimed warning, while separate legitimate claims can produce a false ambiguous warning.

---

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["Read the contributing-failure links a ca..."](https://github.com/monad-developers/ultrafuzz/commit/8f102d875c045e85f63689d5999f362d6f422f25) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=51411231)</sub>

**Context used:**


aviggiano:
Closing. A review argued this should not merge, and it is right on two independent counts. I verified both myself rather than taking them on trust.

**1. The warnings are structurally unobservable.** I claimed in the PR body that they "make the disagreement visible in artifacts that are already collected." That is false. `defaultRowSync` in `packages/evals/src/runner.ts` reads `result.diagnostics` only when `!result.ok`, and warning severity guarantees `ok`. The attempt ledger has no diagnostics field, and the gate only runs on a status transition, so nothing can re-surface them later. The comment explaining exactly this is in `packages/runtime/src/workflow-sync.ts` — in the same file I was editing:

> Returning it in `diagnostics` alone is not enough: the eval runner drops `result.diagnostics` whenever `result.ok`, and the Modal worker drains the CLI child's stdout without persisting it.

Someone had already written down why this approach cannot work. I asserted the behaviour of a neighbouring module instead of deriving it — the same error that produced three other corrections in this sequence.

**2. The check does not detect what I said it detects.** `PROPERTY_CAMPAIGN_FAILURE_UNCLAIMED` fires only when a producer declares `contributing_backend_failures` *and declares it incompletely*. A producer that over-collapses lists everything under its one finding and stays silent. #391's canonical example — one finding carrying the union of every property, with no `contributing_backend_failures` at all — produces zero diagnostics here, because `anyFindingDeclares` stays false. So this is a provenance-completeness check, not the over-collapse signal I described in the commit message and PR body.

The reviewer also caught that my "R55 shape" clean-case assertion codifies one finding absorbing every failure as correct. For R55 it *was* correct — 27 shrink duplicates of one property — but the gate cannot distinguish that from over-collapse, which is precisely #391's point.

**The right path**, which I am recording on #391 rather than half-shipping here:

1. Name the field in `invariant-testing-campaign.md` step 4, which today says only "preserve all contributing backend provenance" — vague enough that I reverse-engineered the field name from a single run.
2. Add it to `findingSchema`, required when property-derived.
3. Then gate it as an error, with the cardinality check the issue actually asks for.

Contract → schema → gate, in that order, which is what #386 and #392 were teaching. There is also a cheaper first step: gate `campaign-summary.json`'s pre/post-deduplication counts (#390) — a field the prompt **already mandates** and no gate reads, which gives a real cardinality signal without inventing a contract.

Two secondary defects worth carrying over: `failures` is the concatenation across backend records, so colliding IDs across backends could inflate the unclaimed count and misfire the ambiguity check; and `String(finding.id)` yields the literal `"undefined"` for a finding without an ID, since `findingSchema` is loose.

PR #399, opened 2026-08-08
Judge the reconciliation grace on fatal diagnostics only (#398)

### LINKED ISSUE #398: A warning-severity diagnostic silently collapses the artifact-reconciliation grace window
Found by an adversarial review of #397 (since closed). Confirmed by reading the code, not yet observed in a run — filing before it costs one.

## The defect

`onlyTransientArtifactDiagnostics` in `packages/runtime/src/workflow-sync.ts` decides whether a node's gate failure is the benign "artifacts have not landed on the durable volume yet" case:

```ts
return diagnostics.length > 0 && diagnostics.every((diagnostic) => transientCodes.has(diagnostic.code));
```

`every`, over the **whole** diagnostic list, against a four-code allowlist (`REQUIRED_ARTIFACT_MISSING`, `REQUIRED_ARTIFACT_EMPTY`, `GENERATED_TEST_FILE_MISSING`, `GENERATED_TEST_FILE_EMPTY`). **There is no severity filter.** Any diagnostic outside that allowlist — including a `severity: "warning"`, which by construction cannot fail a node — makes this `false`, so:

```ts
const gracePending = !gate.ok && reconciliationError === undefined &&
  onlyTransientArtifactDiagnostics(gate.diagnostics) && !artifactReconciliationGraceExpired(grace, input.nowMs);
```

collapses. The bounded reconciliation grace (`ARTIFACT_RECONCILIATION_GRACE_MS` / `ARTIFACT_RECONCILIATION_MAX_ATTEMPTS`) is skipped entirely and the node fails immediately with `category: "artifact-contract"`, consuming a retry — instead of re-checking for five minutes, which is exactly the stranding the grace was built to prevent.

## It is reachable today, on property-lens nodes

`artifact-gates.ts` has three warning-severity sites, and all three are lens-gated:

- `ARTIFACT_VERIFICATION_DIGEST_REFRESHED` — inside `verifyLensReferenceExpectationAuthority`
- `PROPERTY_REFERENCE_EXPECTATION_SANITIZED`
- `PROPERTY_REFERENCE_EXPECTATION_CATALOG_ABSENT`

So a `property-specification-*` node that emits any of those **while a required output is still landing** loses its grace and hard-fails on the first sync. Campaign and final-report nodes are safe only because no warning is currently reachable on them — which means this is also a tripwire for anyone adding a warning anywhere else. That is how it was found: a proposed warning on the campaign node would have silently disabled the grace on a node with eight required outputs.

## Fix

Judge the grace on fatal diagnostics only:

```ts
const errors = diagnostics.filter((diagnostic) => diagnostic.severity === "error");
return errors.length > 0 && errors.every((diagnostic) => transientCodes.has(diagnostic.code));
```

`gate.ok === false` already implies at least one error, so this is equivalent for every case except the one that is currently wrong. It also makes "a warning cannot affect a node" true in general rather than true-by-accident.

Regression test to add: a node whose gate emits one transient `REQUIRED_ARTIFACT_MISSING` **plus** one warning must still enter the grace window, and must not be marked failed on the first sync.

Related, cosmetic and pre-existing: `missingCount: gate.diagnostics.length` at the grace call site counts all diagnostics, not missing artifacts, and is persisted into `artifact_reconciliation_grace.missing_count` and the `node-artifact-reconciliation-grace-*` events. It does not affect expiry, which uses `deadline_at`/`attempts`.

aviggiano:
Fixes #398.

## The defect

`onlyTransientArtifactDiagnostics` decides whether a gate failure is the benign *"the durable volume has not caught up yet"* case, and therefore whether the node waits out the bounded five-minute reconciliation grace or fails immediately:

```ts
return diagnostics.length > 0 && diagnostics.every((diagnostic) => transientCodes.has(diagnostic.code));
```

`every`, over the **whole** list, against a four-code allowlist, with **no severity filter**. Any diagnostic outside that allowlist collapses the grace — including a `severity: "warning"`, which by construction cannot fail a node.

The result is exactly the stranding the grace was built to prevent: a required output still landing on the volume, plus one warning, and the node fails on the first sync with `category: "artifact-contract"` and a consumed retry, instead of re-checking for five minutes.

## It is reachable today

The three warning-severity sites in `artifact-gates.ts` are all lens-gated (`ARTIFACT_VERIFICATION_DIGEST_REFRESHED` sits inside `verifyLensReferenceExpectationAuthority`; the other two are the reference-expectation warnings). So a `property-specification-*` node that sanitizes a lens reference expectation while an output is still in flight loses its grace.

Campaign and `final-report` nodes are safe only because no warning is currently reachable on them — which makes this a **tripwire for anyone adding one**. That is how it was found: a warning proposed on the campaign node (PR #397, since closed) would have silently disabled the grace on a node with eight required outputs.

## The fix

```ts
const fatal = diagnostics.filter((diagnostic) => diagnostic.severity === "error");
return fatal.length > 0 && fatal.every((diagnostic) => transientCodes.has(diagnostic.code));
```

`gate.ok` is false only when at least one error exists, so filtering to errors is equivalent in every case except the one that was wrong. It also makes "a warning cannot affect a node" true in general rather than true by accident.

## Tests

Five assertions, red before the fix:

- a lone transient miss keeps the grace (unchanged behaviour)
- **a warning alongside a transient miss must not collapse the grace** — the regression
- a genuine contract error still skips the grace (non-vacuity)
- warnings alone are not a transient miss
- an empty list is not a transient miss

The predicate is exported so the severity rule is tested where it lives; the wiring into `gracePending` is already covered by the grace integration tests in `runtime.test.ts`, which still pass.

`@ultrafuzz/runtime`: 436 passed, 0 failed, 15 skipped. eslint and prettier clean.

## Note

Pre-existing and deliberately not changed here: `missingCount: gate.diagnostics.length` at the grace call site counts all diagnostics rather than missing artifacts, and is persisted into `artifact_reconciliation_grace.missing_count` and the `node-artifact-reconciliation-grace-*` events. It does not affect expiry, which uses `deadline_at`/`attempts`. Recorded on #398.

<h3>Greptile Summary</h3>

The PR changes artifact-reconciliation grace classification to consider only error-severity diagnostics and preserves non-fatal gate diagnostics when returning early during the grace period.
- Moves the transient-diagnostic predicate into `artifact-reconciliation.ts` for direct testing without widening the package export surface.
- Prevents warning and informational diagnostics from collapsing a transient artifact grace period.
- Adds focused tests covering transient errors, warnings, informational diagnostics, genuine contract errors, and empty input.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| packages/runtime/src/artifact-reconciliation.ts | Adds the severity-aware transient-artifact predicate and keeps it outside the package's public export surface. |
| packages/runtime/src/workflow-sync.ts | Uses the relocated predicate for grace decisions and preserves non-error gate diagnostics on the early-return path. |
| packages/runtime/test/artifact-reconciliation.test.ts | Adds direct coverage for mixed diagnostic severities and non-transient contract failures. |

</details>

```mermaid
flowchart TD
  A[Verify required artifacts] --> B{Gate contains errors?}
  B -- No --> C[Continue successful finalization]
  B -- Yes --> D[Filter to error diagnostics]
  D --> E{All errors are transient artifact misses?}
  E -- No --> F[Fail artifact contract immediately]
  E -- Yes --> G{Grace still active?}
  G -- Yes --> H[Return running and preserve non-fatal diagnostics]
  G -- No --> F
```
</details>

<sub>Reviews (2): Last reviewed commit: ["Keep the predicate private and stop swal..."](https://github.com/monad-developers/ultrafuzz/commit/395c9ebc7796485fa02ced462ac0ca22673fd7f0) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=51411997)</sub>

**Context used:**


aviggiano:
Two independent reviews landed, one asked to argue against merging. It failed to break the premise and strengthened it; both found real defects, fixed in `395c9eb`.

**Reachability is worse than I filed, and I verified it.** No node in `topology.yml` declares a reference-expectations output, and the comment on `PROPERTY_REFERENCE_EXPECTATION_CATALOG_ABSENT` says a run without that catalogue is *"the normal case, not the exceptional one"*. That warning therefore fires on **every** gate evaluation of all nine `property-specification-*` nodes. So on `main` the grace is not occasionally disabled — it is already inert across the entire property fan-out on every catalogue-less run, and only the volume lag is left to chance. The adversarial reviewer's kill shot was that the warning path might require reading the artifact that would be missing; it does not — the gate's output loop pushes `REQUIRED_ARTIFACT_MISSING` and keeps going, and the warnings gate on the lens JSON alone, while a lens node has three outputs and reconciliation is per file.

**The export was wrong, and the repo already says so.** `index.ts` does `export * from "./workflow-sync.js"`, so "exported for tests" made the predicate public API. `git-capture-diagnostics.ts` exists precisely to prevent this and states that an `@internal` tag *"is only a comment — it does not stop `export *`"* — which is exactly the rationalisation I had written into the doc comment. Moved to `artifact-reconciliation.ts`, which the index does not re-export. Verified: package API unchanged.

**I introduced a regression, now fixed.** The grace path returns *before* `diagnostics.push(...gate.diagnostics)`, so my change made a sync silently swallow the gate's warnings. That matters because two of them fire only on the pass that actually mutates — a re-sealed verification marker and a sanitized lens output — and node patches persist no diagnostics. The sync that held the grace would have been the only one able to report a security-relevant edit, and it would have reported nothing. Non-error gate diagnostics now flow into the grace return.

**Corrections to my PR body**, both fair:
- "Five assertions, red before the fix" was overstated. Only the warning-alongside-transient assertion is red on `main`; the other four pass unchanged. The *file* is red before the fix, which is what the TDD claim rests on, but the wording overclaimed.
- "changes nothing except that case" was not accurate. Where an artifact is genuinely absent rather than lagging, a lens node now waits out the bounded grace before reaching the identical failure. That is the trade the grace makes everywhere else, but it is a behaviour delta, not equivalence.

Two assertions added for the only direction in which the new predicate is stricter: a transient *code* at warning or info severity is not a gate failure to wait out.

**Not done, and I want to be explicit rather than quiet about it:** the reviewer's preferred fix was an integration test through `syncRun` with a lens-output topology variant, which would pin reachability on a lens node instead of testing the predicate directly. That is the better test and I have not written it — the existing grace integration tests only drive `project-discovery`, which emits no warnings, so nothing end-to-end covers the changed branch. Recorded on #398.

`@ultrafuzz/runtime`: 436 passed, 0 failed, 15 skipped. eslint and prettier clean.

PR #409, opened 2026-08-08
Keep a description that ends with a pipe (#408)

### LINKED ISSUE #408: Markdown parity strips a trailing pipe, so any property quoting a docs table row fails fan-in
Confirmed in production in R58 (`aave-v4-v0012-main-issue400fix-r58-invariant-only`, image `im-01KZGKVS0FST23B6TCX1AZ76VJ`, source `d422fe6`). `property-specification-fanin` failed at 13:37:19Z with five diagnostics:

```
Properties Markdown must preserve canonical property "property-277" with its description,
category, priority, and sources (missing "description")
... identically for property-278, 279, 280, 281
```

**The artifacts are correct.** `properties.md` is 504 KB, complete through `property-388`, with 388 start markers and 388 end markers perfectly paired, each property appearing exactly once. For `property-277` the Markdown `description:` value and the `properties.json` description are **byte-identical** — 654 characters, `equal: True`.

## Root cause

`markdownFieldValues` captures the field value and then strips a trailing pipe:

```ts
const parts = [match[1].replace(/\s*\|\s*$/u, "")];
```

That strip exists for Markdown table cells (`| description | value |`). But when a description legitimately **ends** with `|`, the parser deletes it, the extracted value no longer equals the JSON value, and `markdownFieldEqualsValue` returns false — reported as the field being *missing*.

The correlation is exact:

| property | description ends with `|` | gate verdict |
|---|---|---|
| 276 | no  | pass |
| 277 | yes | **fail** |
| 278 | yes | **fail** |
| 279 | yes | **fail** |
| 280 | yes | **fail** |
| 281 | yes | **fail** |
| 282 | no  | pass |

Seven consecutive properties, and the split is perfectly explained by that one character.

## Why this is not the producer's fault

`property-specification-fanin` is required to preserve verbatim source text. Aave v4's `docs/overview.md` documents parameters in Markdown tables, so quoting a row verbatim yields a description that ends in `|`:

```
description: Project discovery evidence evidence-doc-liquidation-targethf-row (bound) at
docs/overview.md:line 237; ... verbatim: | `TargetHealthFactor` | A spoke-wide value set by
the Governor ... | Must be ≥ the `HEALTH_FACTOR_LIQUIDATION_THRESHOLD` constant.      |
```

Any target whose docs use tables will hit this. It is not specific to Aave.

## Reproduction

Replaying R58's real `properties.json`, `properties.md` and the project-discovery `invariant-evidence-ledger.json` through `verifyRequiredArtifactsForAttempt` on `d422fe6` reproduces all five diagnostics exactly.

## Suggested fix

The leading `|` in the field pattern already identifies a table row:

```ts
new RegExp(`^\\s*(?:\\|\\s*)?(?:[-*+]\\s*)?${escapeRegExp(field)}\\s*(?::|\\|)\\s*(.*)$`, "iu")
```

Strip the trailing pipe **only when the matched line actually opened with one**, so `| description | value |` still normalises while `description: ...ends with a pipe |` is preserved. A regression test should use a description that terminates in `|`, which is the shape R58 produced.

## Impact

R58 continued past it — downstream consumes `properties.json`, which is valid — but the node is `failed`, and on the evidence of R56 a failed node can strand the tail. Nine property lenses feed this node on every run, and any of them quoting a docs table row will trip it.

aviggiano:
Fixes #408.

## The defect

`markdownFieldValues` strips a trailing pipe from every captured value to normalise Markdown table cells:

```ts
const parts = [match[1].replace(/\s*\|\s*$/u, "")];
```

A value that legitimately **ends** with `|` is truncated by that strip, so it no longer equals its JSON counterpart, and parity reports the field as **missing** on an artifact that is byte-identical to the catalog.

## Confirmed in production

R58 (`aave-v4-v0012-main-issue400fix-r58-invariant-only`, image `im-01KZGKVS0FST23B6TCX1AZ76VJ`, source `d422fe6`) failed `property-specification-fanin` at 13:37:19Z with five diagnostics, on five consecutive properties. The correlation is exact:

| property | description ends with `|` | verdict |
|---|---|---|
| 276 | no  | pass |
| 277–281 | yes | **fail** |
| 282 | no  | pass |

The artifact was fine: `properties.md` is 504 KB, 388 start markers and 388 end markers perfectly paired, each property exactly once, and `property-277`'s rendered description is **byte-identical** to the JSON at 654 characters.

## Why the producer is not at fault

`property-specification-fanin` is required to preserve verbatim source text. Aave v4's `docs/overview.md` documents parameters in Markdown tables, so quoting a row verbatim yields a description ending in `|`:

```
description: ... verbatim: | `TargetHealthFactor` | A spoke-wide value set by the
Governor ... | Must be ≥ the `HEALTH_FACTOR_LIQUIDATION_THRESHOLD` constant.      |
```

Any target that documents parameters in tables hits this. It is not specific to Aave.

## The fix

The leading pipe is what makes a line a table row, so capture it and strip the trailing one only when the line actually opened with it. `| description | value |` still normalises; `description: … |` survives.

## Verification

- New test using R58's shape — prose followed by a verbatim docs table row — is **red before the fix**.
- **Replayed R58's real artifacts** (`properties.json`, the 504 KB `properties.md`, and the project-discovery `invariant-evidence-ledger.json`) through `verifyRequiredArtifactsForAttempt`: `ok: true`, **0 diagnostics**, where `d422fe6` returns the five production diagnostics.
- `@ultrafuzz/runtime`: 437 passed, 0 failed, 15 skipped. eslint and prettier clean.

## Impact

R58 continued past it because downstream consumes `properties.json`, which is valid — but the node is `failed`, and R56 showed a failed node can strand the entire tail. Nine property lenses feed this node on every run.

<h3>Greptile Summary</h3>

The PR fixes Markdown/JSON artifact parity when a prose field legitimately ends with a pipe.
- Captures whether a field line opened as a Markdown table row.
- Removes a trailing pipe only from actual table rows.
- Adds a production-shaped regression test covering a pipe-terminated property description.

The PR appears safe to merge with no actionable defects identified.

The parser preserves legitimate trailing pipes on prose fields while retaining trailing-cell-delimiter normalization for lines that begin as Markdown table rows, and the regression test exercises the resulting artifact gate.

| Filename | Overview |
|----------|----------|
| packages/runtime/src/artifact-gates.ts | Correctly distinguishes table-row delimiters from trailing pipes that are part of prose field values. |
| packages/runtime/test/artifact-gates.test.ts | Adds an end-to-end artifact parity regression test using the affected property-description shape. |

</details>

<sub>Reviews (1): Last reviewed commit: ["Keep a description that ends with a pipe..."](https://github.com/monad-developers/ultrafuzz/commit/ff81a290d7fe35340a65640e9eefa34ba8fad657) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=51440655)</sub>


aviggiano:
I set out to attack this PR's premise myself, on the strongest counter-argument available: the repo contains an existing parity test that renders a pipe-bearing description as `\|` in Markdown while the JSON holds a bare `|`. If escaping were the documented contract, then a raw trailing pipe would be the producer's error, the gate would be right, and this PR would be papering over it.

**It is not the contract, and escaping is not a workaround either.** `.ultrafuzz/prompts/properties/property-specification-fanin.md` mentions "escape" or "backslash" exactly **zero** times. And measured against `main` (`d422fe6`), a description ending in `|` has **no valid rendering at all**:

```
--- MAIN (pre-fix) ---
  verbatim (bare trailing pipe)      parity-failures=1  ok=false
  fully escaped pipes                parity-failures=1  ok=false
  only trailing pipe escaped         parity-failures=1  ok=false
```

Every way a producer could render it is rejected. The escaped forms fail for the same reason as the bare one: the unconditional `\s*\|\s*$` strip eats the final pipe, leaving a dangling backslash that cannot match the JSON. There was no correct answer to give.

With the fix, all three are accepted:

```
--- WITH FIX ---
  verbatim (bare trailing pipe)      parity-failures=0  ok=true
  fully escaped pipes                parity-failures=0  ok=true
  only trailing pipe escaped         parity-failures=0  ok=true
```

That is the right outcome rather than an over-broad one: `normalizeMarkdownFieldValue` already unescapes, so escaped and verbatim renderings are supposed to be equivalent — the trailing-strip was the single thing breaking all of them.

For completeness on the existing test I was worried about: its `PROPERTY_MARKDOWN_PARITY_MISSING` assertion fires because `property-2` is absent from that Markdown entirely, not because of the pipe. Its `property-1`, whose description carries a **mid-string** pipe rendered as `\|`, matches fine both before and after. The full runtime suite is green at 437 passed / 0 failed.

Both review agents I dispatched were terminated by a session teardown before reporting, so this is my own adversarial pass rather than an independent one — flagging that plainly rather than implying a review happened. The reasoning above is reproducible from the probe: three renderings, two builds, six measurements.

Issue #405, opened 2026-08-08
What the generated Aave v4 invariant suite actually tests, and what it cannot


aviggiano:
Filed so this is not re-derived later, and so nobody reads the first completed Aave v4 run as a verdict on invariant testing that it cannot support.

The open question behind this benchmark is whether ultrafuzz with invariant testing finds bugs that the no-fuzz release configuration cannot. The strong form of that case is multi-transaction bugs: violations only reachable through a sequence of operations, which no amount of reading a single function surfaces. This note records what the pipeline actually generated when measured against that bar.

## Evidence

The R52 suite from `stateful-invariant-implement-properties` — 13 properties, 33 KB of `Properties.sol` — classified by what it takes to violate each one:

| class | properties | needs a sequence? |
|---|---|---|
| State-local conservation — sums of parts equal the whole | `catalogHubAssetAccounting`, `catalogHubSpokeAggregates`, `catalogSpokeReserveAndUserAccounting`, `catalogUserAccountData` | No — checkable in any single state |
| Math identity against a re-derived oracle | `catalogPreviewRounding`, `catalogRoundTripPreviews`, `catalogInterestRateMath`, `catalogLiquidationBonusMath` | No — single call |
| Config and sanity | `catalogConfigurationBounds`, `catalogOracleSources`, `reconReservesStayListed`, `reconSetupStable` | No |
| Temporal | `catalogAddedSharePriceNonDecreasing` | Yes, with caveats below |

Twelve of thirteen are single-state or single-call. Those are structural facts about the code, which is the class a no-fuzz agent reading the contracts can derive directly. Fuzzing confirms them; it is not what discovers them.

The math-identity group is weaker still. `catalogPreviewRounding` asserts `hub.previewAddByAssets(...) == _mulDivDown(assets, shares + VIRTUAL_SHARES, ...)` — a formula re-derived from the implementation under test. That catches divergence between two code paths; it cannot catch a wrong shared assumption.

## The one temporal property is weaker than the reference's

```solidity
function property_catalogAddedSharePriceNonDecreasing() public view {
    if (!_before.initialized || !_after.initialized) return;
    if (_before.activeAddedShares == 0 || _after.activeAddedShares == 0) return;
    assert(_after.activeAddedAssets * _before.activeAddedShares
        >= _before.activeAddedAssets * _after.activeAddedShares);
}
```

[...]

PR #506, opened 2026-08-11
Withhold benchmark reference paths from pinned targets

### LINKED ISSUE #503: Aave v4 benchmark target ships the reference invariant suite, so the invariants group copies it instead of deriving properties
## Problem

The pinned Aave v4 benchmark target ships the benchmark author's **reference invariant suite** at `tests/recon/`. The invariants group reads it during `project-discovery` and carries it forward verbatim, so the R61 Aave v4 campaign cannot answer the question it exists to answer: whether Ultrafuzz independently derives an invariant campaign and properties comparable to `scfuzzbench/aave-v4-scfuzzbench`.

This is an input-contamination defect in benchmark target materialization. It is not a prompt-quality problem, and it must not be fixed by changing prompts.

## How this happened

#405 measured the generated suite at R52 against the reference and recorded two conclusions:

> There is **no analogue of `invariant_shouldNotBecomeLiquidatable`** — "no operation other than a price change may make a user liquidatable". That is the strongest example of the thesis: violable only by a sequence, encoding economic intent rather than code structure. The generated `BeforeAfter` cannot express it — no operation tracking, no liquidatable flag.

and, in "To actually answer the question":

> 1. The **`scfuzzbench` fork** as target, so ground truth exists.

R52 ran against upstream `aave/aave-v4`, which has no `tests/recon/`, so its generated suite was genuinely independent. Adopting the `scfuzzbench` fork supplied the ground truth needed for scoring — and, in the same step, placed the reference solution inside the agent's working tree. The two changes were not separable, and the contamination was not noticed.

## Evidence

Target `scfuzzbench/aave-v4-scfuzzbench@edd6c82721512540c8c90e7a36a4a8e19fd7bdf3`, tree `3b3910d0657e4a639888ccae45822af680de6ef6`, ships 12 tracked files under `tests/recon/`:

```
tests/recon/BeforeAfter.sol                3150 bytes
tests/recon/CryticTester.sol                438 bytes
tests/recon/CryticToFoundry.sol             644 bytes
tests/recon/Properties.sol                13722 bytes
tests/recon/Setup.sol                      2534 bytes
tests/recon/TargetFunctions.sol             968 bytes
tests/recon/targets/AdminTargets.sol        573 bytes
tests/recon/targets/DoomsdayTargets.sol     702 bytes
tests/recon/targets/IAaveOracleTargets.sol 2293 bytes
tests/recon/targets/IHubTargets.sol        5463 bytes
tests/recon/targets/ISpokeTargets.sol     13371 bytes
tests/recon/targets/ManagersTargets.sol    2594 bytes
```

`tests/recon/Properties.sol` is the pre-vetted invariant list from Sherlock contest 1209, fully implemented: 9 `invariant_*` bodies and 18 named invariant constants. `tests/recon/BeforeAfter.sol` carries exactly the ghost state #405 said the pipeline could not produce — `ghost_maxSupplySharePriceAssetsSeen`, `ghost_maxDrawnIndexSeen`, an `Operation` enum, and `isAnyUserLiquidatable`.

The target repository's own README documents the upstream layout with `tests/invariant/`; `tests/recon/` is the benchmark author's reference solution layered on top, not part of the protocol under test.

### The invariants group consumes it

In the r5 run (`0f5117fb`, 32/32 succeeded), `project-discovery` recorded `tests/recon/Properties.sol` in both its report and `setup/invariant-evidence-ledger.json`.

The suite produced by `stateful-invariant-campaign` contains 14 `invariant_*` functions. **All 9 reference functions are retained by name**, and only 5 are new:

| retained verbatim from the target | added by the run |
|---|---|
| `invariant_canary` | `invariant_hubAssetDebtIndexAndFeeFormulas` |
| `invariant_hubAddedAssetsGreaterThanSpokeSum` | `invariant_hubPremiumAndDeficitSumsMatchSpokes` |
| `invariant_hubAddedSharesMatchesSpokeSum` | `invariant_spokeConfigurationAndOracleBounds` |
| `invariant_shouldNotBecomeLiquidatable` | `invariant_spokeReserveAccountingMatchesHub` |
| `invariant_supplySharePriceAndDrawnIndexMonotonic` | `invariant_userDebtAndAccountDataMatchFormulas` |
| `invariant_totalBorrowedLessThanSupplied_v0` | |
| `invariant_totalBorrowedLessThanSupplied_v1` | |
| `invariant_totalBorrowedLessThanSupplied_v2` | |
| `invariant_totalBorrowedSharesMatchesSpokeSum` | |

`invariant_shouldNotBecomeLiquidatable` is **byte-identical** to the target's copy, down to the `INVARIANT_HOLY_GRAIL_SHOULD_NOT_BECOME_LIQUIDATABLE` constant name:

```solidity
function invariant_shouldNotBecomeLiquidatable() public returns (bool) {
    int256 maxViolation = type(int256).min;
    if (_after.operation != Operation.SetPrice) {
        int256 diff = (_before.isAnyUserLiquidatable || !_after.isAnyUserLiquidatable) ? int256(0) : PERCENT;
        if (diff > maxViolation) {
            maxViolation = diff;
        }
    }
    if (!OPTIMIZATION_MODE) {
        t(maxViolation <= 0, INVARIANT_HOLY_GRAIL_SHOULD_NOT_BECOME_LIQUIDATABLE);
    }
    return maxViolation <= 0;
}
```

The five reference formulations #405 called out as absent — the deliberate v0/v1/v2 solvency variants and the liquidatable "holy grail" property — are all present in r5 solely because they were copied.

## Consequence

Any comparison of a post-fork Aave v4 run against the reference is circular. In particular:

- The r5 headline of 84 implemented properties overstates independent capability by an unknown margin.
- The apparent improvement from R52 to r5 on exactly the properties #405 identified as missing is an artifact of contamination, not a capability gain.
- Every Aave v4 run since the fork adoption is affected, including any run currently in flight.

Scoring against `findings.yml` is unaffected — ground truth is bug-level and already held out. Only property generation is contaminated.

## Required remediation

Hold reference-solution paths out of the materialized target the same way ground truth is already held out, and make the exclusion provable.

1. Let a benchmark target declare held-out reference paths (for this target, `tests/recon/**`) somewhere outside the target tree, next to the existing private ground truth rather than inside the repository under test.
2. Apply the hold-out during pinned target materialization, before any node runs.
3. Record the hold-out in the pinned source proof — the excluded paths, their digests, and the resulting tree — so a reviewer can prove which bytes the run could not see. Silent deletion is not acceptable; the proof must continue to bind the upstream commit and tree.
4. Keep the existing byte/tree/gitlink verification working, and fail closed if a held-out path is present in the materialized target when a run starts.
5. Do not change any prompt to compensate. If the invariants group cannot rebuild this harness from the protocol source, that is the measurement.
6. Preserve behavior for targets that declare no hold-out, and for ordinary non-benchmark projects.

An agent re-creating a similar file inside its own workspace is a legitimate result and must stay allowed; only the materialized input is constrained.

## Acceptance criteria

- A benchmark target can declare held-out paths, and Aave v4 declares `tests/recon/**`.
- Materialization removes them, and the source proof records the exclusion with digests while still binding the upstream commit/tree.
- A run against a target whose held-out paths are still present fails closed with a clear diagnostic.
- Regression coverage for: hold-out applied, hold-out declared but absent, no hold-out declared, and a glob that matches nothing.
- A fresh Aave v4 invariant-only run on the fixed materialization produces a suite whose overlap with `tests/recon/Properties.sol` is reported, so independent capability is measurable.
- Existing runtime/Modal/eval suites, builds, typechecks, formatting, lint, docs, and `git diff --check` pass.

Refs #405, #404

aviggiano:
## Summary

Fixes #503.

The Aave v4 benchmark target ships the benchmark author's reference invariant suite at `tests/recon/`. `project-discovery` reads it, and the campaign carries it forward: in run r5 the generated suite retained **all nine** reference `invariant_*` functions by name and added only five of its own. `invariant_shouldNotBecomeLiquidatable` was byte-identical to the target's copy, down to the `INVARIANT_HOLY_GRAIL_SHOULD_NOT_BECOME_LIQUIDATABLE` constant — the exact property #405 recorded as one the pipeline could not generate.

So the suite could not measure what the invariants group derives on its own. This is an input-contamination defect in target materialization, not a prompt-quality problem, and it is deliberately **not** fixed by touching prompts.

## What changed

A target may declare paths the run must never read:

```yaml
targets:
  - id: target
    repo: "https://github.com/scfuzzbench/aave-v4-scfuzzbench"
    ref: "edd6c82721512540c8c90e7a36a4a8e19fd7bdf3"
    ground_truth: aave-v4-scfuzzbench/findings.yml
    held_out_paths: ["tests/recon"]
```

- `materializePinnedSource` removes them and commits the removal onto the pinned branch. It has to be a commit, not a dirty worktree: every node workspace is a Git worktree of that branch and would otherwise restore the withheld files.
- The hold-out is applied immediately after checkout, before submodule hydration, because the pinned submodule manifest is stored per commit and must key off the final revision.
- The commit uses a fixed identity and date, so the same inputs always produce the same hold-out commit.
- The proof gains `held_out`. `commit`/`tree` are the hold-out revision; `source_commit`/`source_tree` bind the benchmark commit it descends from, and `entries` record every withheld file with its upstream blob and size.

Verification is strict:

- HEAD's only parent must be the benchmark commit, with the expected message.
- The diff against that parent must be deletions only, matching the recorded entries exactly — a hold-out revision that removed anything undeclared is rejected.
- Each recorded entry's blob and size are re-derived from the benchmark commit, so an under- or over-declared record fails.
- A record claiming descent from another commit fails.
- A declaration matching no tracked path fails closed rather than silently running without the hold-out.
- Paths that are absolute, traverse, or name `.git` are rejected.
- `assertHeldOutPathsAbsent` refuses to launch a row whose held-out paths are still on disk, so a hand-prepared or stale checkout cannot reintroduce the answer key.

Callers that only know the benchmark commit — including `worker.ts` — keep passing the upstream ref and need no change; `inspectPinnedSource` reconciles the hold-out itself. Targets without a declaration are byte-for-byte unchanged (`held_out: null`, `revision_count: 1`).

An agent that writes a similar file inside its own workspace stays allowed; only the materialized input is constrained.

## Verification

Against the real Aave v4 target with `held_out_paths: ["tests/recon"]`:

```
commit (holdout): 111469459f85ed3cf8058b2524cac91b573b49c8
tree   (holdout): 5e9c604d9474da946403369f652425880babee78
revision_count: 2   commit_object_count: 2
held_out.source_commit: edd6c82721512540c8c90e7a36a4a8e19fd7bdf3
held_out.source_tree:   3b3910d0657e4a639888ccae45822af680de6ef6
held_out entries: 12 files, 46452 bytes
submodules gitlinks: 9
re-inspect against the upstream ref: ok
```

`tests/recon` is absent; `src/` (100 files), `tests/unit`, `tests/gas`, `tests/misc`, `tests/mocks`, and all nine hydrated submodule gitlinks are intact; `git status` is clean. The hold-out commit hash reproduced exactly across independent runs.

Six new tests in `packages/modal/test/pinned-source.test.ts` cover: withholding and binding to the benchmark commit; no declaration leaving the target unchanged; a declaration matching nothing; declarations that escape the checkout; a hold-out revision that removed more than it declared; and three tampered-record shapes.

- `pnpm --filter @ultrafuzz/modal exec vitest run test/pinned-source.test.ts` — 13 passed
- `pnpm -w format:check`, `pnpm -w lint`, `pnpm -w build`
- `git diff --check` clean

## Follow-up

This changes what the Aave v4 campaign measures. The next invariant-only run should report the overlap between the generated suite and `tests/recon/Properties.sol` so independent capability is quantified rather than assumed.

<h3>Greptile Summary</h3>

The PR prevents benchmark reference suites from contaminating evaluation inputs by rewriting held-out targets to isolated parentless revisions.
- Propagates hold-out declarations through private Modal configuration, workspace materialization, and generated eval suites.
- Removes declared tracked paths, prunes recoverable source objects, and records/verifies hold-out provenance.
- Adds launch-time absence checks and path validation for local eval targets.
- Preserves upstream source identity when binding private ground truth.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| packages/modal/src/pinned-source.ts | Rewrites held-out targets to parentless revisions, prunes recoverable reference objects, and verifies hold-out provenance and isolation. |
| packages/modal/src/worker.ts | Propagates private target hold-outs into materialization and binds ground truth to the original benchmark commit. |
| packages/modal/src/config.ts | Adds validated private benchmark hold-out declarations. |
| packages/modal/src/private-suite.ts | Carries configured hold-out paths into generated private eval suites. |
| packages/evals/src/suite.ts | Validates held-out paths as contained relative target paths. |
| packages/evals/src/runner.ts | Refuses benchmark launch when a declared held-out path remains present. |
| packages/modal/test/pinned-source.test.ts | Covers parentless isolation, object pruning, unchanged undeclared targets, unsafe declarations, and record tampering. |
| packages/evals/test/suite.test.ts | Covers suite propagation, launch-time absence enforcement, and unsafe path rejection. |

</details>

```mermaid
flowchart LR
    C[Private benchmark config] --> M[Materialize pinned source]
    M --> H[Create parentless hold-out revision]
    H --> P[Prune source commit and withheld blobs]
    P --> V[Verify isolated source proof]
    V --> S[Generate eval suite with held_out_paths]
    S --> A[Assert paths absent before launch]
    A --> R[Run benchmark]
```
</details>

<sub>Reviews (5): Last reviewed commit: ["fix(evals): validate held-out declaratio..."](https://github.com/monad-developers/ultrafuzz/commit/307a7b1a05b55a9a0552a47de913053a31af3051) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=52050817)</sub>

- Knowledge Base — [Security](https://app.greptile.com/monad-foudnation/-/custom-context/knowledge-base/monad-developers/ultrafuzz/-/docs/security.md)
</details>


aviggiano:
Both review findings were correct and are fixed in c5055069.

**Parent commit exposes held-out files (P1, security).** Confirmed on the staged Aave v4 target before the fix — `git show HEAD^:tests/recon/Properties.sol` returned all 13,722 bytes. The hold-out was cosmetic.

The checkout is now rewritten to a **parentless** revision that never contained the declared paths, followed by `reflog expire --expire-unreachable=now` and `gc --prune=now`. Materialization then asserts the benchmark commit and every withheld blob are gone before it returns.

After the fix, on the real target:

```
git show HEAD^:tests/recon/Properties.sol   -> fatal: invalid object name 'HEAD^'
git cat-file -e edd6c82721512540c8c90e7a36a4a8e19fd7bdf3 -> missing
revisions: 1 | reflog: 0 | unreachable: 0
```

An exhaustive scan of every blob in the object store finds no trace of the reference suite. Because the benchmark commit is destroyed, verification can no longer diff against a parent, so it proves absence instead: HEAD is the recorded parentless revision, no declared path is tracked, no recorded blob is readable, the declaration matches nothing, and the benchmark commit is unreadable. This also restores the original strict invariants — `revision_count` and `commit_object_count` are back to `1`.

**Modal drops hold-out declarations (P1).** Also correct. `held_out_paths` is now part of the private benchmark target config, rendered into the generated suite, and passed to `materializePinnedSource`. While wiring it I found a related bug: ground truth asserts its subject revision equals the target commit, which would now be the hold-out revision, so it binds `held_out.source_commit` instead.

Verification: modal 414/414, evals 220/220, lint, format, docs, and workspace build all pass, with global and system Git configuration suppressed so the committer identity is proven to come from the hold-out itself.

aviggiano:
**Hold-out paths bypass normalization (P2).** Valid, fixed in 307a7b1a.

`held_out_paths` entries now go through the same absolute-path, traversal, and Git-metadata rules that source materialization applies, enforced in the suite schema so a bad declaration fails at load. The launch guard repeats the check independently, because that is where the join to the checkout actually happens — `../shared/reference`, `/etc`, `tests/../../etc`, and `.git` are all rejected there rather than silently probing an unrelated directory.

Covered by two new cases: a suite that declares an escaping path fails to load, and the guard rejects each escaping shape.

evals 221/221, modal 414/414, lint and format clean.

PR #512, opened 2026-08-11
Accept a held-out target whose benchmark commit was destroyed

### LINKED ISSUE #511: Launching a held-out benchmark fails because its declared commit is intentionally destroyed
## Problem

A benchmark target that declares `held_out_paths` cannot be launched. `eval run` fails immediately with:

```
target ref edd6c82721512540c8c90e7a36a4a8e19fd7bdf3 is not present in /home/ubuntu/r61-local-targets/target-r8
```

`assertTargetRef` requires the declared `ref` to resolve inside the checkout. Withholding reference paths (#503, merged in #506) rewrites the checkout to a parentless revision and destroys the benchmark commit on purpose, so that requirement can never be satisfied. The two features are mutually exclusive as shipped.

## Reproduction

Materialize any target with `held_out_paths` and launch a suite that declares the same `ref`. The run fails before any node starts.

```
$ git -C <target> rev-parse HEAD
ac548733fd9c36cf9a32284a653a6429c089abfa
$ git -C <target> cat-file -e edd6c82721512540c8c90e7a36a4a8e19fd7bdf3
fatal: Not a valid object name
```

## Required remediation

Accept a checkout whose HEAD is the hold-out revision derived from the declared ref, using the target's own hold-out record as the authority, and keep rejecting everything else:

- a record naming a different benchmark commit
- a record naming a different HEAD
- an absent, unreadable, or malformed record
- targets with no hold-out at all, which must behave exactly as today

Refs #503

aviggiano:
Fixes #511.

A benchmark that declares `held_out_paths` could not be launched at all. `assertTargetRef` requires the declared `ref` to resolve inside the checkout, but the hold-out (#503, merged in #506) rewrites the target to a parentless revision and destroys that commit precisely so it cannot be read back. The two features were mutually exclusive as shipped:

```
target ref edd6c82721512540c8c90e7a36a4a8e19fd7bdf3 is not present in .../target-r8
```

This only surfaces end to end — the suites for #506 all passed, because none of them launch a run against a materialized hold-out.

## What changed

`assertTargetRef` consults the target's own hold-out record as a last resort and accepts the checkout only when that record states HEAD is exactly the revision derived from exactly the declared ref.

Still rejected: a record naming a different benchmark commit, a record naming a different HEAD, an absent/unreadable/malformed record, and everything rejected before for targets with no hold-out.

## Verification

New test covers the accepted case and each rejected shape, starting from a real Git checkout whose declared benchmark commit does not exist.

- evals: 222 passed
- `pnpm -w lint`, `pnpm -w format:check`, workspace build
- Confirmed against the real Aave v4 target: HEAD `ac548733`, declared ref `edd6c827` absent from the object store.

<h3>Greptile Summary</h3>

The PR allows eval launches to validate materialized held-out targets after the original benchmark commit has been removed.
- Falls back to the checkout’s pinned hold-out record when ordinary Git resolution fails.
- Requires the recorded materialized commit to equal HEAD and the recorded source commit to match the declared full or abbreviated hexadecimal ref.
- Adds end-to-end Git-checkout coverage for valid and rejected hold-out records.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| packages/evals/src/utils.ts | Adds a fail-closed hold-out metadata fallback and correctly supports abbreviated hexadecimal source refs. |
| packages/evals/test/utils.test.ts | Exercises acceptance of valid full and abbreviated refs plus rejection of mismatched, malformed, and non-SHA records. |

</details>

```mermaid
flowchart TD
  A[assertTargetRef] --> B{Declared ref resolves to HEAD?}
  B -->|Yes| C[Accept checkout]
  B -->|No| D{HEAD has declared SHA prefix?}
  D -->|Yes| C
  D -->|No| E[Read pinned hold-out record]
  E --> F{Recorded commit equals HEAD?}
  F -->|No| G[Reject target ref]
  F -->|Yes| H{Declared hex ref prefixes recorded source commit?}
  H -->|Yes| C
  H -->|No| G
```
</details>

<sub>Reviews (2): Last reviewed commit: ["fix(evals): accept an abbreviated pin fo..."](https://github.com/monad-developers/ultrafuzz/commit/2023bbf3829e55d6786fd81e2c056b7ef9e790b3) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=52072152)</sub>

**Context used:**


aviggiano:
**Abbreviated hold-out refs rejected.** Correct — the non-held-out path already accepts an abbreviated commit via `head.startsWith(ref)`, so requiring exact equality against the recorded full `source_commit` would reject a valid checkout whose suite pins a short SHA. Adopted the suggested prefix match, keeping the `^[0-9a-f]{7,40}$` guard so a branch name cannot prefix-match its way in.

Added coverage for an abbreviated pin being accepted, and for `main` and a non-matching 12-hex ref still being rejected. evals 222/222, lint and format clean.

Issue #523, opened 2026-08-12
Aave v4 postmortem: completed invariant campaign passed gates without executable benchmark coverage


aviggiano:
## Context

This is the postmortem for the first completed sealed Aave v4 invariant-only run after unblinding it against the ScFuzzBench Aave v4 benchmark.

- Ultrafuzz run: `aave-v4-v056-gpt55-xhigh-invariant-only-20260811-full-timeout`
- Target: [`aave/aave-v4@6959e321`](https://github.com/aave/aave-v4/commit/6959e3219b5506bf2acae18551cbb2a68a5b8fba) (`v0.5.6`)
- Profile: invariant-only, GPT-5.5 with xhigh reasoning
- Comparison: [`scfuzzbench/aave-v4-scfuzzbench`](https://github.com/scfuzzbench/aave-v4-scfuzzbench) and [run `gh-30059118265-1`](https://scfuzzbench.com/runs/gh-30059118265-1/a0a57a3e0b5b533094f20079e9a9b7ef/)

The workflow completed all 75 nodes. Recon 0.4.17 ran with 32 workers for 3,510.5 campaign seconds (3,601 process seconds), executed 11,183,800 calls, retained a 63-entry corpus, and reported zero failures. The final report contained zero protocol issues.

After excluding two deliberate canaries, the linked ScFuzzBench Recon run had 10 broken invariant/assertion identities. Ultrafuzz did not rediscover any of them. This is not mainly a timeout result: every one of the 10 identities had at least one ScFuzzBench first sighting within 2,580.35 seconds (about 43 minutes).

ScFuzzBench's own analysis warns that ground-truth mapping was not applied because its run commit differs from the evidence-pinned catalog revision. The 10 items below are therefore broken benchmark observations, not a claim of 10 independently validated production vulnerabilities.

### 1. The final "stateful" campaign had an effective sequence length of one

The generated configuration retained the smoke value:

```yaml
testMode: assertion
testLimit: 1
seqLen: 1
```

The long campaign overrode `testLimit`, but not `seqLen`. Its effective command used the maximum `uint64` test limit and a one-hour timeout while Recon confirmed `Sequence length: 1`.

Mutation handlers and global properties were separate public targets, and handler `_after()` asserted only stable inventory counts. All 63 retained corpus entries contained exactly one call. A test could therefore mutate state or evaluate a global property, but never evaluate that property after a fuzzed mutation.

[...]

Issue #522, opened 2026-08-12
Can invariant-only Ultrafuzz autonomously rediscover an Aave v4 benchmark bug family?


aviggiano:
## Research question

Can the sealed, stock Ultrafuzz invariant-only workflow—using GPT-5.5 with xhigh reasoning—autonomously rediscover at least one semantic bug family from the ScFuzzBench Aave v4 benchmark when given a genuinely stateful one-hour Recon campaign?

## Controlled target

- Repository: `aave/aave-v4`
- Release: `v0.5.6`
- Commit: `6959e3219b5506bf2acae18551cbb2a68a5b8fba`
- Mode: invariant-only
- Model for every model-authored workflow node: GPT-5.5, xhigh reasoning
- Prompts/topology/configuration: sealed stock inputs; no manual harness or property steering
- Campaign requirement: Recon runs for approximately 3,600 seconds with `--timeout 3600`, `--test-limit 18446744073709551615`, and a stateful sequence length of 100
- Evaluation: preserve benchmark blindness until generation, campaign, verification, dedupe, triage, severity classification, aggregation, and final reporting are complete; then compare semantic bug families rather than names or exact test shapes

## Why this remains unanswered

The first otherwise-successful controlled run completed all 75 workflow nodes and executed an approximately one-hour Recon campaign with 11,183,800 calls and no property-derived findings. However, it inherited `seqLen: 1` from the generated smoke configuration. Each action therefore reset to constructor state, making multi-transaction protocol states unreachable. That run is evidence about a one-step campaign, not a valid answer to the stateful research question.

The general validity defect is tracked in #520 and addressed by #521, which requires and verifies `--seq-len 100` for current stateful invariant campaigns.

## Completion criterion

After #521 is merged and the sealed experiment is rerun without manual steering:

1. Verify the immutable target, model/reasoning profile, invariant-only topology, and sealed inputs.
2. Verify the effective Recon command and typed artifacts prove the full timeout, unbounded test limit, and sequence length 100.
3. Verify the workflow and all artifact gates complete successfully.
4. Only then unblind against the ScFuzzBench scaffold/run and classify semantic overlap.
5. Record whether at least one benchmark bug family was autonomously rediscovered, including exact run IDs and reproducible evidence; a well-formed negative result also answers the question.

## Status

In progress. The detached systemd supervisor is healthy and is currently validating the general sequence-length fix before the controlled stateful rerun.

aviggiano:
[...]

PR #528, opened 2026-08-12
fix(runtime): reject stale stateful campaign prompts

### LINKED ISSUE #525: Reject stale invariant campaign prompts that cannot satisfy sequence gates
## Context

#520 / #521 made current stateful Recon campaigns require an explicit `--seq-len 100` and added post-campaign artifact validation. A controlled Aave v4 rerun was then launched from the current Ultrafuzz `main`, but the target project had been initialized before #521 and retained its older project-local copy of `invariant-testing-campaign.md`.

The fresh run therefore compiled a sealed execution snapshot without the sequence-length instruction even though the Ultrafuzz checkout contained the fix. It was canceled before model-authored work and the project was repaired with `ultrafuzz init --force`, after which the stock prompt correctly contained `--seq-len 100`.

## Problem

Project-local prompts are intentionally preserved across ordinary `init` calls so custom prompts are not overwritten. As a result, upgrading the Ultrafuzz package does not make a safety-critical stock prompt change effective for an already initialized project. The current artifact gate prevents such a campaign from being accepted, but it detects the mismatch only after the model has generated and potentially executed a one-hour campaign.

For requirements that the runtime gate makes mandatory, validation should reject an incompatible project prompt before launch instead of allowing an expensive run that cannot produce admissible artifacts.

## Reproduction

1. Initialize a project using a pre-#521 stock prompt catalog.
2. Update Ultrafuzz to a revision containing #521 without forcing replacement of project prompts.
3. Run `ultrafuzz validate --audit-profile invariant-only`; validation currently succeeds.
4. Start an invariant-only run. Its sealed prompt snapshot can omit the required `--seq-len 100` instruction even though the current runtime will later require that evidence.

## Expected behavior

Before submission, validation/planning should report an actionable error when the effective stateful invariant campaign prompt cannot direct the model to produce the sequence evidence required by the current runtime gate. The diagnostic should identify the project-local prompt and explain how to refresh or intentionally update it.

## Acceptance criteria

- [ ] `ultrafuzz validate` and `ultrafuzz run` reject an invariant campaign prompt missing the runtime-required explicit `--seq-len 100` command evidence.
- [ ] The check applies to the effective project prompt, not only the packaged stock template.
- [ ] The diagnostic is protocol-agnostic and points to a safe refresh/update path without silently overwriting custom prompts.
- [ ] Existing compatible custom prompts continue to validate.
- [ ] A regression covers an initialized project retaining the pre-#521 campaign prompt.

Related: #520, #521, #522, #523.

aviggiano:
## Summary

- reject an effective stateful invariant campaign prompt that omits the runtime-mandated --seq-len 100 argument
- surface an actionable validation diagnostic before planning or launching an inadmissible campaign
- cover stale initialized projects in both validate and plan paths

## Why

PR #521 added the runtime sequence evidence gate, but initialized projects can retain an older project-local stock prompt. That mismatch otherwise wastes a campaign before the post-campaign gate rejects it. This closes the pre-launch gap exposed by the controlled rerun associated with #522 and the postmortem in #523.

## Validation

- focused runtime regression: 1 pass, 0 failures
- Prettier check
- ESLint
- runtime typecheck

Closes #525.
Related to #521, #522, and #523.


aviggiano:
Closing as unnecessary. Merged PR #521 already updates the stock invariant campaign prompt to require --seq-len 100 for fresh runs and validates the resulting campaign evidence. Supporting stale project-local prompt copies does not justify this additional prompt-text validation.

PR #576, opened 2026-08-13
Require the canonical current agent registry

### LINKED ISSUE #570: Require the canonical current agent registry during validation
## Problem

Runtime validation still has compatibility behavior for older project registries:

- it accepts the historical .smithers/agents.ts location in addition to .smithers/agents/index.ts;
- it treats a bare textual mention of an agent as an export instead of requiring a current agentFactories entry;
- it validates only selected agents so the default config can remain valid beside a preserved registry from an older release that lacks current opt-in agents.

This leaves an old or incomplete registry accepted until a later launch path selects the missing agent.

## Desired behavior

- Require only the canonical .smithers/agents/index.ts registry.
- Require an agentFactories entry for every agent referenced by configured model profiles.
- Reject old, incomplete, or merely name-containing registries during validate and before run launch.
- Remove documentation and tests that promise compatibility with preserved older registries.
- Preserve normal support for intentional custom current registries.

aviggiano:
## Summary

- accept only `.smithers/agents/index.ts` as the project agent registry
- require an `agentFactories` entry for every configured model-profile agent
- reject incomplete registries and bare textual agent mentions before launch

## Testing

- `pnpm -w build`
- `pnpm -w format:check`
- `pnpm -w lint`
- `pnpm -w typecheck`
- `pnpm -w docs:check`
- affected validation, fan-out planning, workflow compilation, and sync Runtime tests

Fixes #570

<h3>Greptile Summary</h3>

The PR makes `.smithers/agents/index.ts` the sole accepted registry and requires every configured model-profile agent to have an `agentFactories` entry.
- Adds TypeScript-AST-based registry inspection with support for quoted keys, local spreads, static computed keys, aliases, and frozen objects.
- Updates validation, initialization diagnostics, generated workflow factory lookup, documentation, and focused tests.
- Extracts status-duration formatting into a reusable CLI helper.

The PR is not yet safe to merge because valid canonical registries composed from imported factory maps can still be rejected before launch.

The prior spread-validation failure remains reachable: the new collector follows local const spread sources but not imports, while validation treats the missing statically collected keys as fatal even though runtime object lookup would resolve them.

**Files Needing Attention:** packages/runtime/src/agent-registry.ts, packages/runtime/src/validate.ts

| Filename | Overview |
|----------|----------|
| packages/runtime/src/agent-registry.ts | Adds static registry-key extraction, but identifier traversal remains limited to same-file const declarations and leaves the previously reported spread-validation failure reachable for imported maps. |
| packages/runtime/src/validate.ts | Enforces the canonical registry and validates every configured profile against statically discovered factory keys. |
| packages/runtime/src/init.ts | Reuses the new registry parser when detecting stale generated agent registrations. |
| packages/runtime/src/templates/smithers/workflows/workflow.tsx | Imports the canonical agentFactories export directly and fails explicitly when a task factory is absent. |
| packages/runtime/test/runtime.test.ts | Covers quoted keys and local factory spreads, but the outstanding imported-spread behavior is not covered. |

</details>

<sub>Reviews (11): Last reviewed commit: ["fix(runtime): support composed current a..."](https://github.com/monad-developers/ultrafuzz/commit/5d36e8ddade630aed6cd926d08e6c1b26e14c326) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=53037580)</sub>

**Context used:**


aviggiano:
Addressed the independent review findings in 673aaeff: validation now resolves local named export aliases, statically unwraps `Object.freeze`, and recognizes static computed string keys. Added a regression covering all three forms; focused tests and workspace quality checks pass.

aviggiano:
CI follow-up: the CLI lane exposed an existing duration-rendering fixture race twice (at different loop iterations) while the full local suite passed 150/150. Commit 742f913a keeps the 10-minute end-to-end assertion and tests the pure duration formatter directly for minute/hour/day boundaries, eliminating fake-engine clock synchronization from that formatting coverage.

aviggiano:
@aviggiano please review

packages/cli/src/status-rendering.ts
```diff
@@ -0,0 +1,14 @@
+export function formatStatusDuration(seconds: number): string {
```
aviggiano, line 1: Can we use date-fns or something

aviggiano: Keeping the dedicated formatter here. date-fns is not an existing CLI dependency, and this status output has a small product-specific contract (for example `less than a minute`, `1h 30m`, and `3d 00h`) that a generic date formatter would still need custom wrapping to preserve.

aviggiano:
Final-review decision on the current Greptile summary: no change for opaque imported factory maps. Issue #570 requires validation to prove that every configured agent has an entry in the canonical registry. Treating an imported spread as coverage without inspecting its keys would restore name-presence-style false acceptance, while executing project-owned modules during validation would be unsafe. Commit 5d36e8dd supports local const/frozen/spread composition whose keys are statically provable in the canonical registry; imported factory functions remain supported when those keys are declared there.

Issue #678, opened 2026-08-19
Non-pinned runs base every task worktree on main instead of the revision the run was launched from


aviggiano:
## Summary

When a run is not using the pinned benchmark branch, the generated workflow passed no `baseBranch` to `<Worktree>`. Smithers defaults that to `"main"`, so every task worktree is branched from `main` regardless of the revision the run was actually launched from. Agents then audit a tree the run was never pointed at, and the run's own commit pinning is not what executes.

## Evidence

In the shipped template, `baseBranch` was only supplied for the pinned-source case:

```tsx
<Worktree
  key={task.id}
  path={task.workspacePath}
  branch={task.branch}
  {...(usesPinnedSource ? { baseBranch: pinnedSourceBranch } : {})}
>
```

and `usesPinnedSource` is true only when a local `ultrafuzz-pinned` branch exists:

```ts
function sourceUsesPinnedBranch(): boolean {
  return invariantPinnedSourceRefExists(process.cwd(), pinnedSourceRef);
}
```

Smithers' own documentation confirms the default and the requirement:

> `baseBranch` defaults to `main` and should name a committed branch or described commit

So for any run launched from a branch other than `main` without `ultrafuzz-pinned` present — i.e. every ordinary non-benchmark run — task worktrees silently derive from `main`.

## Impact

This is a correctness problem, not a convenience one. An audit's central promise is that it examined a specific revision:

- The run pins and reports a target commit, while its task worktrees are based on something else. The reported provenance and the executed source disagree, and nothing in the run surfaces the divergence.
- If `main` is ahead of the target revision, agents analyse code the target does not contain; if behind, they miss code it does.
- If `main` does not exist at all in the checkout, worktree creation fails outright — a plausible contributor to opaque `prepare:*` failures, since worktree setup is exactly what those nodes do (see the `prepare:` TypeError issue).

## Fix applied

Resolve the launch revision rather than naming a branch:

```ts
function resolveLocalSourceCommit(): string | undefined {
  if (usesPinnedSource) return undefined;
  try {
    const commit = execFileSync("git", ["rev-parse", "HEAD^{commit}"], {
      cwd: process.cwd(),
      encoding: "utf8"
    }).trim();
    return /^[0-9a-f]{40}$/u.test(commit) ? commit : undefined;
  } catch {
    return undefined;
  }
}
```

used as `baseBranch={usesPinnedSource ? pinnedSourceBranch : localSourceCommit}`.

[...]

Issue #675, opened 2026-08-19
goals/strategies group timeout_seconds: 7200 shadows the profile default and kills high-reasoning nodes at 2h with total loss of node work


aviggiano:
## Summary

Follow-up to #645. That issue raised the default-profile per-agent timeout, but the `goals` and `strategies` groups pin `timeout_seconds: 7200` in the shipped topology, and the group pin wins over the profile/config default. High-reasoning routed models are still killed at the 2h boundary, and a timeout kill discards the node's entire output.

## Evidence

`.ultrafuzz/topology.yml` group defaults:

```yaml
groups:
  goals:
    defaults:
      loops: 1
      timeout_seconds: 7200
      model_profiles: [default]
  strategies:
    defaults:
      loops: 3
      timeout_seconds: 7200
      model_profiles: [default]
```

The resolution order in `packages/runtime/src/smithers.ts:3734` takes the node value first, so the group pin shadows both the profile and `run.default_timeout_seconds`:

```ts
const timeoutMs =
  (input.node.timeoutSeconds ?? profile.timeoutSeconds ?? input.config.run.defaultTimeoutSeconds) * 1000;
```

In an 18-hour run configured with `default_timeout_seconds = 3600`, goal nodes were killed at exactly the group pin, not the config value:

```
+07:23:46  NodeStarted  → node:<dynamic-goal-a> (attempt 1, iteration 0)
+09:23:50  NodeFailed   ✗ node:<dynamic-goal-a> (attempt 1): CLI timed out after 7200000ms
+09:07:26  NodeStarted  → node:<dynamic-goal-b> (attempt 2, iteration 0)
+11:07:32  NodeFailed   ✗ node:<dynamic-goal-b> (attempt 2): CLI timed out after 7200000ms
```

Nine failures in the run were this exact 7200000ms kill. Every one lost the whole node: the artifact directories for timed-out goal nodes contain only `prompt.rendered.md` and `workspace-patch-preparation.json`, no published output.

## Two distinct problems

**The window is too short for the workload.** Retrying does not help, because a node that needs more than 2h of high-reasoning wall time hits the identical wall on attempt 2. Observed: a node killed at 2h on attempt 1, restarted, and on track to be killed at 2h again.

[...]

Issue #672, opened 2026-08-19
Concurrent prepare:* worktree tasks fail nondeterministically with a Bun-erased TypeError: undefined is not an object (evaluating 'get')


aviggiano:
## Summary

Concurrently dispatched `prepare:*` (artifact-mirror / `Worktree`) tasks fail nondeterministically with `TypeError: undefined is not an object (evaluating 'get')`. Because these nodes are compiled with `retries={0}`, a single occurrence permanently fails the node and cascades into every downstream artifact contract.

This was the **largest single source of node failures** in an 18-hour local `default`-profile run: 55 of 214 `NodeFailed` events, **53 of them on `prepare:` nodes**.

## Evidence

Failure taxonomy for the run (214 `NodeFailed` events total):

| count | failure class |
| --- | --- |
| 97 | `artifact-contract failure: artifact dependency has not …` (downstream cascade) |
| **55** | **`TypeError: undefined is not an object (evaluating 'get')` (53 on `prepare:` nodes)** |
| 41 | `unexpected status 402 Payment Required` (separate provider-credit outage) |
| 9 | `CLI timed out after 7200000ms` |
| 5 | `stream disconnected before completion` |
| 4 | artifact schema validation |
| 3 | one each: workspace patch base tree mismatch, goal-plan child-count, 3600000ms timeout |

The 97 artifact-contract failures are consequences, not independent defects — which makes this TypeError the dominant root cause in the run.

### Race signature

Two `prepare:` nodes dispatched in the same second, one succeeds, the other throws:

```
[00:14:53] → prepare:threat-model (attempt 1, iteration 0)
[00:14:53] → prepare:setup-foundry (attempt 1, iteration 0)
[00:14:55] ✓ prepare:threat-model (attempt 1)
error task execution failed  nodeId=prepare:setup-foundry
      iteration=0 attempt=1 maxAttempts=1
      error="undefined is not an object (evaluating 'get')"
[00:14:56] ✗ prepare:setup-foundry (attempt 1): undefined is not an object (evaluating 'get')
```

A later burst shows the same pattern at higher fan-out — six `prepare:` nodes started within five seconds, some finishing and some failing with the identical message:

```
+03:01:12  NodeFailed   ✗ prepare:<dynamic-goal-a> (attempt 1): undefined is not an…
+03:01:13  NodeFailed   ✗ prepare:<dynamic-goal-b> (attempt 1): undefined is not an…
[...]

Issue #653, opened 2026-08-18
Support provider-agnostic coding harnesses via tight Smithers adapters


aviggiano:
Umbrella for provider-agnostic harness support, rewritten around the tight-adapter ruling.

## Governing constraint

Ultrafuzz carries **tight adapters** over Smithers' built-in harness registry and never re-implements Smithers internals. Smithers ships 14 built-in CLI harnesses and already exposes every binding axis as constructor options: harness = the agent class, provider = baseURL + apiKey, model = an opaque provider catalogue ID, protocol = the class. `smithers agents capabilities --format json` publishes each harness's binary, emittedFlags, optionMappings and resume kind; `smithers agents doctor` and `smithers agents test` already validate reachability.

A tight adapter is one ~76-line file in `packages/runtime/src/templates/smithers/agents/`, modelled on `claude.tsx`, that reads only `auth` / `api_key_env` / `config_dir` from its own `[agents.<Ref>Agent]` table, maps them onto existing Smithers options, wraps child env in `workflowControlChildEnvironment` and credentials in `workflowControlCredentialValue`, and does nothing else.

## What this umbrella no longer includes

- **No new configuration schema.** No `[providers.*]`, no `[harnesses.*]`, no independent `harness` / `provider` fields, no `ProviderBinding` / `HarnessCapabilities` / `QualifiedHarnessBinding` types, no binding validator. `models.*.agent` is already an open agent-ref schema and `[agents.*]` is already a generic record, so a new harness needs zero config work. (#658 closed; PR #682 rejected.)
- **No ultrafuzz-side capability contract.** Capability truth is `smithers agents capabilities` / `doctor`. Ultrafuzz consumes it; it does not define a second one.
- **No G1-G10 conformance suite.** Qualification for a pairing is: one real provider smoke run on a pinned CLI, plus two unit assertions (credential absent from argv, credential present in child env) and one filesystem assertion (no harness state outside the run-scoped roots).

## Sequencing (children are now independent)

- #659 — Pi + OpenRouter tight adapter
- #660 — correct the DeepSeek pairing terminology (docs/comments only)
- #661 — DeepSeek Harness (`dsh`): **not a tight adapter** — no Smithers registry entry exists. Owner decision required; recommended resolution is upstream.
- #662 — OpenCode + OpenRouter tight adapter with run-scoped state
- #663 — Modal image: pin the new harness packages and extend the existing command prober
- #664 — closed; residual doc wording folded into #660

[...]

PR #689, opened 2026-08-19
fix(runtime): give agentic nodes a retry budget, base worktrees on the launch revision, and make preparation failures diagnosable

### LINKED ISSUE #673: Shipped topology declares no max_attempts, so every agentic node compiles with retries: 0 and #572's bounded retries never engage
## Summary

The shipped topology declares `max_attempts` nowhere, and `maxAttemptsFor` falls back to `1`. Every agentic node in a real run therefore compiles with `retries: 0`, so the bounded agent retry policy added in #572 never engages. Any single stochastic provider failure permanently fails the node.

## Evidence

`packages/topology/src/expand.ts:193-197`:

```ts
function maxAttemptsFor(node: NormalizedTopologyNode, topology: NormalizedProjectTopology): number {
  if (node.kind !== "agentic") {
    return 1;
  }
  return node.max_attempts ?? (node.group ? topology.groups[node.group]?.defaults?.max_attempts : undefined) ?? 1;
}
```

Both override sites are unused — `.ultrafuzz/topology.yml` contains **zero** occurrences of `max_attempts`:

```
$ grep -c max_attempts .ultrafuzz/topology.yml
0
```

That flows straight through to the compiled plan. `packages/runtime/src/smithers.ts:3742`:

```ts
const retries = Math.max(0, input.node.retryPolicy.maxAttempts - 1);
```

and in the compiled task manifest for an 18-hour `default`-profile run, all 179 tasks:

```
retries distribution: {0: 179}
```

Confirmed in the run log at the point of failure:

```
error task execution failed  nodeId=prepare:setup-foundry
      error="undefined is not an object (evaluating 'get')"
```

## Why this matters more than it looks

#572 established the policy — bounded, error-agnostic retries with a fresh generation on the original prompt — and it is correctly implemented in the runtime. It is simply never reached, because nothing in the shipped configuration opts a single node into `max_attempts > 1`. The feature is effectively dead code in the default profile.

The operational consequence observed in a real run: with no retry budget, the only way to give a failed node another attempt was to hand-edit the *sealed execution snapshot* of the workflow to force `retries={Math.max(task.retries, 1)}`. That patch bought the retries but permanently invalidated the run's workflow control evidence, disabling product-level status and reporting for the rest of the run's life. A missing default turned into an unrecoverable run. See the control-seal issue filed alongside this one.

## Impact

- Any transient provider error, stream drop, or race is terminal for that node, and usually for everything downstream of it.
- #572's retry work delivers no benefit in practice.
- Operators are pushed toward unsupported snapshot edits to recover, which breaks integrity verification.

## Proposal

- Set `max_attempts` in the group defaults of `.ultrafuzz/topology.yml` for the agentic groups (at minimum `goals` and `strategies`, which carry the long high-reasoning nodes).
- Reconsider whether the `?? 1` fallback in `maxAttemptsFor` is the right default for `kind: agentic` at all, or whether agentic nodes should default to more than one attempt with `1` reserved as an explicit opt-out.
- Assert the wiring in a test: compile a plan from the shipped topology and assert agentic tasks carry `retries >= 1`, so this cannot silently regress to zero again.

## Acceptance criteria

- Agentic nodes compiled from the shipped topology have a non-zero retry budget.
- A regression test fails if the shipped topology ever compiles agentic tasks to `retries: 0`.
- Recovering a failed node never requires editing a sealed execution snapshot.

aviggiano:
Fixes the reliability defects that made a long agentic run unable to survive its own transient failures. Every change here is committed as `aac0667d` and was validated before this PR was opened.

Closes #673. Refs #672, #675, #678, and #572 / #645 which regressed.

> **Rebased onto current `release/v0.1.0`.** Two things this branch originally carried are now
> dropped because upstream got there first and did it better: the Codex api-key auth fix (#642) is
> superseded by the provider-routing work from #653/#665, which resolves the base URL through
> `routing.route.baseUrl` rather than the narrower patch here; and the `fuzz-only.yml` topology
> change is moot because `fix(config): simplify audit profiles (#651)` removed that file. What
> remains is the retry budget, the worktree base fix, and the preparation diagnostics.

## Why

An 18-hour local `default`-profile run finished 138 of 537 steps and then stopped advancing. Its 214 `NodeFailed` events break down as:

| count | failure class |
| --- | --- |
| 97 | `artifact-contract failure: artifact dependency has not …` (downstream cascade) |
| 55 | `TypeError: undefined is not an object (evaluating 'get')` — 53 on `prepare:` nodes |
| 41 | `unexpected status 402 Payment Required` (provider credit outage) |
| 9 | `CLI timed out after 7200000ms` |
| 5 | `stream disconnected before completion` |
| 4 | artifact schema validation |
| 3 | one-offs |

The 97 cascade events are consequences. The run's inability to absorb any of the rest is the defect.

## Changes

**Retry budget (#673).** `maxAttemptsFor` falls back to `1`, and no production topology set `max_attempts` anywhere — only `smoke.yml` did. Every agentic node therefore compiled to `retries: 0`, so #572's bounded retry policy never engaged and a single stochastic provider failure was terminal for a node and everything downstream. Verified against a real compiled plan: all 179 tasks had `retries: 0`, and failures logged `maxAttempts=1`.

The cheap CI lane retried; the three real audit lanes did not:

| topology | agentic nodes | groups with `max_attempts` |
| --- | --- | --- |
| `smoke.yml` | 7 | setup, strategies, review |
| `full.yml` | 55 | none |
| `fuzz-only.yml` | 50 | none |
| `invariant-only.yml` | 23 | none |

`max_attempts: 2` now ships on every group holding agentic nodes, following `smoke.yml`'s existing key and value.

**Preparation diagnostics (#672).** Bun discards the user frames of an error thrown inside a task body, so all 53 `prepare:` failures arrived with a stack of only `at run (node:async_hooks:68:37)` — 26% of every failure in the run, with no file or line to bisect. Preparation is synchronous, so naming each of its twelve steps recovers the one fact the stack cannot carry: which step threw. `cause` keeps the original error intact.

This does not fix the underlying race — the evidence points to concurrent worktree preparation contending, but the erased stack is exactly why that is not yet confirmed. It makes the next occurrence diagnosable. Note that `prepare:` tasks compile with `retries={0}` independently of the group defaults above, so these failures remain terminal; that is tracked separately and deliberately not changed here.

**Worktree base (#678).** A non-pinned run passed no `baseBranch`, so Smithers defaulted every task worktree to `main` regardless of the revision the run was launched from — the reported target commit and the executed source could diverge silently, and a checkout without `main` failed outright. Now resolves `HEAD` to a commit id, which is what the run pinned, needs no branch to exist, cannot drift mid-run, and stays within Smithers' documented contract (`baseBranch` accepts a described commit). Falls back to the previous default outside a git checkout or on an unborn `HEAD` rather than failing generation.

**Group timeout resolution (#675).** Reverts a `7200 → 14400` change to the `goals` and `strategies` pins. Raising the window treated the symptom; the real cause of those 9 timeouts is addressed in the companion PR. The resolution-order problem #675 describes is real and stays open — a group pin shadows the profile default, which is why #645 raising the profile default never reached these nodes — and is now covered by a two-sided, mutation-tested guard rather than a one-sided floor.

**Also landed, previously uncommitted and untested in the tree.** `runGit` supplies a consistent environment on both the index and non-index paths, and goal-plan selected records accept YAML. The Codex auth patch that was also sitting there is dropped in favour of upstream's provider routing, as noted above.

## Validation

Both topology changes are guarded by tests that fail on the pre-fix files, so neither can silently regress to zero retries or a stale pin. The `prepareArtifactMirror` rewrite preserves every source-text assertion the generated-workflow verifier makes about it, including call ordering.

- runtime: 645 pass / 0 fail / 17 skipped
- topology 59, config 39, artifacts 139 — all pass
- `prettier --check`, `eslint`, `runtime typecheck`, `validate-audit-profile-package.mjs`: all clean

<h3>Greptile Summary</h3>

The PR improves long-running audit reliability by giving agentic nodes a bounded retry budget, basing ordinary task worktrees on the launch commit, and attributing preparation failures to individual steps.

- Adds one retry to agentic groups in the shipped full and invariant-only topologies, with expansion-level regression coverage.
- Resolves non-benchmark worktrees from the current commit while preserving the pinned benchmark branch behavior.
- Wraps artifact-mirror preparation stages with named diagnostics while preserving original errors as causes.
- Extends goal-plan selected-record support and verification to YAML vulnerability records.
- Normalizes Git subprocess environments across temporary-index and ordinary workspace-handoff operations.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| packages/runtime/src/templates/smithers/workflows/workflow.tsx | Resolves ordinary worktrees from the launch commit, adds step-specific preparation diagnostics, and verifies selected goal-plan record snapshots. |
| packages/runtime/src/workspace-handoff.ts | Supplies a consistent inherited environment to Git subprocesses whether or not a temporary index is active. |
| packages/artifacts/src/goal-plan.ts | Allows selected vulnerability records to use the YAML extension required by newer database schemas. |
| packages/config/topologies/full.yml | Adds a two-attempt budget to every agentic group in the packaged full topology and raises long-running group timeout pins. |
| packages/config/topologies/invariant-only.yml | Adds the same bounded retry policy and long-running timeout coverage to the invariant-only topology. |
| packages/topology/test/packaged-topologies.test.ts | Adds expansion-level guards ensuring shipped agentic nodes retain retry budgets and group timeout pins do not undercut the long-running window. |

</details>

```mermaid
flowchart LR
  T[Topology group defaults] --> E[Expanded agentic node]
  E -->|maxAttempts 2| C[Smithers task with one retry]
  L[Launch checkout HEAD] --> R[Resolve immutable commit]
  R --> W[Task worktree base]
  W --> P[Named preparation steps]
  P --> A[Agent execution]
  A -->|transient failure| C
  A --> V[Artifact and selected-record verification]
```
</details>

<sub>Reviews (2): Last reviewed commit: ["Merge release/v0.1.0 into fix/run-reliab..."](https://github.com/monad-developers/ultrafuzz/commit/bf88dec499088d83e301f49a509c123d0ed76900) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=54921278)</sub>

<details><summary><h4>Context used (4)</h4></summary>

- Knowledge Base — [Config (@ultrafuzz/config)](https://app.greptile.com/monad-foudnation/-/custom-context/knowledge-base/monad-developers/ultrafuzz/-/docs/config.md)
- Knowledge Base — [Topology](https://app.greptile.com/monad-foudnation/-/custom-context/knowledge-base/monad-developers/ultrafuzz/-/docs/topology.md)
</details>


aviggiano:
Good to merge, just fix conflicts

Issue #688, opened 2026-08-19
continueOnFail on the cloud worker's own render costs a goal lane its retry: the completed checkpoint makes the sandbox retry skip the agent


aviggiano:
## Summary

Applying `continueOnFail` to a goal lane on the **cloud worker's own render** silently costs that lane its retry. The inner run finishes despite a killed agent, the worker records a `completed` durability checkpoint, and the sandbox retry then skips re-running the agent entirely — spending the second attempt on a publication retry instead. A cloud goal lane gets one real attempt where the topology grants two.

Found while adversarially re-verifying the goal-lane fault-tolerance change. Local execution is unaffected.

## The chain

The relocated worker renders the same `<Worktree>` branch the local path uses, for its own single task. With `continueOnFail` applied there:

1. The worker's goal agent is killed. The agent task is `failed`, but `continueOnFail` makes it **terminal**, so its verifier becomes runnable.
2. `verifyArtifacts` correctly throws — the agent produced no output row — but `continueOnFail` on the verifier absorbs that too, so **the inner run finishes**. `smithers up` exits 0 and the worker's `runChecked("resume-workflow", …)` does not throw.
3. `durableWorkspace.recordCheckpoint("completed")` is therefore recorded *before* publication is attempted.
4. Publication then fails closed, correctly: `workerResultPublicationMode` resolves to `verified-v2` because the workflow contains the marker strings, and `copyAttemptVerificationMarker` throws `cloud publication verification marker is missing`.
5. **But** the sandbox retry re-enters with `hasCompletedCheckpoint === true`, so the `if (!durableWorkspace.hasCompletedCheckpoint)` guard skips `runDurableWorkflow` altogether and merely re-fails publication.

Before the tolerance was added, step 2 left the inner run *failed*, no `completed` checkpoint was written, and the retry re-ran the agent as intended.

## Why the worker does not need the tolerance

`continueOnFail` exists to stop one failing lane from taking its siblings and the fan-ins down with it. A worker render has exactly one task — `selectedTaskSpecs` is that single task — so there are no siblings to protect and nothing to gain, while there is a retry to lose.

An honest worker failure also produces the identical controller-side outcome: no agent output row, so the controller's `agentReturned` is false either way and the lane is correctly left unattested.

## Fix

Gate the two local-path sites on the worker:

```tsx
continueOnFail={goalSearch && !cloudWorker}
```

[...]

Issue #686, opened 2026-08-19
Goal-hunter record-lookup instructions name fields that do not exist for threat goals and a manifest path that never resolves


aviggiano:
## Summary

The goal-hunter prompt's record-lookup instructions name fields that do not exist for one of the two goal kinds, point at a manifest field that resolves to a non-existent path, and omit a join the old inlined payload had pre-done. Measured against a real 88-goal plan, the instructions are wrong for **all 88 goals** in at least one respect.

These surfaced when full records stopped being inlined into the goal sentence and the hunter began having to locate them itself.

## Defect 1 — 11 of 88 hunters are told to read fields that do not exist

The instructions name `class_id`, `threat_ids`, `attack_surface_ids`, `selected_record`, and `selection_rationale` as "the definition of your scope", and tell the agent to read `selected_record.path`.

Threat-goal entries have **no `class_id` and no `selected_record` at all.** What a threat goal carries is `class_ids` — plural, between 1 and 29 entries — which the instructions never mention. Those 11 hunters have to guess.

Every threat-goal `class_ids` value does resolve into the plan's `selected_class_records`, with zero misses, so the information is available; only the instruction is wrong. The fix is to distinguish the two goal kinds and name the plural key, stating that a threat goal selects among the selected class records by those IDs.

## Defect 2 — a 77-of-77 foot-gun that yields ENOENT every time

The instructions say the record is "indexed by `vulnerability-db-manifest.json`". Each `selected_records[]` entry in that manifest carries **both**:

- `path` — database-relative
- `artifact_path` — goal-plan-artifact-relative

They differ for all 77 entries. Only `artifact_path` resolves under the artifact directory, and it equals the plan's own `selected_record.path` for all 77. The manifest additionally has a 141-entry `records` array whose files were never copied into the artifact — only `vulnerability-db/selected/` exists on disk.

An agent that follows "the index" and takes the obvious field name `path` gets ENOENT every time. Either name `selected_records[].artifact_path` explicitly, or drop the manifest clause entirely, since `selected_record.path` is already correct and sufficient.

## Defect 3 — a missing dereference the old payload had pre-done

The instructions tell the agent to read the threat entry's "assets, actors, attack surfaces, preconditions, invariants, assumptions, unknowns". A threat entry carries only `*_ids` for six of those; the objects live in the threat model's **top-level** `assets` / `actors` / `attack_surfaces` / `invariants` / `assumptions` / `unknowns` arrays.

[...]

Issue #685, opened 2026-08-19
No way to gate run success on goal-search coverage: a 3-of-77 audit reaches the same succeeded status as a complete one


aviggiano:
## Summary

There is no way to make a run's terminal status depend on how much of the audit actually ran. Now that goal lanes may legitimately end empty, a run that searched a small fraction of its planned goals reaches the same `succeeded` status as one that searched all of them. An unattended consumer — CI, a dashboard, the eval harness — reads that status, not the report prose.

Filed as the counterpart to the goal-coverage reporting gap: reporting fixes the human reader, this fixes the machine reader.

## Why a hard built-in threshold is the wrong answer

The obvious fix is refusing `succeeded` below some coverage percentage. That should not be the default, for two reasons:

- **Any baked-in threshold is arbitrary.** Some goals legitimately cannot complete — a goal targeting a capability the target does not have is correctly abandoned, and the applicability decisions in the plan already record that. A fixed floor either blocks legitimate runs or is set so low it certifies nothing.
- **`succeeded` has a defensible existing meaning:** the workflow ran to completion. Redefining it to mean "and covered enough" conflates orchestration health with audit thoroughness, which are separately useful signals.

The actual defect is that coverage is invisible to anything that consumes the run programmatically, and that there is no supported way for an operator to say "do not treat this as a passing audit".

## Proposal

Three parts, in priority order:

1. **Expose coverage as a first-class machine-readable verdict** on the run, beside the terminal status — not only as a field inside `report.json`. A consumer should be able to read completed/planned and the stopped-early count without parsing the report.
2. **Add an opt-in gate** — e.g. `--min-goal-coverage <ratio>` on the run and status paths, plus a config equivalent — that fails validation when coverage falls below the operator's chosen bar. Opt-in keeps the policy with the operator, where it belongs, and gives CI and the eval harness a real gate.
3. **Never let unknown read as complete.** A missing census must produce an explicit unknown verdict, and the opt-in gate must treat unknown as failing rather than passing.

Deliberately not proposed: a new terminal status such as `succeeded-partial`. Terminal statuses are consumed in several places (`isTerminalRunStatus`, the deadline check in `workflow-control.ts`, the eval runner) and adding one risks changing control-flow behaviour well outside reporting.

## Acceptance criteria

- Goal coverage is readable programmatically from run state without parsing the report.
- An operator can configure a minimum coverage below which the run does not validate as a passing audit.
[...]

Issue #684, opened 2026-08-19
Goal-search coverage reaches report.json but never report.md, so a partial audit reads as clean


aviggiano:
## Summary

Goal-search coverage is recorded and reaches `report.json`, but the renderer never emits it and the CLI overwrites `report.md` from that renderer. So the document a reader actually receives says nothing about how much of the audit ran. Once goal lanes are allowed to end empty, that silence lets a run which searched 3 of 77 goals present itself as a clean audit.

## Evidence

The census itself is sound. `recordGoalSearchCoverage` derives per-lane status from durable output rows rather than from any agent:

| status | meaning |
| --- | --- |
| `stopped-early` | no task output row — timed out, killed, or failed outright |
| `unverified` | task row but no verification row |
| `completed-with-findings` / `completed-no-findings` / `completed` | ran to completion |

with totals `planned`, `completed`, `completed_with_findings`, `completed_no_findings`, `stopped_early`, `unverified`, plus a per-goal array of `node_id`, `logical_node_id`, `attempt_id`, `status`, `finding_count`.

`reconstructAuthoritativeGoalSearchCoverage` stamps `report.goal_search_coverage` into `report.json`, and `reconcileIssuePresentation` returns `{...report, ...}` so it survives into the written JSON.

It stops there:

1. **The renderer never reads it.** `renderCanonicalReport` (`packages/runtime/src/final-report-markdown.ts`) appends run summary, audit context, campaign outcome, issues, `appendPropertyImplementationCoverage`, property provenance, prior disposition, and non-production outcomes. `report.goal_search_coverage` is never referenced. Grepping `goal` in that file finds only an `## Audit context` link to `goal-plan.json` — a pointer to what was *planned*, with nothing about what *ran*.
2. **That renderer is the sole producer of `report.md` on both paths,** and the CLI writer (`packages/cli/src/report-artifacts.ts`) overwrites whatever the agent wrote. Its reconciler chain has `reconcileCampaignOutcome` and `reconcilePropertyImplementationCoverage` but no goal-coverage equivalent.
3. **No prompt names the census.** `grep -rn "goal-search-coverage\|goal_search_coverage" .ultrafuzz/prompts` returns nothing. `review/final-report.md` requires the report to state *property* coverage and says nothing about goal coverage. The workflow code admits this itself: the stamp lands after the report agent's own writes, so that agent cannot see its own coverage while writing.

[...]

PR #690, opened 2026-08-19
fix(goals): shrink goal prompts from inlined JSON to path references, and stop killed goal lanes being attested as verified empty results

### LINKED ISSUE #680: Goal planner inlines entire JSON records into the goal prompt, making each goal statement a multi-thousand-token wall
## Summary

The goal planner is instructed to put **entire JSON records** into each dynamic goal's `replacements` map, so a 161-character goal prompt expands into a multi-thousand-token JSON wall — which the hunter prompt then introduces as "the first sentence above is the authoritative focused goal". This is the root cause of the goal-node timeouts previously attributed to model speed in #675.

## Evidence

Measured across all 88 goals of one real `default`-profile run:

| | characters | approx tokens |
| --- | --- | --- |
| `goal_prompt` template | 161 | ~40 |
| replacement payload, min | 3,751 | ~940 |
| replacement payload, **median** | **13,956** | **~3,500** |
| replacement payload, max | 42,483 | ~10,600 |

The template itself is exactly the crisp instruction you would want:

```
Your /goal is to find a vulnerability of type {{class:<class-id>}} using threat model {{<threat-id>}}.
```

Both placeholders resolve from the item's `replacements` map. For one representative class goal, that map held two entries of 10,051 and 4,744 characters — each a full raw JSON record: the vulnerability-class record with its `guidance` sub-object, and the threat-model entry with nested `attack_surfaces`, `assets`, `actors`, and every `evidence` array.

### It is mandated by the planner prompt

`.ultrafuzz/prompts/setup/goal-plan.md`, under "## MDX replacement contract":

> The replacement value for each threat key contains the full selected threat, assets, actors, surfaces, preconditions, invariants, assumptions, unknowns, and evidence. The class replacement contains the database record's focused hunter instructions and relevant examples. **Never flatten a placeholder to a bare literal ID.**

So the bloat is the specified behaviour, and the last sentence explicitly forbids the fix.

### Nothing constrains the size

`replacementsSchema` in `packages/artifacts/src/goal-plan.ts` accepts any non-empty string:

```ts
const replacementsSchema = z
  .record(replacementKey, z.union([nonEmptyString, z.number().finite(), z.boolean()]))
  .refine((value) => Object.keys(value).length > 0, { message: "At least one item-scoped replacement is required" });
```

A 42,483-character "replacement" is contract-valid.

## Impact

At maximum reasoning effort, every goal node must ingest and reason over several thousand tokens of dense nested JSON **before** it begins analysing any repository code, and it is told that blob is its goal statement.

In the run measured: 9 nodes were killed at exactly the 7200000ms group timeout, and only **3 of 77** class-goal nodes ever produced output. The goal phase is the most valuable phase of the audit and it completed 4% of its work.

This also explains why #675 looked like a timeout problem. Raising the timeout is the wrong fix: with a 14k-character goal statement, more time buys a longer failure and doubles the worst-case cost of a stuck node.

## The fix is already the pattern used three lines away

`goal-hunter.mdx` references its other large inputs **by path**, not by value:

```
Canonical threat model:
{{artifact_path:threat-model}}/threat-model.json

Goal plan and node mapping:
{{artifact_path:goal-plan}}/goal-plan.json

Goal-plan artifact directory containing the selected records under
`vulnerability-db/selected`:
{{artifact_path:goal-plan}}
```

The agent already has every full record available on disk and is already told where. Inlining the same JSON into the goal sentence is redundant with the file's own established convention.

Proposed:

- Replacement values become a short human-readable label — the record's `title`, optionally with its `id`. The hunter reads the full record from the paths it is already given.
- Drop the "never flatten to a bare literal ID" instruction; keep the requirement that placeholders stay brace-wrapped in `goal_prompt` until the dynamic renderer resolves them, since `validatePromptReplacements` depends on it.
- Make `replacementsSchema` enforce a maximum length and reject values that are structurally JSON, so a planner that ignores the prompt cannot silently reintroduce the regression.

## Acceptance criteria

- A goal node's rendered goal statement is on the order of hundreds of characters, not tens of thousands.
- The hunter can still locate the complete class record and threat-model entry for its assigned goal without guessing.
- An oversized or JSON-shaped replacement value fails contract validation with an actionable message.
- A regression test fails if a contract-valid goal plan can carry a replacement large enough to be a JSON record.

aviggiano:
Fixes the reason the goal phase completed 4% of its work, and makes a partially-covered audit unable to present itself as a clean one.

Closes #680, #683, #684, #686, #688. Refs #675, #685.

## The root cause

The goal planner was instructed to inline **entire JSON records** into each dynamic goal's `replacements` map. The `goal_prompt` template is a crisp ~161-character sentence; the placeholders in it resolved to full records. Measured across all 88 goals of a real plan:

| rendered goal statement | min | median | max |
| --- | --- | --- | --- |
| before | 3,861 | **14,024** | 42,555 |
| after | 156 | **296** | 591 |

**47× reduction at the median, 72× at the max**, ~342,000 tokens saved across the goal fleet. The "after" figures are measured, not estimated: each replacement was rebuilt from the referenced record's actual `title` in the fixture and re-substituted into that goal's real template.

So every goal node's *goal statement* was a 3.5k–10.6k-token wall of nested JSON, introduced to the agent as "the first sentence above is the authoritative focused goal", at maximum reasoning effort, before it read a line of the target. In the measured run 9 nodes were killed at exactly the 7200000ms limit and only **3 of 77** class-goal nodes produced any output.

`.ultrafuzz/prompts/setup/goal-plan.md` mandated it, and explicitly forbade the fix: *"Never flatten a placeholder to a bare literal ID."* Nothing constrained the size — a 42,483-character "replacement" was contract-valid.

The fix is the pattern the same prompt already uses three lines away: the threat model and goal plan are referenced **by path**, not by value. Replacements become short labels; the hunter reads the records from paths it is already given. `replacementsSchema` now enforces a 200-character cap and rejects structurally-JSON values, so a planner that ignores the prompt cannot silently reintroduce the wall.

## Goal searches may end empty — without laundering

A goal search that finds nothing is a legitimate negative result; the hunter prompt already contracts `[]` for it. Previously one killed lane cascaded into 97 downstream `artifact-contract failure` events. Goal lanes now carry `continueOnFail`, scoped by `metadata.node.group === "goals"` (verified present on all 179 compiled tasks of a real plan).

**That tolerance, alone, created a worse defect (#683), which this PR also fixes.** Preparation deliberately seeds every goal lane's mirror with a canonical empty `findings.json`. Once a killed lane is terminal, its verifier becomes runnable, and a filesystem-only verifier finds the seeded `[]`, validates it, publishes it to the canonical artifact directory and attests it. A run where 3 of 77 lanes searched would emit 74 verified, published, empty findings artifacts for lanes that never ran, counted downstream as genuine zero-finding sources.

`verifyArtifacts` now requires `{ agentReturned }`, sourced from Smithers' own durable output row rather than from the workspace — bytes in the worktree prove nothing, since an agent killed mid-search after writing an early empty `findings.json` leaves exactly what a real negative leaves. A refusal **throws** rather than returning a row, deliberately: `outputs.verification` is itself consumed as evidence, so a "verified nothing" row would relocate the lie one level up. The check is not scoped to goal lanes, so any future `continueOnFail` fails closed here by default.

## Coverage is now stated, not implied

The census (`goal-search-coverage.json`) derives per-lane status from durable output rows, never from an agent. It reached `report.json` but the renderer never emitted it and the CLI overwrote `report.md` — so a 3-of-77 audit produced a document whose entire statement about coverage was "No issues reported" (#684). Coverage is now rendered into `report.md` on both paths, an absent or unreadable census renders as an explicit *unknown* rather than as silence or as complete, and `review/final-report.md` and `review/dedupe-findings.md` name the census so the reporting and reviewing agents can see the numbers while they work.

Deliberately not included: a threshold gating run success on coverage. That is #685, and a default threshold would be a guess — better set from a real coverage number than invented now.

## Cloud worker keeps its retry (#688)

Applying the tolerance on the cloud worker's own single-task render cost the lane its retry: the inner
run finished despite a killed agent, the worker recorded a `completed` durability checkpoint, and the
sandbox retry then skipped re-running the agent and merely re-failed publication. A worker render has
one task and therefore no siblings to protect, so the two local-path sites are now gated
`goalSearch && !cloudWorker`; the controller-path sites keep the bare predicate, where the tolerance
is what prevents the sibling and fan-in cascade. Found by adversarial re-verification, not by a test.

## Hunter record-lookup corrections (#686)

With records no longer inlined, the hunter has to locate them, and the first version of those instructions was wrong for all 88 goals in at least one respect: threat goals have no `class_id` and no `selected_record` (they carry `class_ids`, plural) so 11 hunters were told to read absent fields; the named manifest's `path` is database-relative while only `artifact_path` resolves, an ENOENT for all 77 class goals; the threat-entry `*_ids` → top-level-array dereference the inlined payload had pre-done was not stated; and 18 coverage-gap goals lost their only description, now recovered by pointing them at the threat model's `coverage_gaps` section.

## Prompt size (review feedback, addressed in bcc5e849)

*"The prompt edits are too large, are we sure we need all this prose?"* — correct, and the cost was larger than the diff showed. Since #652 the topology runtime context is part of the mandatory prefix prepended to **every** task prompt, not just goal nodes, so the deadline lines cost +1,000 bytes on all 179 compiled tasks. Measured, bytes:

| | baseline | as first written | now | trim recovered | net vs baseline |
| --- | --- | --- | --- | --- | --- |
| mandatory prefix (**per task**) | 1,233 | 2,233 | **1,478** | −755 | +245 |
| `strategies/goal-hunter.mdx` (**per goal node**) | 1,634 | 6,752 | **3,775** | −2,977 | +2,141 |
| `strategies/roaming-goal.md` | 1,288 | 2,064 | **1,782** | −282 | +494 |
| `setup/goal-plan.md` | 8,798 | 9,879 | **9,192** | −687 | +394 |
| `review/final-report.md` | 37,674 | 43,764 | **40,333** | −3,431 | +2,659 |
| `review/dedupe-findings.md` | 13,327 | 15,045 | **14,088** | −957 | +761 |

Weighted by where each file is paid — prefix × 179 tasks, hunter × 88 goal nodes, the rest once — the trim recovers **~402,000 bytes / ~100,000 tokens per run**. The prose had been giving back ~47% of this PR's own ~342,000-token win; it now gives back ~17%. Per goal node, everything read before the agent reaches the target (prefix + hunter prompt + median rendered goal statement) goes 16,891 → 9,281 → **5,549** bytes.

Cut: rationale rather than instruction, and every second statement of a rule already stated once — why records are no longer inlined, why the deadline is a recipe rather than a timestamp, why the census exists, the `date -u -d` ISO-formatting invocations, and the "Write, in this order" layout list that `final-report-markdown.ts` already emits. The runtime-context reasoning moved to the JSDoc on `topologyRuntimeContextForTimeout`, where it costs no tokens.

Kept: the corrected record-lookup facts (#686); that an empty findings result is a valid negative outcome and must not be replaced by an invented finding; the mechanism by which the agent measures its own deadline; and the census naming plus "state coverage rather than imply it" (#684). All 33 `semantic-anchors.test.ts` assertions still pass; none was deleted, and the one repin (`date -u` → `date -u +%s`) fixed a factually wrong instruction — the block states relative seconds and a start epoch, so `date -u` printed a string that could not be compared to `START+N`.

## Validation

- runtime **667 tests / 649 pass / 0 fail / 18 skipped**; cli 57, prompts 68, artifacts 139, topology 58, config 41 — all pass
- `prettier --check`, `eslint`, `runtime typecheck`, `bun build` of the workflow template, `validate-audit-profile-package.mjs`: all clean
- The #683 regression test is behavioural rather than a source grep: it transpiles 13 real slices of the workflow template over a real filesystem with the real artifact-contract collaborators, plants a stale marker, and asserts it is removed, nothing is published, and the lane contributes no provenance even when a contract-valid `findings.json` is later written into its canonical directory. **Mutation-proven against three separate ways to reintroduce the defect**, plus a companion test pinning the mandatory-evidence call shape that no behavioural test can reach.

<h3>Greptile Summary</h3>

The PR reduces dynamic goal prompts by replacing inlined records with bounded labels and artifact-path lookups, while preventing stopped goal lanes from being attested as valid empty searches.
- Adds durable per-lane goal-search coverage and renders it in runtime- and CLI-generated reports.
- Requires a durable agent output row before artifact verification and publication.
- Preserves cloud-worker retry behavior while allowing failed local/controller goal lanes not to cascade.
- Updates goal-plan contracts, hunter instructions, topology settings, and regression coverage.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| packages/runtime/src/templates/smithers/workflows/workflow.tsx | Adds durable agent-return evidence, fail-closed verification, goal-lane failure tolerance, and controller-authored coverage census generation. |
| packages/runtime/src/final-report-markdown.ts | Adds deterministic goal-search coverage parsing and Markdown rendering, including explicit unknown and incomplete states. |
| packages/cli/src/report-artifacts.ts | Reconciles reports against the runtime-owned census and removes agent-authored coverage values. |
| packages/artifacts/src/goal-plan.ts | Bounds replacement labels, rejects serialized record values, and preserves selected-record path validation. |
| .ultrafuzz/prompts/strategies/goal-hunter.mdx | Directs hunters to resolve threat, class, and coverage-gap context from canonical artifact paths rather than inlined JSON. |
| packages/runtime/src/smithers.ts | Carries the changed workflow behavior through Smithers compilation and generated workflow assertions. |
| packages/runtime/test/generated-workflow-verifier.test.ts | Adds broad behavioral regression coverage for stopped lanes, stale markers, publication refusal, and census generation. |

</details>

```mermaid
sequenceDiagram
  participant Planner as Goal planner
  participant Hunter as Goal hunter
  participant Smithers as Smithers runtime
  participant Verifier as Artifact verifier
  participant Census as Coverage census
  participant Report as Final report

  Planner->>Hunter: Short labels and record paths
  Hunter->>Hunter: Read full threat/class records
  alt Hunter returns durably
    Hunter->>Smithers: Durable output row
    Smithers->>Verifier: "agentReturned = true"
    Verifier->>Verifier: Validate and publish artifacts
    Verifier->>Census: Completed or completed-no-findings
  else Hunter stops early
    Smithers->>Verifier: "agentReturned = false"
    Verifier-->>Verifier: Refuse verification and publication
    Verifier->>Census: Stopped early
  end
  Census->>Report: Targeted and roaming coverage
  Report->>Report: State incomplete or unknown coverage explicitly
```
</details>

<sub>Reviews (4): Last reviewed commit: ["fix(prompts): trim the prose these fixes..."](https://github.com/monad-developers/ultrafuzz/commit/bcc5e8497eeff87454b93b751ccb29a0425475c1) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=54921284)</sub>

<details><summary><h4>Context used (5)</h4></summary>

- Knowledge Base — [CLI (`@ultrafuzz/cli`)](https://app.greptile.com/monad-foudnation/-/custom-context/knowledge-base/monad-developers/ultrafuzz/-/docs/cli.md)
- Knowledge Base — [Topology](https://app.greptile.com/monad-foudnation/-/custom-context/knowledge-base/monad-developers/ultrafuzz/-/docs/topology.md)
</details>


aviggiano (commented):
The prompt edits are too large, are we sure we need all this prose?

aviggiano:
> The prompt edits are too large, are we sure we need all this prose?

You were right, and the cost was bigger than the diff showed. Trimmed in bcc5e849.

**The worst case was not in the files you were looking at.** Since #652 the topology runtime
context is part of the mandatory prefix prepended to *every* task prompt, not just goal nodes.
The deadline lines I added took it from 1,233 to 2,233 bytes — **+1,000 bytes on all 179 compiled
tasks** of a real plan. `packages/prompts/test/agent-preamble.test.ts` pins that length, and the
honest fix was to shrink the text rather than raise the assertion.

### Measured, bytes

| | baseline | as I first wrote it | now | trim recovered | net vs baseline |
| --- | --- | --- | --- | --- | --- |
| mandatory prefix (**per task**) | 1,233 | 2,233 | **1,478** | −755 | +245 |
| `strategies/goal-hunter.mdx` (**per goal node**) | 1,634 | 6,752 | **3,775** | −2,977 | +2,141 |
| `strategies/roaming-goal.md` | 1,288 | 2,064 | **1,782** | −282 | +494 |
| `setup/goal-plan.md` | 8,798 | 9,879 | **9,192** | −687 | +394 |
| `review/final-report.md` | 37,674 | 43,764 | **40,333** | −3,431 | +2,659 |
| `review/dedupe-findings.md` | 13,327 | 15,045 | **14,088** | −957 | +761 |

Baseline is `origin/release/v0.1.0`, which for all six is byte-identical to 10ced3ec. The prefix
figures are rendered at 1800/300/1500 and computed the way the test computes them, not estimated.

Weighted by where each file is actually paid — prefix × 179 tasks, hunter × 88 goal nodes, the
rest once each — the trim recovers **~402,000 bytes, ~100,000 tokens per run**. Against this PR's
own ~342,000-token win, the prose was giving back **~47%**; it now gives back **~17%**.

Per goal node, everything the agent reads before it reaches the target (prefix + hunter prompt +
rendered goal statement at the measured median):

| | bytes |
| --- | --- |
| before this PR | 16,891 |
| this PR, as first written | 9,281 |
| this PR, now | **5,549** |

### What I cut

Rationale, not instruction — and every second statement of a rule already stated once:

- why records are no longer inlined; why the deadline is a recipe rather than a timestamp; why the
  census exists at all. All of it explains the design to a human reader, and the PR description and
  the commit messages already carry it. The runtime-context reasoning moved into the JSDoc on
  `topologyRuntimeContextForTimeout`, where it costs nothing.
- the `date -u -d "+N seconds" +%Y-%m-%dT%H:%M:%SZ` invocations. The arithmetic over `date -u +%s`
  is the actionable part; the ISO formatting was one of several ways to render it.
- in `final-report.md`, the four-item "Write, in this order" layout list —
  `final-report-markdown.ts` already emits exactly that layout — plus the per-goal field-by-field
  census walkthrough and the `ultrafuzz report` regeneration aside.
- in `goal-plan.md`, the token-cost paragraph and the restatement of the three lookup paths, which
  belong in `goal-hunter.mdx` and are there.

### What I kept, and why

Four things, each of which is a bug already filed against this branch if it goes:

- the corrected record-lookup facts (#686): `class_ids` plural on threat goals, the
  `selected_record.path` versus database-relative manifest trap, the threat-entry `*_ids`
  dereference targets, the coverage-gap path. The first version of these instructions was wrong for
  all 88 goals in at least one respect.
- that an empty findings result is a valid negative outcome and must not be replaced by an invented
  finding. That is a safety property, not decoration.
- the mechanism by which the agent can measure its own deadline.
- naming the census, and requiring the report to state goal coverage rather than imply it (#684).

All 33 assertions in `semantic-anchors.test.ts` still pass and none was deleted.

### One correction while I was in there

Both strategy prompts claimed the Topology Runtime Context "states an absolute UTC deadline" and
told the agent to check `date -u` against it. Neither was true: the block states relative seconds
plus a start epoch, and `date -u` prints a string that cannot be compared to `START+N`. Both now
derive the deadline from `date -u +%s` and compare epochs, and the anchor is repinned to
`date -u +%s` — the same behaviour, pinned to the command that can actually observe it.

### Notes

- Your merge of `release/v0.1.0` into this branch (57685f20) and your lint fix (3962b226) are
  preserved — the trim is a fast-forward commit on top, not a force-push.
- Green on the pushed tree: prompts 68/68, runtime 667 / 649 pass / 0 fail / 18 skipped, cli 57,
  artifacts 139, topology 58, config 41; `prettier --check`, `eslint`, runtime `typecheck`,
  `bun build` of the workflow template, and `validate-audit-profile-package.mjs` all clean.

aviggiano:
Merged into `release/v0.1.0` as f1b8c22378152859de069baccdd79c102565c080 (merge commit, matching how #665 and #689 landed on this base).

**What landed**

- Goal statements move from inlined JSON records to path references — 47x smaller at the median, and `replacementsSchema` now caps replacements at 200 characters and rejects structurally-JSON values so a planner that ignores the prompt cannot reintroduce the wall. (#680)
- Goal lanes carry `continueOnFail`, so one killed lane no longer cascades into downstream artifact-contract failures. (#685 refs)
- `verifyArtifacts` requires `agentReturned`, sourced from Smithers' durable output row rather than from the worktree, so a lane that never ran can no longer be attested as a verified empty result. The refusal throws instead of returning a row, deliberately. (#683)
- Coverage is stated rather than implied: the census is named so the reporting and reviewing agents can read it, and the report must state goal coverage. (#688 cloud-worker retry also preserved.)
- Hunter record-lookup corrections: threat goals carry `class_ids` plural, the plan's `selected_record.path` is authoritative over the manifest's database-relative `path`, threat-entry `*_ids` dereference into the document's top-level arrays, and coverage-gap goals read the threat model's `coverage_gaps`. (#686, #684)

Closes #680, #683, #684, #686, #688. Refs #675, #685.

**Review feedback on prompt size was addressed before merge** (bcc5e849): the mandatory prefix is back from 2,233 to 1,478 bytes against a 1,233 baseline, and `strategies/goal-hunter.mdx` from 6,752 to 3,775 against 1,634. Weighted by where each file is paid, that recovers ~100,600 tokens per run; the prose now gives back ~17% of the PR's win instead of ~47%. Details and the full table are in the comment above and in the PR body.

`review/final-report.md` is still +2,659 bytes over baseline against a ~+1,500 target. That is deliberate: the remaining bulk is the four behaviours the semantic-anchor tests pin — including that an empty findings result is a valid negative outcome and must not be replaced by an invented finding. Getting to +1,500 means deleting one of those, which would reintroduce #684 or #686.

**Gates on the merged head (bcc5e849), all green:** release-gates (push) 17m51s pass, release-gates (pull_request) pass, Greptile pass, semgrep pass, Socket Security x2 pass. Local suite: runtime 667 tests / 649 pass / 0 fail / 18 skipped; cli 57, prompts 68 (semantic-anchors 33/33, agent-preamble 3/3), artifacts 139, topology 58, config 41.

Note: the head branch was auto-deleted by the repository's `delete_branch_on_merge` setting, as happened with #689. `bcc5e849` is preserved as the second parent of the merge commit, and the branch can be restored from this page if wanted.

aviggiano:
Post-merge verification of `release/v0.1.0` at `f1b8c223`: all five deliverables survived both hand resolutions, and `f1b8c223`'s tree is byte-identical to `bcc5e849` (`cfaeb47a8fc2e6a43bd72ad8e907e28af9d46e30`), so the merge commit contributed no unreviewed content. One shipped gap worth recording where the change landed.

**Goal-search coverage is not enforced in `report.md` on the healthy runtime path.**

The census is stamped into `report.json` only. `reconstructAuthoritativeGoalSearchCoverage` (`packages/runtime/src/templates/smithers/workflows/workflow.tsx:6785-6816`) rewrites the report JSON and never touches the Markdown. The renderer that emits `## Goal search coverage` runs only through `projectCanonicalFinalReport`, and on the runtime path that is reached only from `materializeMissingFinalReportArtifacts`, which returns early at `workflow.tsx:3415-3420` unless an output fails its contract. `report.md`'s contract is `ultrafuzz/nonempty-markdown@1`, i.e. `contents.trim().length > 0` (`packages/artifacts/src/artifact-contracts.ts:338-341`). So for a final-review attempt that produced any non-empty Markdown, the agent's own prose ships unchecked: nothing in `workflow.tsx` or the artifact gates calls `isDirectiveConformingFinalReportMarkdown`, whose only non-test callers are `packages/runtime/src/final-report-markdown.ts:119` and `packages/cli/src/report-artifacts.ts:83`.

Consequence: an agent-authored `report.md` can be more optimistic about goal coverage than the census stamped into its sibling `report.json`, and the two can contradict each other until someone runs `ultrafuzz report`. The only guards are prompt text (`.ultrafuzz/prompts/review/final-report.md`, "Never be more optimistic than the census") and `packages/prompts/test/semantic-anchors.test.ts`, which tests the prompt, not the output. The CLI path is sound: `reconcileGoalSearchCoverage` discards the model's value unconditionally (`report-artifacts.ts:413-424`).

Suggested fix: have the runtime verifier assert `isDirectiveConformingFinalReportMarkdown` against the stamped report before publishing `report.md`, so a non-conforming agent report is recovered by projection instead of shipped.

Second, smaller: the `## Goal search coverage` heading requirement rides on the `requireImplementationCoverage` parameter (`final-report-markdown.ts:143`), and `report-artifacts.ts:83` passes `false`. A current-run report reaching the preserved-historical branch could be preserved with no coverage section at all. Latent today, since both packaged goal-bearing topologies carry the v2 properties handoff that throws at `report-artifacts.ts:76`, but the coupling is worth splitting into its own flag.

For the record, two deliberate items that read like merge damage and are not: `timeout_seconds` 14400 -> 7200 in `full.yml`/`invariant-only.yml`/`.ultrafuzz/topology.yml` is #690 reverting #689's raise, test-pinned at `packages/topology/test/packaged-topologies.test.ts:149`; and `fuzz-only.yml` is absent because release deleted it in `7a064638` (#651). #689's retry budget (`max_attempts: 2`), worktree base, and preparation-diagnostics node are all present.

PR #709, opened 2026-08-21
fix(runtime): honor conditional skips under Bun

### LINKED ISSUE #705: Bun's node:test shim silently ignores the skip option, so option-guarded tests run anyway
## Problem

Bun's `node:test` shim **silently ignores the `skip` option**. A test written as:

```ts
test("name", { skip: someCondition }, () => { ... });
```

runs under `bun test` regardless of the condition, and fails if its body assumes the skip held. Verified with a two-line probe: `test("x", { skip: true }, () => assert.equal(1, 2))` executes and fails under bun, while `test.skip("x", ...)` is correctly reported as skipped.

Under `node --test` the same code skips as intended, so the divergence only appears in the bun lane. That lane exists specifically to exercise the generated adapter contracts, which is where environment-dependent tests concentrate — a test guarded on "is this CLI installed" is exactly the shape that hits this.

The failure mode is worse than a plain break: a contributor writes an option-based guard, sees it skip locally under `node --test`, and the bun lane then either fails on a machine lacking the dependency or — if the body happens to tolerate it — passes while asserting nothing.

Two related observations from the same investigation, both confirmed:

- Bun's shim **does** honour `timeout` in that same options object, so the object is not ignored wholesale — only `skip` is dropped. That asymmetry makes the bug hard to spot by inspection.
- Bun applies a **5s default timeout** when none is given. A contract test doing project initialization plus a transpile exceeded it on a cold cache and passed on every subsequent run, which reads as a flake rather than a missing timeout.

## Proposal

1. Use `test.skip(...)` (or a `const t = cond ? test : test.skip` selector) for anything that must actually skip. Do not rely on the `skip` option in any test that the bun lane runs.
2. Add a check that fails if `{ skip:` appears in a test file the bun lane executes, so this cannot regress silently.
3. Give bun-lane tests that do real filesystem or process work an explicit `timeout`, since the 5s default is well below what project initialization costs on a cold cache.
4. Document both behaviours where the bun lane is described, so the next person does not rediscover them from a red CI run.

## Acceptance criteria

- [ ] No test executed by the bun lane relies on the `skip` option.
- [ ] A check fails on reintroduction.
- [ ] Bun-lane tests doing init or transpile work declare an explicit timeout.

aviggiano:
## Summary

- register conditional runtime tests through a shared selector that uses `test.skip`, which Bun's `node:test` shim honors
- give every generated-adapter contract a stable Bun-lane prefix and an explicit cold-cache timeout
- add source and manifest regressions, an executable skipped-contract sentinel, and contributor documentation

## Validation

- runtime test TypeScript compilation
- runtime shard/regression tests: 4/4
- converted Modal-style conditional test: 1/1
- Bun adapter lane: 36 pass, 199 filtered, 0 fail
- Bun skipped-contract sentinel: 1 skip, 0 fail
- touched-file ESLint and Prettier
- docs checks and `git diff --check`

The broad release-supporting command reached 346 passing supporting tests without an assertion failure before the host terminated it; the affected regression and complete Bun phases were then rerun separately as listed above.

Closes #705

<h3>Greptile Summary</h3>

The PR makes conditional runtime contracts use Bun-compatible `test.skip` registration and gives generated-adapter tests stable lane names and cold-cache timeouts. It also introduces audit-profile retry budgets and maintains compatibility with previously serialized resolved configurations.
- Adds shared conditional and prefixed test registration for the Bun adapter lane.
- Updates runtime test scripts and regression coverage for shard skipping and exact Bun test selection.
- Adds profile-specific retry budgets, removes topology-level retry shadowing, and documents precedence.
- Extends resolved-config schemas and fixtures for retry provenance while retaining historical v3 compatibility.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| packages/runtime/test/runtime-test-shard.ts | Adds explicit `test.skip` registration and a shared conditional selector while preserving supported callback and options overloads. |
| packages/runtime/test/runtime.test.ts | Migrates conditional adapter contracts to the shared selector and assigns stable Bun-lane names and timeouts. |
| packages/runtime/scripts/run-pr-smoke-tests.mjs | Updates Bun smoke selection to use exact prefixed contract names and verifies each expected test passes. |
| packages/config/src/resolve.ts | Applies and records profile retry budgets while preserving project and runtime override precedence. |
| packages/config/src/resolved-config-schema.ts | Makes setting-origin validation partial so historical resolved configurations remain accepted after adding retry provenance. |
| packages/config/audit-profiles.yml | Defines one-attempt low-cost profiles and a five-attempt exhaustive profile while removing the thorough preset. |
| packages/config/topologies/smoke.yml | Removes node and group retry values that previously shadowed the smoke profile’s retry budget. |

</details>

```mermaid
flowchart LR
  Profile[Audit profile retry setting] --> Resolve[Resolve configuration]
  Project[Project retry override] --> Resolve
  Runtime[Runtime override] --> Resolve
  Resolve --> Attempts[Effective same-agent attempts]
  Attempts --> Expand[Expand topology]
  Expand --> Workflow[Compiled workflow retry policy]

  Condition[Runtime contract condition] --> Selector[testWhen]
  Selector -->|true| Shard[Shard-aware test registration]
  Selector -->|false| Skip[test.skip]
  Shard --> Prefix[Bun adapter contract prefix]
  Skip --> Prefix
  Prefix --> Bun[Bun adapter test lane]
```
</details>

<sub>Reviews (3): Last reviewed commit: ["test(runtime): preserve Bun smoke select..."](https://github.com/monad-developers/ultrafuzz/commit/30ff5fd1ad0dfe0d6f62fe9cec63356669aa1df5) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=55400572)</sub>


aviggiano (commented):
One shard-composition edge remains in `packages/runtime/test/runtime-test-shard.ts:63`: `testWhen(true)` delegates to `test`, but that wrapper’s off-shard branches at lines 47 and 50 still register `nodeTest(..., { skip: reason }, fn)`. With Bun 1.3.0 and `ULTRAFUZZ_RUNTIME_TEST_SHARD=1/4`, I reproduced an excluded `testWhen(true)` body executing and failing; the same probe confirmed `testWhen(false)` skips and caller timeout options override defaults. Current checked-in CI does not combine Bun with the shard env, but the package scripts do not reject or clear that composition. Please make off-shard registration use `nodeTest.skip` (the option can remain only to preserve the reason), or explicitly reject/clear Bun sharding.

aviggiano (commented):
Independent re-review of exact head `98fc5ee201201f429865cfbb01c0fda9c4aa8cbd` against `fe0922ea9b5792905cbef921d73965d6ab20130c`: **no findings**.

The prior off-shard blocker is resolved. Both callback-only and options-plus-callback off-shard paths now register through `nodeTest.skip`, while preserving the shard reason and caller options. The new Node-driven real Bun subprocess sets `ULTRAFUZZ_RUNTIME_TEST_SHARD` to a shard excluding both sentinels; each body throws if invoked, and the test requires Bun exit 0, `2 skip`, and absence of either sentinel failure.

Local validation:
- `pnpm --filter @ultrafuzz/runtime exec tsc -p tsconfig.test.json --pretty false` — pass
- `node --test --test-reporter=tap dist-test/test/runtime-test-shard.test.js` — 5/5 pass, including the Bun 1.3.0 subprocess
- focused ESLint — pass
- focused Prettier check — pass
- `git diff --check fe0922ea...HEAD` — pass

GitHub checks at this head are clean: PR build and Node.js 24 runtime smoke, release gates, Greptile, Semgrep, and both Socket checks pass; full release validation is expectedly skipped for the pull-request event. I found no correctness, regression, test-gap, or scope issues in the full diff.

aviggiano:
@aviggiano please review

aviggiano:
Ready to merge

Issue #716, opened 2026-08-21
Cloud handoff rejects committed files matched by gitignore


aviggiano:
## Summary

Modal cloud handoff creation in v0.1.0 rejects an otherwise clean Git repository when a committed file is also matched by the repository's `.gitignore`.

The failure occurs before Modal resource creation:

```text
cloud node project contents changed while the handoff archive was built
```

## Generic reproduction

1. Create a temporary Git repository with `source.txt`.
2. Add `.gitignore` containing `tracked-ignored.txt`.
3. Create `tracked-ignored.txt`, force-add it, and commit both files.
4. Initialize a stock UltraFuzz project, configure Modal cloud execution, and launch the first node.

This can be reduced to the existing `createProjectFixture()` in `packages/modal/test/node-provider.test.ts`: commit an ignored file with `git add --force`, then expect `createModalNodeHandoffArchive(fixture.root, fixture.input)` to resolve. On v0.1.0 it rejects with the error above.

## Cause

`createModalNodeHandoffArchive()` fingerprints the original `HEAD^{tree}`, extracts `git archive HEAD` into a staging directory, initializes a new repository, and runs:

```ts
git add -A
```

In that newly initialized repository, Git applies the extracted `.gitignore` and omits the committed ignored path. The reconstructed tree therefore differs from the original committed tree even though the source never changed.

## Expected behavior

The cloud handoff should preserve the exact committed Git tree, including committed paths that match ignore rules, while continuing to exclude undeclared untracked files and reject unsafe filesystem entries.

## Upstream status checked

No exact issue or PR was found by searching for the terminal error, tracked ignored files, and cloud handoff reconstruction. PR #655 (merged as `be00602f370852efb6f2ffe87df268db15e09d02`) replaced the baseline construction on current `main` for source-revision binding and appears compatible, but #650 tracks a different source-selection defect and I did not find a focused regression covering committed ignored paths.

## Acceptance criteria

- Add a synthetic regression with a committed ignored file.
- Handoff creation succeeds and the archive contains that file.
- Untracked ignored files remain excluded.
- Existing handoff safety and semantic-fingerprint tests continue to pass.
- Establish whether PR #655 completely fixed this case; avoid a redundant implementation if it did.

aviggiano:
Investigated on current `main` (`60d6bd0a`).

[...]

PR #717, opened 2026-08-21
fix(modal): keep committed ignored paths in the cloud handoff baseline

aviggiano:
## Summary

- force-add the cloud handoff staging tree so the reconstructed baseline keeps committed paths matched by the project's `.gitignore`
- add a synthetic regression covering both handoff baselines: a committed ignored file stays tracked, an untracked ignored file stays excluded

## Cause

`createDeterministicGitBaseline()` extracts `git archive HEAD` into a staging directory, runs `git init`, then `git add --all`. The extracted tree carries the project's `.gitignore`, so the fresh repository applied those rules and omitted committed paths matching them. The reconstructed tree therefore differed from the committed tree it was built from, even though the source never changed.

Forcing the add is safe and exact here: at that point the staging tree is exactly the extracted committed tree, and untracked project files never reach staging, so nothing undeclared can be admitted. The run root, execution snapshot, and artifact directories are all created after the baseline commit and remain untracked as before.

## Relationship to #655

PR #655 replaced the baseline used when a run records `source_revision`/`source_ref`: `createRecordedGitBaseline()` fetches the recorded commit and hard-resets, which already preserves committed ignored paths. The `git archive` + `git add --all` fallback baseline was left in place and still dropped them. The new regression asserts both baselines, so the recorded path from #655 is now covered too.

Evidence that the gap remained on `main` — the new test run against unmodified `main` source:

```
FAIL  test/node-provider.test.ts > preserves committed paths matched by gitignore and still excludes untracked ignored files
AssertionError: expected '.gitignore\nsource.txt\n' to be '.gitignore\nsource.txt\ntracked-ignor…'
- tracked-ignored.txt
```

## Validation

- `npx vitest run test/node-provider.test.ts` in `packages/modal` — 86/86 passed (85 before this change plus the new regression)
- new regression fails on unmodified `main` and passes with the fix
- `npx prettier --check` on both changed files — clean
- `npx eslint` on both changed files — clean
- `npx tsc -p tsconfig.json --noEmit` in `packages/modal` — clean

Refs #716

<h3>Greptile Summary</h3>

The PR fixes deterministic cloud handoff baselines so committed files matched by `.gitignore` remain tracked after reconstruction.
- Force-adds the freshly extracted committed tree when creating the deterministic Git baseline.
- Adds regression coverage for deterministic and recorded baselines, including exclusion of untracked ignored files.

The PR appears safe to merge, with the forced add constrained to a temporary tree populated exclusively from the committed source archive.

The changed command restores committed ignored paths without admitting local untracked files, and the regression exercises both deterministic and recorded handoff baselines.

| Filename | Overview |
|----------|----------|
| packages/modal/src/node-provider.ts | Force-adds the archive-derived staging tree so ignore rules cannot remove committed paths from the reconstructed baseline. |
| packages/modal/test/node-provider.test.ts | Adds synthetic coverage proving committed ignored files remain tracked while untracked ignored files remain excluded in both baseline modes. |

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(modal): keep committed ignored paths..."](https://github.com/monad-developers/ultrafuzz/commit/86e8a988a908d66ee7faa781927abaf2bc3a94cc) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=55615934)</sub>


aviggiano:
can merge

PR #798, opened 2026-08-24
fix(runtime): treat the workflow engine database as controller-owned

### LINKED ISSUE #797: Workflow engine database in the target root dirties the governed worktree and blocks later private runs
## Summary

The workflow engine opens its SQLite database in the target repository root, creating `smithers.db`, `smithers.db-shm`, and `smithers.db-wal` there. Those paths are not listed in the governance `controllerOwnedPaths`, so they count as untracked changes in the governed worktree.

The consequence is self-inflicted: the first run that launches far enough to open the database dirties the target identity, and every subsequent private-target run then fails before planning with:

```
error: DATA_GOVERNANCE_PRIVATE_TARGET_UNBOUND: private campaigns require a clean, exact Git target identity
```

Because the target identity feeds `input_digest`, any previously recorded disclosure acknowledgement also stops matching and has to be regenerated:

```
error: DATA_DISCLOSURE_ACKNOWLEDGEMENT_REQUIRED: private disclosure requires acknowledgements for model:<destination>
```

So a private campaign is launchable at most once per manual cleanup, and the second attempt fails for a reason unrelated to anything the operator changed.

## Where

`packages/runtime/src/plan-run.ts` passes:

```js
controllerOwnedPaths: [
  runRoot,
  path.join(projectRoot, ".ultrafuzz", "runs"),
  path.join(projectRoot, ".smithers", "node_modules"),
  path.join(projectRoot, ".smithers", "workflows")
]
```

`targetIdentity()` in `packages/runtime/src/data-governance.ts` filters untracked paths against that list, so anything Ultrafuzz or the engine writes into the target outside those four entries makes `dirty` true.

## Reproduction

1. `ultrafuzz init` and configure a private target so `sensitivity` resolves to `private`.
2. Launch a run and let it reach workflow submission. `smithers.db*` appear in the project root.
3. Launch again. It fails `DATA_GOVERNANCE_PRIVATE_TARGET_UNBOUND` even though the operator changed nothing.

Timing confirms the origin: the database files are created during the submit window of the preceding run.

## Proposed fix

Add the three engine database paths to `controllerOwnedPaths`. They are engine runtime state, not reviewable project content, so excluding them from the target identity is consistent with the existing entries.

A cleaner alternative would be to open the engine database under the run directory instead of the target root, so the governed worktree is never written to at all. That is a larger change to the runner invocation, and it would also stop the database from polluting the operator's repository, so it may be the better long-term fix.

## Note on scope

This issue is deliberately limited to engine runtime state. `ultrafuzz.toml`, `.ultrafuzz/topology.yml`, and `.ultrafuzz/prompts/**` also make a freshly initialised target dirty, but those are operator-reviewable content that the trust model intends to bind into the governance input, so committing them (or ignoring them locally) looks like the intended workflow rather than a defect. Flagging it only because a first-time private-target user hits `PRIVATE_TARGET_UNBOUND` immediately after `init`, with no guidance pointing at that resolution.

aviggiano:
Closes #797

## Problem

The workflow engine opens its SQLite database in the target repository root, so `smithers.db`, `smithers.db-shm`, and `smithers.db-wal` land in the governed worktree as soon as a run reaches submission. `plan-run.ts` did not list them in `controllerOwnedPaths`, so `targetIdentity()` counted them as untracked changes and set `dirty`.

The result is self-inflicted and hard to read from the operator's side. The first run that launches dirties the target identity; every later private-target run then fails before planning:

```
error: DATA_GOVERNANCE_PRIVATE_TARGET_UNBOUND: private campaigns require a clean, exact Git target identity
```

And because the target identity feeds `input_digest`, a previously recorded disclosure acknowledgement stops matching as well, producing a second governance error for something the operator never touched. In practice a private campaign is launchable once per manual cleanup.

## Change

Add the three engine database paths to `controllerOwnedPaths`, with a comment explaining why they are there. They are engine runtime state rather than reviewable project content, which is consistent with the existing entries (`.ultrafuzz/runs`, `.smithers/node_modules`, `.smithers/workflows`).

## Verification

Observed directly: the database files are created during the submit window of one run, and the next run with an unchanged working tree fails `PRIVATE_TARGET_UNBOUND`. With this change the same sequence plans and launches without manual cleanup between runs.

## Alternative worth considering

Opening the engine database under the run directory instead of the target root would keep the governed worktree untouched entirely, and would also stop the database from polluting the operator's repository. That is a larger change to the runner invocation, so it is not attempted here, but it is probably the better long-term shape.

## Scope note

Deliberately limited to engine runtime state. `ultrafuzz.toml`, `.ultrafuzz/topology.yml`, and `.ultrafuzz/prompts/**` also make a freshly initialised target dirty, but those are operator-reviewable content the trust model intends to bind into the governance input, so committing or locally ignoring them looks like the intended workflow rather than a defect. Raised as a documentation/UX point in the issue instead.

---

## Update: the list was duplicated, which is the more important half

The first commit here added the engine database paths to `plan-run.ts` only. That was incomplete, and testing it surfaced the underlying defect: the controller-owned path list existed in **two** places.

- `plan-run.ts` built it for the planning gate.
- `start-run.ts` rebuilt the same four entries independently for the submit-time re-verification in `assertCurrentDataGovernanceTarget()`.

With the entries added to only the first, planning saw a clean target and sealed that identity, then submission recomputed a dirty one and failed:

```
error: WORKFLOW_SUBMISSION_FAILED: campaign source changed after its data-disclosure acknowledgement
```

That message blames the operator's source for what is actually a list mismatch inside the runtime, and it is reachable by any future edit that touches one call site and not the other.

The second commit extracts `controllerOwnedGovernancePaths()` next to `targetIdentity()` and uses it from both sites, so the planning gate and the submit-time check cannot disagree. The engine-database entries live in that one function.

<h3>Greptile Summary</h3>

The PR fixes the previously reported planning/submission identity divergence by centralizing controller-owned governance paths and using the shared list at both boundaries.
- Adds Smithers SQLite database, WAL, and shared-memory files to the controller-owned exclusions.
- Reuses one helper during planning and submit-time target re-verification.

The PR appears safe to merge.

No blocking failure remains; planning and submit-time verification now derive target identity from the same controller-owned path list.

| Filename | Overview |
|----------|----------|
| packages/runtime/src/data-governance.ts | Centralizes controller-owned path exclusions and includes all three Smithers SQLite runtime files. |
| packages/runtime/src/plan-run.ts | Replaces the planning-specific exclusion list with the shared governance-path helper. |
| packages/runtime/src/start-run.ts | Uses the same shared exclusions for submit-time target identity verification, resolving the prior divergence. |

</details>

```mermaid
flowchart LR
    A[Plan run] --> B[Shared controller-owned paths]
    B --> C[Planning target identity]
    B --> D[Submit-time target identity]
    C --> E[Persisted governance provenance]
    E --> F{Identity matches at submission}
    D --> F
    F --> G[Submit Smithers workflow]
```
</details>

<sub>Reviews (2): Last reviewed commit: ["fix(runtime): derive controller-owned go..."](https://github.com/monad-developers/ultrafuzz/commit/67204470c12072246ed2d438b08c3f26a88e1945) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=56251575)</sub>

**Context used:**


aviggiano (commented):
LGTM

PR #820, opened 2026-08-24
Scan artifact publications for positively identified secrets only

### LINKED ISSUE #819: Artifact secret gate fails closed on ordinary prose: entropy, 40-hex and keyword heuristics cannot separate credentials from identifiers
## Summary

`assertArtifactPublicationsContainNoSecrets` fails closed on ordinary technical prose. Its speculative heuristics — high-entropy candidates, unlabeled 40-hex, and keyword-assignment — cannot distinguish a credential from a long identifier, a git commit hash, or an English sentence about tokens.

Because the gate deliberately does not transform bytes ("contaminated output must be regenerated explicitly"), a hit is fatal: the verify node fails, and with no retry on that path the whole run dies after the agent work is already paid for.

In one campaign this killed **five consecutive runs**, each on a different innocent string, each fix revealing the next shape. The failures are not concentrated in one rule — three separate heuristics misfired.

## Reproduction

All five reproduce on `main` (`8caf159f`). None contains a secret.

```js
import { containsSensitiveSecrets } from "@ultrafuzz/security";

// A — high-entropy candidate: qualified method group cited in prose
containsSensitiveSecrets("IExampleVaultCore.setAuthority/togglePause/transferOwnership"); // true

// B — high-entropy candidate: a long generated test name
containsSensitiveSecrets("testFuzz_MaximumPrincipalForGrossIsSafeAndBounded"); // true

// C — high-entropy candidate: qualified enum member
containsSensitiveSecrets("IExampleBook.NativeExecInstruction.POST_ONLY"); // true

// D — keyword-assignment rule on an English sentence
containsSensitiveSecrets("For every token:\n  contract token balance >= sum of credits"); // true

// E — unlabeled 40-hex rule on pinned dependency commits
containsSensitiveSecrets(
  "Pinned dependencies: `da0e9c1b9c2f0d3e4a5b6c7d8e9f0a1b2c3d4e5f`, `1a2b3c4d5e6f708192a3b4c5d6e7f8091a2b3c4d`."
); // true
```

A, B and C are `HIGH_ENTROPY_CANDIDATE_PATTERN` plus `isHighEntropySecretCandidate` (`sensitive-redaction.ts:33`, decision at line 318). D is `SENSITIVE_ASSIGNMENT_PATTERN`. E is `redactUnlabeledFortyHexSecrets`.

## Why the entropy rule cannot be tuned

Shannon entropy does not separate the two populations. Measured:

| Real credentials | | Long identifiers |  |
|---|---|---|---|
| 32-byte hex key | 3.90 | versioned vault method | 4.62 |
| GitHub PAT | 4.25 | versioned proxy method | 4.45 |
| OpenRouter key | 4.31 | versioned book method | 4.41 |
| Anthropic key | 4.48 | qualified method group | 4.31 |
| base64 blob | 5.19 | | |
| Google API key | 5.27 | | |
| JWT | 5.37 | | |

The distributions **overlap completely** — a GitHub PAT (4.25) and a 32-byte hex key (3.90) score *below* the worst identifier (4.62). No threshold separates them. The current cutoff is 4.3, so identifiers at 4.31 fail while 4.25 passes; strings that read identically to a human land on opposite sides.

Long camelCase identifiers score high for the same reason credentials do: many distinct characters, no spaces. Entropy is measuring the wrong thing here.

## The rule is also not what catches real credentials

The GitHub PAT above scores 4.25, *below* the 4.3 threshold, and is still detected — because the vendor patterns (`sk-`, `sk-ant-`, `ghp_`, `github_pat_`, `AKIA`, `AIza`, `xoxb-`, `npm_`, JWT, `BEGIN PRIVATE KEY`) run before the entropy pass. Exact configured credential values are matched byte-for-byte separately and never depend on entropy either.

So for every credential format the codebase actually names, the speculative heuristics contribute nothing. Their only unique coverage is an opaque secret in no recognized format that is also not a configured value.

## Proposed fix

Scope the speculative heuristics out of **artifact publication only**, keeping every positive identification:

- keep: exact forbidden values, all vendor patterns, BIP39 mnemonics, context-labeled private keys, URL credentials, `Bearer` tokens
- skip for artifact publication: high-entropy candidates, unlabeled 40-hex, keyword-assignment

Other callers keep full behavior. They redact for display rather than failing closed (`attempt-ledger.ts`, `state.ts`, `config/redaction.ts`), or publish genuinely outward (`modal/public-bundle.ts`), so over-redaction there is cheap and appropriate.

### Residual exposure, stated plainly

An opaque secret in no recognized format, which is also not a configured credential value, would no longer be caught in a published artifact. That is a real reduction, not a free win. It is a good trade because run artifacts are local run output, the campaign's own credentials remain byte-matched, and the current behavior destroys entire runs on false positives at a rate of five in five.

## Notes

There is no environment variable or config option to tune or disable this gate; it is unconditional.

Narrower alternatives were tried first — teaching the exemption successive shapes of source notation (dotted chains, single tokens, `/` and `::` separators) and applying `MIN_EXACT_SECRET_VALUE_LENGTH` to the keyword rule. Each unblocked one run and the next run failed on a new shape. Case D above still trips even with the length floor applied, because in Markdown the captured value is the following word rather than an escape.

aviggiano:
Closes #819.

## What changes

`assertArtifactPublicationsContainNoSecrets` now scans in `"positive-only"` mode.

| | kept for publication | skipped for publication |
|---|---|---|
| exact configured credential values (byte-for-byte) | ✅ | |
| vendor patterns (`sk-`, `sk-ant-`, `ghp_`, `github_pat_`, `glpat-`, `hf_`, `xox*-`, `AKIA`/`ASIA`, `AIza`, `npm_`, `ya29.`, JWT, PEM, alchemy/infura URLs) | ✅ | |
| BIP39 mnemonics, context-labeled private keys, URL credentials, `Bearer` tokens | ✅ | |
| high-entropy candidates | | ⛔ |
| unlabeled 40-hex | | ⛔ |
| key-name assignment (`token:` / `secret =` / …) | | ⛔ |

`SecretScanMode` defaults to `"all"`, so **every existing caller is unchanged**. Only the publication gate opts into `"positive-only"`.

The other callers are left alone deliberately — `attempt-ledger.ts`, `state.ts` and `config/redaction.ts` rewrite text for display rather than failing, and `modal/public-bundle.ts` publishes genuinely outward. Over-redaction is cheap there. It is not cheap in the gate, which fails closed and refuses to rewrite bytes, so a false positive destroys a run.

## Why these three rules cannot stay

Shannon entropy does not separate the populations:

| Real credentials | | Long identifiers | |
|---|---|---|---|
| 32-byte hex key | 3.90 | versioned vault method | 4.62 |
| GitHub PAT | 4.25 | versioned proxy method | 4.45 |
| OpenRouter key | 4.31 | versioned book method | 4.41 |
| Anthropic key | 4.48 | qualified method group | 4.31 |
| base64 blob / Google key / JWT | 5.19–5.37 | | |

The cutoff is 4.3. A GitHub PAT (4.25) and a 32-byte hex key (3.90) score *below* the worst identifier (4.62) — there is no threshold that separates them, so this is not tunable.

It also is not what catches real credentials: that GitHub PAT scores below the cutoff and is still detected, because the vendor patterns run first. For every format the codebase can name, the speculative rules contribute nothing.

The other two are plainly miscalibrated for technical prose: unlabeled 40-hex matches every git commit hash, and the key-name rule fires on `For every token:`.

## Residual exposure

An opaque secret in no recognized format, which is also not a configured credential value, is no longer caught in a published artifact. **That is a real reduction, not a free win.**

The existing fixture asserting exactly that case is *updated rather than deleted*, so the tradeoff is visible in the test suite instead of implied. It is a good trade because artifacts are local run output, configured credentials remain byte-matched, and the current behavior destroyed five consecutive runs on false positives.

## Second commit — a flaky test that was reporting a real bug

`a runtime document read rejects an atomic path replacement during its snapshot` fails ~50% of the time on `main`, in isolation as well as in suite. It is not slow, it is correct only when it wins a race.

`readRegularFileSnapshot` compares `fstat` on the open descriptor before and after reading. A rename over the path leaves that descriptor on the **original inode**, so `dev`, `ino` and `size` are unchanged and the only remaining evidence is the ctime bump from the unlink — which the kernel records at timestamp granularity. Open/read/rename/fstat routinely complete inside one tick on NVMe, and the replacement then goes undetected: the read returns content for a path that no longer refers to that inode.

Comparing `nlink` fixes it exactly — the rename drops the original inode's link count to zero, independent of the clock. 10/10 deterministic in isolation afterwards.

`VALIDATOR_BUILD_IDENTITY` covers the compiled validator modules, so touching `schema-registry.ts` moves it by design; both pinned copies are updated. The tripwire still trips on the next deliberate change.

## Verification

- 316/316 artifacts across three consecutive runs; 18/18 security
- `pnpm -w lint` clean, `pnpm -w format:check` clean
- new coverage: the five prose shapes that each killed a run are asserted publishable; nine credential formats plus exact configured values are asserted still rejected

## Alternatives tried first

Teaching the entropy exemption successive shapes of source notation — dotted chains, single tokens, `/` and `::` separators — and applying `MIN_EXACT_SECRET_VALUE_LENGTH` to the key-name rule. Each unblocked one run and the next failed on a new shape, five times. The `For every token:` case still trips with the length floor applied, because in Markdown the captured value is the following word rather than an escape.

<h3>Greptile Summary</h3>

The PR introduces a positive-only secret-scanning mode for fail-closed artifact publication and strengthens runtime-document snapshot replacement detection.
- Keeps exact configured values, recognized credential formats, mnemonics, labeled private keys, URL credentials, and Bearer tokens blocked from publication.
- Excludes entropy, unlabeled 40-hex, and key-assignment heuristics from the publication gate while preserving them for default redaction callers.
- Detects atomic path replacement by comparing the open inode’s link count before and after reading.
- Updates publication coverage and pinned validator build identities.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| packages/artifacts/src/secret-gate.ts | Selects positive-only scanning for the fail-closed artifact publication gate while retaining byte-level checks for configured credentials. |
| packages/security/src/sensitive-redaction.ts | Adds scan-mode selection and limits Bearer matching to token-shaped values, with default callers retaining all speculative detectors. |
| packages/artifacts/src/schema-registry.ts | Adds link-count comparison to detect replacement of an open runtime document during snapshot reads. |
| packages/artifacts/test/artifacts.test.ts | Updates publication-gate expectations and covers publishable prose alongside recognized secret formats. |
| packages/artifacts/test/schema.test.ts | Updates the pinned validator build identity after the schema-registry implementation change. |
| packages/runtime/test/runtime.test.ts | Updates the runtime compatibility test to the corresponding validator build identity. |

</details>

```mermaid
flowchart LR
  A[Maintained artifact bytes] --> B[Exact configured-value scan]
  B -->|Match| X[Reject publication]
  B -->|No match| C[Decode text]
  C --> D[Positive-only recognized-secret scan]
  D -->|Recognized secret| X
  D -->|No recognized secret| E[Publish original bytes]
```
</details>

<sub>Reviews (2): Last reviewed commit: ["fix(security): require a token-shaped va..."](https://github.com/monad-developers/ultrafuzz/commit/915bf7cfbe5b8fc3050c34a49050cead181f61e0) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=56417316)</sub>

- Knowledge Base — [Artifact management](https://app.greptile.com/monad-foudnation/-/custom-context/knowledge-base/monad-developers/ultrafuzz/-/docs/artifact-management.md)
- Knowledge Base — [Security controls and reference catalog](https://app.greptile.com/monad-foudnation/-/custom-context/knowledge-base/monad-developers/ultrafuzz/-/docs/security-controls-and-references.md)
</details>


aviggiano:
Added a third commit: the `Bearer` rule is case-insensitive, so it matches the English word *bearer*. OpenZeppelin's `IAccessControl` documents "the role bearer (i.e. \`account\`)" and the rule redacted the following prose — enough to fail the gate, which cannot rewrite bytes.

The captured token must now reach `MIN_EXACT_SECRET_VALUE_LENGTH` and be made of token characters, which `(i.e.` is not. A real `Authorization: Bearer <token>` is unaffected.

Worth noting this one survived the `positive-only` scoping in the first commit — it is a *positive* pattern, not a speculative heuristic, and it was still wrong. 18/18 security and 316/316 artifacts remain green.

aviggiano (commented):
We want to avoid manually scaning secrets, we should use a third party library

aviggiano:
Superseded by #865, which lands the requested third-party migration (secretlint 13.0.4): the publication gate keeps this PR's positive-only semantics and exact-value matching, the speculative heuristics stay retired from the gate, and this PR's regression cases (role-bearer prose, the docs-example AKIA rejection, the token-shape Bearer fix) are ported as tests.

PR #823, opened 2026-08-24
Stop the ak-/as- key pattern matching kebab-case English

### LINKED ISSUE #822: ak-/as- vendor pattern matches kebab-case English, silently skipping goal-directed audit lanes
## Summary

`SECRET_PATTERNS` contains

```js
/\b(?:ak|as)-[A-Za-z0-9_-]{16,}\b/gu
```

Because the suffix class permits `-`, this matches ordinary **kebab-case English** wherever the fragment `as-` or `ak-` appears followed by 16 or more characters. `as-` in particular is extremely common in hyphenated identifiers and slugs: "treated-**as-**private-to-the-protocol", "classified-**as-**internal-only-configuration-value".

This is a *vendor* pattern, not one of the speculative heuristics, so it fires even under the `"positive-only"` publication scan proposed in #819 / #820.

## Impact

It fails artifact publication on clean output, and the failure is silent in a way the other gate failures were not.

In a real campaign this killed the `verify:goal-plan` node. That node is `continueOnFail`, so **the run did not stop** — it kept going and reported healthy progress. But two dynamic groups are sourced from `verify:goal-plan`:

```
class-goals  <- verify:goal-plan
threat-goals <- verify:goal-plan
```

Neither can expand once that verifier has failed, so the entire goal-directed portion of the audit is skipped while the run still finishes and reports success. An operator watching node counts sees nothing wrong.

## Reproduction

All three reproduce on `main` (`8caf159f`). None contains a secret.

```js
import { containsSensitiveSecrets } from "@ultrafuzz/security";

containsSensitiveSecrets("policy.record-treated-as-private-to-the-service");      // true
containsSensitiveSecrets("classified-as-internal-only-configuration-value");      // true
containsSensitiveSecrets("resources-marked-ak-restricted-to-the-owner-role");     // true
```

Agent-generated goal identifiers are exactly this shape — dotted namespace plus a hyphenated description — so the collision is routine rather than unlucky.

## Proposed fix

Drop `-` from the suffix character class:

```js
/\b(?:ak|as)-[A-Za-z0-9_]{16,}\b/gu
```

Credentials in this family are opaque tokens and do not contain hyphens; kebab-case prose always does. Verified against realistic key shapes — `ak-AbCdEf1234567890XyZwVuTs`, `as-9f2b7c1d4e6a8b0c3d5e7f90`, `ak-AbCdEf_1234567890_XyZwVu` are all still detected, and all three prose cases above become clean.

Note that snake_case prose (`treated_as_private_to_the_protocol_value`) never matched, because `_` is a word character and the leading `\b` cannot fire. Only the hyphenated form is affected.

## Related

Same family as #819: a rule calibrated for credential text misfiring on ordinary technical prose. This one is narrower and has an exact fix, so it is filed separately.

aviggiano:
Closes #822.

## Change

```diff
- /\b(?:ak|as)-[A-Za-z0-9_-]{16,}\b/gu,
+ /\b(?:ak|as)-[A-Za-z0-9_]{16,}\b/gu,
```

One character. The suffix class permitted `-`, so the pattern matched any hyphenated prose containing the fragment `as-` or `ak-` followed by 16+ characters.

## Why it matters more than it looks

This is a **vendor** pattern, not one of the speculative heuristics from #819, so it fires even under a `"positive-only"` publication scan. And the failure it causes is quiet.

In a real campaign it killed `verify:goal-plan`. That node is `continueOnFail`, so the run did **not** stop — it carried on and kept reporting healthy node counts. But two dynamic groups are sourced from that verifier:

```
class-goals  <- verify:goal-plan
threat-goals <- verify:goal-plan
```

Neither can expand once it has failed. The goal-directed portion of the audit was silently skipped while the run continued toward a "successful" finish. Nothing in the progress output indicates the loss.

## Verification

Prose that must publish (all three fail on `main`, all clean after):

- `policy.record-treated-as-private-to-the-service`
- `classified-as-internal-only-configuration-value`
- `resources-marked-ak-restricted-to-the-owner-role`

Opaque tokens that must stay detected (all still detected):

- `ak-AbCdEf1234567890XyZwVuTs`
- `as-9f2b7c1d4e6a8b0c3d5e7f90`
- `ak-AbCdEf_1234567890_XyZwVu` — underscores still allowed

The existing `modal ak-…` / `modal as-…` fixtures are unaffected, since neither contains a hyphen in the suffix. A regression test covers both directions.

`snake_case` prose (`treated_as_private_to_the_protocol_value`) never matched and still doesn't — `_` is a word character, so the leading `\b` cannot fire. Only the hyphenated form was affected.

19/19 security tests; `lint` and `format:check` clean.

## Note on the artifacts suite

`a runtime document read rejects an atomic path replacement during its snapshot` fails on this branch, as it does on `main` — it is a pre-existing ~50% flaky test, unrelated to this change. #820 fixes it (it turns out to be a real TOCTOU gap, not a slow test). Independent of this PR.

<h3>Greptile Summary</h3>

This PR narrows the Modal-style `ak-`/`as-` credential pattern so kebab-case prose no longer triggers secret detection while documented opaque token forms remain protected.
- Removes hyphens from the credential suffix character class.
- Adds regression coverage for false-positive prose, redaction behavior, and supported tokens containing alphanumerics or underscores.

The PR appears safe to merge, with focused tests covering both the corrected false-positive behavior and retained credential detection.

The narrower character class matches the established ak-/as- token forms while preventing long hyphenated prose from being rejected as secret-bearing output, and no supported credential form was found to regress.

| Filename | Overview |
|----------|----------|
| packages/security/src/sensitive-redaction.ts | Narrows the ak-/as- suffix pattern to avoid matching kebab-case prose without excluding any established valid credential format. |
| packages/security/test/sensitive-redaction.test.ts | Adds focused bidirectional regression tests for prose false positives and supported opaque credential forms. |

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(security): stop the ak-/as- key patt..."](https://github.com/monad-developers/ultrafuzz/commit/f436e647198be40992d9401db48fe2bd188f7277) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=56462836)</sub>


aviggiano (commented):
This sensitive redaction is giving us too much headache, we want preferably to use a 3rd party library instead of reinventing the wheel

aviggiano:
Superseded by #865: the ak-/as- vendor pattern (including this PR's no-hyphen suffix fix) now lives as a documented supplemental pattern alongside secretlint 13.0.4, with the kebab-case-prose regression ported as a test.

Issue #833, opened 2026-08-25
Workspace patch captures vendored Solidity dependencies: lib/ is not excluded the way node_modules is


aviggiano:
## Summary

`stageWorkspaceTree` excludes `node_modules` from workspace-patch capture, but not `lib` — the Foundry and dapptools convention for Solidity dependencies. In a Solidity fuzzing harness, that is the dependency root that matters.

```js
const WORKSPACE_RUNTIME_ROOTS = [".ultrafuzz", ".smithers", "node_modules", "artifacts"] as const;
```

Setup prompts tell the agent that dependencies may be absent from the worktree and to install them there. An agent that vendors Solidity dependencies under `lib/<name>` has that entire tree captured into the workspace patch.

## Impact

Observed twice in one campaign, with the agent choosing a different directory name each time (`lib/vendored`, then `lib/vendor`), so excluding a specific path does not help.

Each capture was **505 files and ~4.9 MB** of third-party library source in a single `workspace.patch`.

Size is not the real problem. Third-party sources legitimately contain credential-shaped text — `forge-std`'s `StdChains.sol` ships a placeholder Infura RPC URL — so the patch then **fails the artifact secret gate**, which is fail-closed and does not rewrite bytes. The node dies.

That failure is quiet. `verify:setup-foundry` is `continueOnFail`, so the run keeps going and keeps reporting healthy node counts, while everything depending on that verifier is lost. In our graph `node:base-test-setup` depends on it non-optionally, so the Foundry test-harness foundation never materialises and the run still finishes "successfully".

## Reproduction

```
git init && commit a src/ file
mkdir -p lib/vendor/forge-std/src
echo 'ChainData("Sepolia", 11155111, "https://sepolia.infura.io/v3/<32 chars>")' \
  > lib/vendor/forge-std/src/StdChains.sol
echo 'contract H {}' > test/foundry/Harness.t.sol
captureWorkspacePatch(root, baseline)
```

On `main` the manifest contains the whole `lib/vendor` tree alongside `test/foundry/Harness.t.sol`, and the resulting patch fails `assertArtifactPublicationsContainNoSecrets`.

## Proposed fix

Exclude `lib` from the **untracked** listing only, alongside the existing generated-root exclusions:

```js
const WORKSPACE_DEPENDENCY_ROOTS = ["lib"] as const;
```

[...]

PR #840, opened 2026-08-25
Tell setup agents that lib/ is already materialized

### LINKED ISSUE #839: Setup agents vendor already-present Solidity dependencies, putting megabytes of third-party source into workspace patches
## Summary

Setup agents vendor a **second copy of dependencies that are already present** in their worktree. Every untracked file is captured into `workspace.patch`, so the copy becomes megabytes of third-party source in an audit artifact — which then fails the artifact secret gate and kills the node.

This happened on three consecutive runs of one campaign, each time under a different directory name:

```
lib/vendored      (505 files, 4.9 MB)
lib/vendor        (505 files, 4.9 MB)
lib-vendor        (505 files, 4.9 MB)
```

Because the name changes every time, path-based exclusion does not fix it — including the `lib/` exclusion in #833/#834, which `lib-vendor/` sidesteps by being a sibling rather than a child.

## The dependencies were already there

This is not a missing-dependency workaround. In the failing worktree:

```
lib/forge-std/src/Base.sol          <- present and populated
lib-vendor/forge-std/...            <- agent's duplicate copy
```

The submodules are materialized before the agent runs. It copied them anyway.

## Why the agent does it

The worktree-isolation preamble injected ahead of the task prompt says:

> Dependencies (node_modules) may be absent here. Install them fresh inside the worktree (e.g. `pnpm install` at the worktree root)

That is JavaScript-specific guidance, but it is the only statement about dependencies the agent sees, and nothing tells it that `lib/` is already populated. A Solidity agent generalises "dependencies may be absent, install them here" to its own ecosystem and defensively vendors `lib/`.

## Why it is expensive

`verify:setup-foundry` is `continueOnFail`. The run does not stop — it keeps reporting healthy node counts while `node:base-test-setup`, which depends on that verifier non-optionally, never runs. The Foundry test-harness foundation is silently absent from a run that finishes "successfully".

The existing size ceiling does not catch it either: `MAX_PATCH_BYTES` is 16 MB and these patches are ~5 MB, so the measured-overflow exclusion path never triggers.

## Fix in this PR

State it in the setup prompt: `lib/` is already materialized, the preamble's "dependencies may be absent" note describes JavaScript packages, and no dependency tree may be copied or vendored under any name — with the reason attached, so the instruction is not arbitrary.

## Recommended follow-up (not in this PR)

A prompt is advisory to a model, and this behaviour has recurred three times. A robust guard belongs in capture: treat a single untracked top-level directory contributing an unusually large number of files or bytes as a bulk contributor and exclude it, recording it in the manifest the way `git-diff-overflow` exclusions already are. That is name-agnostic and reuses machinery that exists, but it is a policy decision about capture semantics rather than a bug fix, so it is proposed here rather than assumed.

aviggiano:
Closes #839.

## The problem

Setup agents vendor a **second copy of dependencies that are already present** in their worktree. Three consecutive runs of one campaign, three different directory names:

```
lib/vendored      505 files, 4.9 MB
lib/vendor        505 files, 4.9 MB
lib-vendor        505 files, 4.9 MB
```

Every untracked file is captured into `workspace.patch`, so each copy put megabytes of third-party source into an audit artifact. That then failed the artifact secret gate on credential-shaped strings those libraries legitimately contain (`forge-std`'s `StdChains.sol` placeholder Infura URL), killing the node.

**Path exclusion cannot fix this.** The name changes every time — `lib-vendor/` sidesteps the `lib/` exclusion in #833/#834 by being a sibling rather than a child.

## The dependencies were already there

```
lib/forge-std/src/Base.sol     <- present and populated before the agent ran
lib-vendor/forge-std/...       <- the agent's duplicate
```

## Why the agent does it

The only statement about dependencies the agent sees is the worktree-isolation preamble injected ahead of the task prompt:

> Dependencies (node_modules) may be absent here. Install them fresh inside the worktree (e.g. `pnpm install` at the worktree root)

JavaScript-specific, and nothing says `lib/` is already populated. A Solidity agent generalises it to its own ecosystem and defensively vendors.

## Why it is expensive

`verify:setup-foundry` is `continueOnFail`, so the run does **not** stop. It keeps reporting healthy node counts while `node:base-test-setup` — which depends on that verifier non-optionally — never runs. The test-harness foundation is silently missing from a run that finishes "successfully".

`MAX_PATCH_BYTES` is 16 MB and these patches are ~5 MB, so the measured-overflow exclusion never triggers either.

## This change

States it in the setup prompt, with the reason attached so the instruction is not arbitrary:

- `lib/` is materialized before the agent runs
- the preamble's "dependencies may be absent" note describes JavaScript packages
- no dependency tree may be copied, vendored or re-downloaded under any name
- a genuinely missing dependency is recorded as **blocked** rather than worked around
- and *why*: untracked files land in the workspace patch, which fails the gate and kills dependent lanes

`prompts` build and tests pass; `docs:check` and `format:check` clean.

## Honest limitation

A prompt is advisory to a model, and this behaviour has now recurred three times. I would not call it fixed until there is a guard in capture.

#839 proposes the robust follow-up: treat a single untracked top-level directory contributing an unusually large number of files or bytes as a bulk contributor and exclude it, recording it in the manifest the way `git-diff-overflow` exclusions already are. That is name-agnostic and reuses machinery that exists — but it is a policy decision about capture semantics rather than a bug fix, so I have proposed it rather than assumed it.

<h3>Greptile Summary</h3>

This PR updates the Foundry harness setup prompt to clarify that Solidity dependencies are already materialized and must not be duplicated. It also explains how duplicate dependency trees contaminate workspace patches and instructs agents to report genuinely missing dependencies as blocked.

- Prohibits copying, vendoring, or re-downloading dependency trees under alternate paths.
- Distinguishes materialized Solidity dependencies from the preamble guidance for JavaScript packages.
- Documents the workspace-patch and artifact-gate consequences of vendored dependencies.

The PR appears safe to merge because the prompt-only clarification preserves the existing local-support fallback and introduces no established blocking behavior.

The changed instructions prevent redundant dependency trees while retaining the existing option to create minimal local test support when forge-std is unavailable.

| Filename | Overview |
|----------|----------|
| .ultrafuzz/prompts/setup/prepare-foundry-harness.md | Adds focused setup-agent guidance preventing duplicate dependency vendoring without introducing a concrete prompt-contract regression. |

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(prompts): tell setup agents that lib..."](https://github.com/monad-developers/ultrafuzz/commit/8496425add43b953e856451a2422aa3b4082882f) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=56493823)</sub>


.ultrafuzz/prompts/setup/prepare-foundry-harness.md
```diff
@@ -26,6 +26,20 @@ interfaces when the target repository already generates them.
 
 Create only the minimal harness layout needed by later fuzzing agents.
 
+Solidity dependencies are already materialized in this worktree. The harness
```
aviggiano, line 29: maybe change to 'should already have been' instead of 'are'

aviggiano: After this change, LGTM

aviggiano:
The review-thread wording change (r3852269925) landed on main via https://github.com/monad-developers/ultrafuzz/pull/862.

Issue #364, opened 2026-08-07
Decide where the six threat-model-dependent dynamic-fanout metrics in #183 are derived from


aviggiano:
Follow-up to #183, split out of the dynamic-fanout observability prerequisite (PR pending against `main`, cross-referencing #327). **A decision is needed that I could not make alone; six of #183's twelve metrics are blocked on it.**

## What is already implemented

`evalRunExpansion` derives, per row, from `state.json` and `graph.json` alone — no reporter, no provider, no network:

| #183 metric | status |
| --- | --- |
| generated / ready / active / terminal / failed / skipped counts | done, per `NODE_STATE_STATUSES` |
| requested vs effective concurrency | done, from `RunState.concurrency` |
| retries and timeouts | done |
| source node IDs | done, read from a dynamic node's provenance |
| actual dynamic child count | done, as "in `state.json`, absent from `graph.json`" |
| wall-clock, tokens, cost | already in `EvalEfficiency`, at run level |

That is enough for the two claims the gate exists to make — dynamic children are independently visible, and a wide ready queue was admitted under the configured limit rather than serialized.

## What is blocked, and why

These six have no derivable definition on `main`:

1. **number of structured threats**
2. **number of applicable database classes**
3. **expected dynamic child count**
4. **`max_dynamic_nodes` setting**
5. **failed goal lanes** (as distinct from failed nodes)
6. **per-lane tokens / cost / wall-clock** (as distinct from per-run)

1–3 and 5 need the threat model, the vulnerability database and goal lanes — #182, #180 and #179, none of which are on `main`. 4 is a setting that does not exist yet. 6 needs a lane-to-node grouping that nothing currently records.

Guessing any of them would produce a metric that reads as measured and is not.

## The decision

**Where does "expected" come from, and who owns the number?** #183 states the rule as `threats + applicable classes`, plus the fixed roaming node. Three ways to make that checkable:

- **(a) The planner writes it down.** `goal-plan.json` records `expected_child_count`, `threat_count`, `applicable_class_count` and `max_dynamic_nodes` at planning time; the eval side reads that file and compares it against the observed dynamic count. Expected and actual come from different places, so a planner that under-expands is caught.
- **(b) The eval side recomputes it** from the threat model and database manifest. Independent of the planner, but it duplicates the cardinality rule in a second codebase, and the two will drift.
[...]

PR #848, opened 2026-08-25
Stop the goal-plan prompt asking for the derived counts it forbids

### LINKED ISSUE #847: goal-plan prompt tells the agent to write the derived counts and not to write them; the resulting failure silently drops all goal lanes
## Summary

`setup/goal-plan.md` tells the planning agent to write the derived count fields and not to write them, in consecutive paragraphs. Agents resolve the contradiction by writing them, which fails the plan closed and silently removes the entire goal-directed portion of the audit.

## The contradiction

```
...retain the selected source-record identities, and include the fixed roaming
goal and reconciled counts required by the schema.          <- write the counts

Do not write `expected_child_count`, `threat_count`, `applicable_class_count`,
`max_dynamic_nodes`, or `goal_lanes`.                       <- do not write them

The canonical schema lists them as required because they are present by the time
the plan is validated.                                      <- the schema requires them
```

Two of the three statements point at writing the counts. A model following instructions writes them.

## What happens next

`recordGoalPlanExpansionFacts` fails closed on any disagreement, by explicit design (#364 option (a)):

> A value the agent already wrote that disagrees with the plan it also wrote is an explicit failure, never a silent correction: a plan that misstates its own cardinality must not reach the gate looking consistent.

That behaviour is correct and I am not proposing to change it. The problem is that the prompt invites the disagreement.

The observed plan was internally consistent and still failed. `goalPlanExpansionFacts` computes

```js
expected_child_count: threatCount + applicableClassCount   // roaming excluded
```

The agent wrote `expected_child_count: 76` for 19 threat goals + 56 class goals + 1 roaming goal, having counted the roaming lane. Ultrafuzz derives `75`. It also wrote `max_dynamic_nodes: 75` while the run was configured for `2048`, apparently confusing the two.

So the agent produced a coherent plan and failed on arithmetic it was told not to perform, that it could not have got right without knowing an undocumented exclusion rule.

## Why it is expensive

`verify:goal-plan` is `continueOnFail`, so the run does **not** stop. It keeps reporting healthy node counts. But two dynamic groups are sourced from that verifier:

```
class-goals  <- verify:goal-plan
threat-goals <- verify:goal-plan
```

Neither can expand once it has failed, so 75 goal lanes never run and the campaign still finishes "successfully". Nothing in the progress output shows the loss. This is the third distinct failure with the same shape in one campaign — a `continueOnFail` verifier that is the *source* of dynamic groups silently deleting those groups.

## Fix in this PR

Remove the instruction to write the counts, state plainly that the five keys must be omitted entirely, and explain the schema's `required` marking so it does not read as a contradiction. Also state the one non-obvious rule — `expected_child_count` excludes the roaming goal — so an agent that writes them anyway is less likely to be wrong.

## Recommended follow-up (not in this PR)

The cost here comes from the silence, not from the plan. A `continueOnFail` verifier that is the declared source of a dynamic group should not be able to delete that group without surfacing it: either the group's absence should be an explicit run-level degradation, or such a verifier should not be `continueOnFail`. That is a topology-semantics decision rather than a bug fix, so it is proposed rather than assumed.

aviggiano:
Closes #847.

## The contradiction

`setup/goal-plan.md`, in consecutive paragraphs:

```
...include the fixed roaming goal and reconciled counts required by the schema.
Do not write `expected_child_count`, `threat_count`, `applicable_class_count`,
  `max_dynamic_nodes`, or `goal_lanes`.
The canonical schema lists them as required because they are present by the time
  the plan is validated.
```

Two of the three point at writing the counts. Agents write them.

## What that costs

`recordGoalPlanExpansionFacts` fails closed on any disagreement, by explicit design (#364 option (a)) — *"a plan that misstates its own cardinality must not reach the gate looking consistent"*. **That behaviour is correct and is unchanged here.** The prompt was inviting the disagreement.

The observed plan was internally consistent and still failed:

```js
expected_child_count: threatCount + applicableClassCount   // roaming excluded
```

The agent wrote `76` for 19 threat + 56 class + 1 roaming — it counted the roaming lane. Ultrafuzz derives `75`. It also wrote `max_dynamic_nodes: 75` while the run was configured for `2048`. So it failed on arithmetic it was told not to perform and could not have got right without knowing an undocumented exclusion rule.

And `verify:goal-plan` is `continueOnFail`, so the run kept going and kept reporting healthy node counts while both dynamic groups sourced from it —

```
class-goals  <- verify:goal-plan
threat-goals <- verify:goal-plan
```

— could never expand. **75 goal lanes silently vanished from a campaign that would have finished "successfully".**

## This change

- drops "and reconciled counts required by the schema" from the instruction paragraph
- says plainly that the five keys must be **omitted entirely**
- explains *why* the schema marks them required, so it stops reading as a contradiction
- states the one non-obvious rule — `expected_child_count` counts threat and class goals only, excluding roaming — so an agent that writes them anyway is less likely to be wrong

`prompts` build and tests pass; `docs:check` and `format:check` clean.

## Honest limitation

Same caveat as #840: a prompt is advisory to a model. The real cost here is the **silence**, not the plan.

A `continueOnFail` verifier that is the declared source of a dynamic group should not be able to delete that group without surfacing it — either the group's absence becomes an explicit run-level degradation, or such a verifier is not `continueOnFail`. This is the third distinct failure with that shape in one campaign (`goal-plan` twice, `setup-foundry` once). It is a topology-semantics decision rather than a bug fix, so #847 proposes it rather than assuming it.

<h3>Greptile Summary</h3>

The PR resolves the prompt/schema contradiction by permitting goal-plan authors to omit runtime-derived expansion fields while preserving all-or-none validation when those fields are supplied.
- Updates the goal-plan prompt to require omission of the five derived fields.
- Aligns the JSON Schema and Zod schema with that authoring contract.
- Records derived expansion facts before runtime parsing and guards eval consumers against unrecorded plans.
- Adjusts affected tests for the newly optional parsed type.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| .ultrafuzz/prompts/setup/goal-plan.md | Removes contradictory count-writing guidance and clearly instructs the agent to omit runtime-derived fields. |
| packages/artifacts/schema/goal-plan.schema.json | Removes the five derived expansion fields from the JSON Schema required list. |
| packages/artifacts/src/goal-plan.ts | Makes derived fields optional at authoring validation while preserving all-or-none and supplied-value consistency checks. |
| packages/evals/src/expansion.ts | Explicitly rejects goal plans that reach evaluation before their derived expansion facts have been recorded. |

</details>

```mermaid
flowchart LR
    A[Agent authors goal plan without derived fields] --> B[Schema and contract validation]
    B --> C[Runtime records expansion facts]
    C --> D[Runtime parses and verifies enriched plan]
    D --> E[Threat and class goal consumers]
    B --> F[Eval reader]
    F --> G{Derived fields present?}
    G -->|Yes| H[Use expansion facts]
    G -->|No| I[Report goal-plan-unreadable]
```
</details>

<sub>Reviews (4): Last reviewed commit: ["fix(runtime): optional-chain goal\_lanes ..."](https://github.com/monad-developers/ultrafuzz/commit/e5d2958bf6c02fff6f1d6d31d5f223d246d447ce) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=56506359)</sub>


aviggiano:
Do not merge this just yet, we're discussing an alternative fix

aviggiano:
Second commit — **the first commit was wrong on its own** and would have made things worse. Thanks to @aviggiano for the push on this.

The premise I missed: the schema marked all five keys `required`. So an agent that obeyed "do not write them" failed its own validation command:

```
$ ultrafuzz json validate --schema goal-plan.schema.json --file plan.json
must have required property 'applicable_class_count'
must have required property 'expected_child_count'
must have required property 'goal_lanes'
must have required property 'max_dynamic_nodes'
must have required property 'threat_count'
exit 1
```

The prompt tells agents to finish only once every validation command exits 0. So omitting was impossible, and my first commit pushed harder toward omitting.

Writing them was equally impossible to get right:
- `recordGoalPlanExpansionFacts` fails closed on any disagreement
- `expected_child_count` excludes the roaming goal, undocumented
- `max_dynamic_nodes` is a run setting the prompt never states — `2048` appears nowhere in the rendered prompt

**No output satisfied both checks.** The contract was unsatisfiable.

Since `recordGoalPlanExpansionFacts` populates all five after the agent returns, they are now **optional**. An obedient agent validates. The pre-existing document-local checks (`expected_child_count must equal N`, the `counts` cross-check, `goal_lanes` completeness) are unchanged and still reject a disagreeing value — they are only guarded so they apply to fields the agent actually wrote. That puts the arithmetic in the validation the agent runs itself.

Verified: keys omitted → `json validate` exit 0 and contract accepts; wrong `expected_child_count: 76` → contract rejects with `must equal 75`.

### Separate finding worth its own issue

`ultrafuzz artifact validate ultrafuzz/goal-plan@1 <plan>` reports **Valid** for a plan the zod contract rejects. `validateJsonContractBytes` runs `validateRegisteredJsonBytesSync` only — JSON Schema, never the registered runtime schema — so cross-field `superRefine` semantics are not enforced by the command whose help says it validates "document-local contract semantics".

That affects every contract with `superRefine` rules, not just goal-plan: agents are told to run it and stop when it exits 0, while the host later applies stricter rules. I have not changed it here — wiring the runtime contract in is a behavioural change across all contracts and deserves its own review.

aviggiano:
LGTM

aviggiano:
LGTM but first fix CI errors

aviggiano:
CI is fixed. The failing job ("PR build and Node.js 24 runtime smoke", run 32841477619) had exactly one real error:

```
test/vulnerability-database.test.ts(236,5): error TS18048: 'snapshot.plan.goal_lanes' is possibly 'undefined'.
```

A direct consequence of this branch making the five host-recorded fields optional in the zod contract: the inferred type of `goal_lanes` now includes `undefined`, and this runtime test still accessed it unguarded. Fixed with the same optional chaining the branch already applied to the equivalent assertions in `packages/artifacts/test/threat-goal-artifacts.test.ts` — the `deepEqual` against the expected lane list still fails if the value were ever actually missing, and `materializeGoalPlanVulnerabilityDatabaseSnapshots` records the facts before returning, so it never is. The `release-gates` failure was just the downstream required-gate of that job.

Also rebased onto current `main`.

Local validation (equivalent of the failing job plus the touched packages):

- `pnpm --filter @ultrafuzz/runtime test:pr-smoke:prebuilt`: `tsc -p tsconfig.test.json` now compiles clean; all node smoke tests pass; the 4 Bun adapter-contract tests pass (4 pass / 0 fail). (The script's `(pass)`-marker scrape doesn't fire on my local Bun output formatting, but the same script + Bun 1.3.14 asserts it fine in CI — see the sibling passing run 32841544022.)
- `pnpm --filter @ultrafuzz/prompts test`: 103/103
- `pnpm --filter @ultrafuzz/artifacts test`: 322/322
- `pnpm --filter @ultrafuzz/evals test`: 426/426
- every runtime test file referencing goal-plan (`vulnerability-database`, `dynamic-workflow`, `dynamic-lifecycle`, `topology-transform`, `cloud-worker-handoff`, `final-report-markdown`): 50/50 after building `@ultrafuzz/modal` locally
- `typecheck` for artifacts, evals, prompts, runtime: clean; prettier + eslint clean on the touched file

Proceeding with the squash-merge per the LGTMs.

PR #870, opened 2026-08-25
Accept SQL-null optional workflow input fields

### LINKED ISSUE #868: Every local run dies at render: workflow input envelope keys reject the SQL nulls its own input table returns
## Summary

No local `ultrafuzz` run can render. The generated Smithers workflow re-validates its own dispatch in `parseWorkflowInput` and rejects it, before the first task node is constructed:

```
ultrafuzz workflow input is invalid: cloud_worker: Invalid input: expected true; task_id: Invalid input: expected string, received null; attempt_id: Invalid input: expected string, received null; execution_generation: Invalid input: expected string, received null
```

This is the default path — a local dispatch, no cloud worker, no dynamic groups needed to trigger it.

## Cause

`packages/runtime/src/templates/smithers/workflows/workflow.tsx` declares one closed input envelope and the comment above it already explains why:

```ts
/**
 * One closed object rather than a union of the local and cloud-worker envelopes.
 * The workflow runner projects this schema into its input table by walking
 * `shape`, and a union exposes no shape to walk, so a union fails every detached
 * submission during preflight. ...
 */
```

That same projection is what breaks it. The runner walks `shape` and materializes **a column for every declared key**, so a key the submission never set does not come back as `undefined` — it comes back as SQL `null`. Every key in the envelope is declared `.optional()`:

```ts
cloud_worker: z.literal(true).optional(),
task_id: z.string().min(1).max(4_096).optional(),
attempt_id: z.string().min(1).max(4_096).optional(),
execution_generation: z.string().regex(...).optional(),
selected_task: z.json().optional(),
```

`.optional()` accepts `undefined` only. A local dispatch leaves all five cloud-worker keys unset by definition, so the round trip through the input table hands `parseWorkflowInput` five explicit nulls and the parse fails.

The envelope discrimination in `superRefine` has the same defect in the other direction:

```ts
const localKeys = LOCAL_WORKFLOW_INPUT_KEYS.filter((key) => value[key] !== undefined);
const cloudKeys = CLOUD_WORKER_INPUT_KEYS.filter((key) => value[key] !== undefined);
```

Were the keys merely widened to accept null, a null-filled local input would be read as a *cloud* dispatch, because `null !== undefined` counts every cloud key as present.

## Fix

Declare the envelope keys `.nullish()` and treat null as absence everywhere absence is tested, including the two discrimination filters. Null must remain absence, never a stand-in for a key the selected envelope requires.

## Impact

P0. `main` cannot complete a local run at all: every local dispatch dies at render.

aviggiano:
## Summary

- accept SQL `null` values for projected optional workflow-envelope columns
- treat `null` as absence during local/cloud dispatch discrimination and required-field checks
- add regressions for SQL-hydrated local and cloud rows, including null required-member rejection

## Tests

- focused generated workflow schema and table-projection tests
- `pnpm --filter @ultrafuzz/runtime test:pr-smoke:prebuilt`

Fixes #868


aviggiano:
Closing as duplicate of #874, which fixes the same workflow.tsx input-envelope null handling and additionally carries cloud-worker-handoff.test.ts coverage (+98 vs +49 over the same files). Triage call per cross-session coordination; #874 is the survivor.

Issue #674, opened 2026-08-19
No supported way to re-seal workflow control evidence: one divergent file permanently disables product-level status and reporting


aviggiano:
## Summary

There is no supported way to re-seal or re-bless workflow control evidence. Once a run's sealed control files diverge for any reason, `WORKFLOW_CONTROL_EVIDENCE_INVALID` is permanent: `ultrafuzz status` refuses the run for the rest of its life, so no state sync, no completion gate, and no final report can ever be produced — even while the underlying Smithers run continues to execute and finish work.

## Evidence

`verifyWorkflowControlSnapshot` compares the published snapshot's control files byte-for-byte (`packages/runtime/src/workflow-integrity.ts:291`):

```ts
throw new Error(`sealed workflow control file changed: ${controlFileLabel(key)}`);
```

which `start-run.ts` converts into the terminal diagnostic:

```
code:    WORKFLOW_CONTROL_EVIDENCE_INVALID
message: sealed workflow control file changed: workflow
```

In an 18-hour local run this fired from a **single** changed file. Verifying the seal by hand:

- all **44,345** sealed `execution_files` hashed clean — zero mismatches
- the only divergence was the published snapshot's `.smithers/workflows/<run>.tsx`:

```
seal expects    f37c4eb6…  2007734 bytes
generated copy  f37c4eb6…  2007734 bytes   ← still byte-identical to the seal
snapshot copy   96c367e8…  2008029 bytes   ← diverged
```

So the damage was one file out of 44,346, and the pristine bytes were still sitting on disk. There is still no way to recover, because `sealWorkflowControlFiles` refuses to write a second seal (`workflow-integrity.ts:233`):

```ts
if (pathEntryExists(paths.integrityPath)) {
  throw new Error("workflow control seal already exists");
}
```

There is no `--reseal`, no `--accept-control-change`, no generation bump. The run is bricked at the product level regardless of how repairable the underlying divergence is.

## Why the divergence happened

Because agentic nodes compile with `retries: 0` (filed separately), the only available way to grant a failed node another attempt was to patch the sealed snapshot's workflow. Smithers itself accepts this via `smithers up --force --accept-workflow-change`, and the run genuinely resumed and kept finishing nodes. Ultrafuzz then had no matching concept, so the product layer was locked out of a run that was still alive and progressing.

That asymmetry is the core defect: **Smithers has a supported "I acknowledge the workflow changed" path and Ultrafuzz does not.**

## Impact

[...]

PR #871, opened 2026-08-25
Let a read-only run report survive manifest re-derivation

### LINKED ISSUE #866: ultrafuzz status is unusable on every run: manifest re-derivation throws past tolerateControlDivergence (#674 incomplete)
## Summary

`ultrafuzz status <run-id>` fails on **every** run in a campaign:

```
$ ultrafuzz status <run-id> --project <project>
error: WORKFLOW_CONTROL_EVIDENCE_INVALID: Smithers task "dedupe-findings" planned dependency nodes do not match
```

Reproduced on 12 consecutive runs, all identical. The only supported way to observe a run is unusable, so an operator either flies blind or writes their own log scraper.

`ultrafuzz ps` works but reports **every** run as `running`, including runs whose controllers were killed hours earlier, so it cannot be used to tell live runs from dead ones either.

## This is the failure #674 already tried to prevent

`getRunHealth` in `state-export.ts` opts into tolerance explicitly, and the comment states the intent:

```ts
// `status` reports on a run; it never executes one. Reading it must not require the authority to
// resume it, or a single divergent control file makes an otherwise healthy run permanently
// unobservable (issue #674). Divergences are surfaced as warnings below rather than suppressed.
const evidence = await readLinkedWorkflowEvidence(projectRoot, input.runId, {
  tolerateControlDivergence: true
});
if (!evidence.ok) {
  return runtimeFailure<RunHealthValue>(evidence.diagnostics);
}
```

The code immediately below is built for the degraded path — it maps divergences to `WORKFLOW_CONTROL_EVIDENCE_DIVERGED` **warnings**, skips synchronisation, and attaches `WORKFLOW_STATE_SYNC_SKIPPED` saying "reported counts come from the workflow runner and local run state may be stale". That path is never reached.

## Why the tolerance does not apply

`tolerateControlDivergence` reaches only `verifyWorkflowControlSnapshot`:

```ts
const verifiedControl = verifyWorkflowControlSnapshot(resolvedProjectRoot, layout, {
  tolerateDivergence: options.tolerateControlDivergence === true
});
```

Everything after that re-derives the sealed manifest and asserts with hard `throw`s that no flag can downgrade — `assertControllerExecutionSnapshotDigest`, `assertSealedDataGovernance`, then `parseSealedTaskManifest`, which reaches:

```ts
// packages/artifacts/src/smithers-task-manifest.ts:1537
assertSameStringSet(
  task.metadata.dependencies.concreteNodeIds.filter((d) => d !== "__start__"),
  node.depends_on,
  `Smithers task ${JSON.stringify(task.attemptId)} planned dependency nodes`
);
```

So a mismatch between the sealed manifest and a re-derivation of it becomes fatal for a read-only command.

`dedupe-findings` is the node that trips it because it depends on every finding producer, including dynamic goal lanes. Any run whose dynamic expansion diverges from the sealed plan — for example one where the verifier that sources a dynamic group failed, so the group never expanded — will mismatch there and become permanently unobservable.

## Impact

Observability is lost exactly when it is most needed: on a long run that has partially failed. In one campaign this meant no working status for the entire session, and the substitute log scraper written in its place produced its own wrong numbers (a stale denominator that ignored dynamic expansion, reporting 16% while 31 dynamic lanes ran).

`ps` reporting killed runs as `running` compounds it — neither command distinguishes a live run from a dead one.

## Suggested fix

1. Make the read-only path honour its own contract: when `tolerateControlDivergence` is set, manifest re-derivation mismatches should become `WORKFLOW_CONTROL_EVIDENCE_DIVERGED` warnings and let `getRunHealth` return degraded state, rather than throwing. The downgrade machinery already exists a few lines below and is unreachable today.
2. Keep the hard assertions for paths that *execute* a run (`start-run`, `resume`, `cancel`), where refusing on divergence is correct.
3. Fix `ps` status derivation so a run whose controller is gone is not reported as `running`.

A regression test should cover the shape that fails today: a recorded run whose sealed manifest disagrees with the re-derived dependency set must still return status, with a divergence warning attached.

aviggiano:
Closes #866.

## What changed

`getRunHealth` already asks `readLinkedWorkflowEvidence` for
`tolerateControlDivergence`, and the code immediately below it is written for the
degraded answer: divergences become `WORKFLOW_CONTROL_EVIDENCE_DIVERGED`
warnings, synchronization is skipped, and a `WORKFLOW_STATE_SYNC_SKIPPED` note
says the reported counts come from the workflow runner and local run state may be
stale. That path was unreachable.

#674 put the tolerance at one place: the digest comparison inside
`verifyWorkflowControlSnapshot`. Everything downstream of that comparison
re-derives the compiled plan rather than checking a digest, and every
re-derivation threw unconditionally. Three checks stood between the flag and the
degraded path, and all three now report a divergence when — and only when — the
caller asked for tolerance:

1. **`packages/runtime/src/workflow-integrity.ts`** — a graph or task plan that
   changed with no compiled dynamic group to authorize the change. Whether the
   change is *authorized* is an execution question; the digest loop one iteration
   earlier has already reported both files as diverged, so refusing here told a
   read-only caller nothing new and cost it the whole run.
2. **`packages/runtime/src/workflow-integrity.ts`** —
   `deriveWorkflowControlBindings` re-runs the compiler's cross-document gates,
   so it can fail outright rather than merely disagree with the seal. A binding
   *mismatch* was already only a divergence; a hard failure on the strictly worse
   version of the same fact bought nothing.
3. **`packages/runtime/src/start-run.ts`** — `parseSealedTaskManifest`, whose
   `assertSmithersTaskManifestMatchesPlannedGraph` is the check that named
   `dedupe-findings` in the report. Observers now keep the manifest they can still
   parse and hand the mismatch back as a divergence.

All three are load-bearing: the new test fails on main both with the
`workflow-integrity.ts` half reverted and with the `start-run.ts` half reverted
(evidence below).

## Why degrading the read-only path is safe while execution still fails hard

- **One flag, one caller.** `tolerateControlDivergence` is passed by exactly one
  call site: `getRunHealth`. `startRun`, the resume/cancel lifecycle in
  `start-run.ts`, `syncRun`, `getRunStatus` and the eight `lifecycle-inspection.ts`
  readers all call `readLinkedWorkflowEvidence` without it.
- **Execution behaviour is byte-identical.** Every downgrade goes through
  `reportDivergence`, which still throws the same message it threw before when
  `tolerateDivergence` is false, or through a branch guarded on the flag. No
  execution path was rewritten.
- **Nothing structural is relaxed.** The seal, the sealed planned graph and the
  task manifest must still parse against their schemas; run-ID, compiled-identity,
  workflow-name, stored-path, durable-state and link-journal checks all still fail
  closed for the observer too.
- **Execution-file divergence stays fatal**, for the reason #674 gave: the
  snapshot environment binds the runner executable and the sealed module URLs an
  observer actually executes. The existing test "an observer still refuses a run
  whose sealed execution files diverged" continues to pass.
- **A degraded read cannot write back.** The graph and task contents are never
  handed to the runner — `getRunHealth` only runs `smithers status <workflow-run-id>`
  under the strictly verified snapshot environment — and synchronization is still
  skipped while any divergence stands, so the stale view is reported, never
  persisted.
- The new test asserts the refusal directly: the strict
  `readLinkedWorkflowEvidence`, `resumeRun` and `cancelRun` all still return
  `WORKFLOW_CONTROL_EVIDENCE_INVALID` on the same run that `status` can now read.

## Evidence the test catches the bug

`packages/runtime/test/runtime.test.ts`, next to the existing #674 divergence
tests: "a sealed manifest that stops re-deriving leaves status readable while
execution stays closed". It starts a real run, then removes one dependency from
the sealed task manifest's own record of what the join node depends on — the
manifest still satisfies its schema and the planned graph is untouched, so the
failure is a disagreement between two control documents rather than a control file
that stopped matching its digest.

Stash/rebuild/run results, one test, on this branch:

| source state | result |
| --- | --- |
| both changes present | **pass** |
| `start-run.ts` + `workflow-integrity.ts` reverted | **fail** — observer read returns `WORKFLOW_CONTROL_EVIDENCE_INVALID: sealed workflow graph or task plan changed` |
| only `start-run.ts` reverted | **fail** — observer read returns `WORKFLOW_CONTROL_EVIDENCE_INVALID: Smithers task "final-report" planned dependency nodes do not match` |

The second row reproduces the reported error verbatim, modulo the node name.

## What I did not do

- **The specific mismatch #866 reported is already fixed on main, by #857.** That
  PR taught the dependency-node check to accept a task whose concrete dependencies
  omit a still-pending dynamic group. Reading the twelve affected runs on disk
  confirms that is the shape they had: `dedupe-findings` recorded 62 concrete
  dependencies against 64 graph edges, the two extra being the unexpanded
  `threat-goals` and `class-goals` placeholders. The campaign ran a build that
  predates #857, whose copy of the assertion compared the two lists directly. So
  the reported symptom will not reproduce on main; what this PR fixes is the
  structural gap behind it, which would take `status` down again on the next
  disagreement between two control documents.
- **`ultrafuzz ps` still reports a run as `running` when its controller is gone**
  (item 3 of the issue). `listRuns` prints `ultrafuzz_status` from the durable
  `state.json` in preference to the joined runner row, and only `sync` and the
  lifecycle commands ever move that status — `ps` never reconciles, so a run whose
  controller was killed keeps whatever it last wrote. Fixing it means deciding
  what liveness signal is
  authoritative (controller ownership record, heartbeat staleness, or the runner's
  own `ps` view) and whether a read-only command may reconcile durable state at
  all. That is a design decision, not a contained change, so I left it out rather
  than guess. It should stay open as its own issue.
- **One re-derivation is deliberately still fatal for observers.** If a dynamic
  run's persisted runtime graph or task plan does not re-derive from its sealed
  base templates and expansion manifests, `verifyDynamicRuntimeMaterialization`
  still throws for everyone. Downgrading that one means threading the tolerance
  into `dynamic-runtime.ts`, and unlike the three above it leaves the observer
  with no trustworthy node set at all rather than a merely inconsistent one. I
  left it closed; say the word if you want it opened too.
- I did not touch `packages/artifacts`, so no schema or zod contract changed.
  `schema:check` was run anyway and passes.
- I did not widen `getRunStatus`, which reads evidence strictly and already
  degrades its own diagnostics to warnings. It is unchanged.

## Verification

Run locally on this branch:

| lane | result |
| --- | --- |
| `pnpm -w lint` | clean |
| `pnpm -w format:check` | clean |
| `pnpm --filter @ultrafuzz/artifacts run schema:check` | clean |
| `pnpm --filter @ultrafuzz/artifacts test` | 321/322 (see flake note) |
| runtime supporting files (`run-tests.mjs supporting`, 108 files) | 752 tests, **752 pass, 0 fail** |
| `runtime.test.js` shard 1/4 | **61 pass, 0 fail** |
| `runtime.test.js` shard 2/4 | **51 pass, 0 fail** |
| Bun adapter contracts | **40 pass, 1 skip, 0 fail** |

**Shards 3/4 and 4/4 of `runtime.test.js` were not completed locally.** This
environment SIGTERMs long-running invocations, so neither the single
`pnpm --filter @ultrafuzz/runtime test` invocation nor those two shards could run
to completion — the monolithic run was killed at ~50 minutes after 693 passing
tests with zero failures, including the new regression test. That was an
infrastructure timeout, not a test failure, but it means **CI should be treated as
the authority for shards 3 and 4**. Everything the local lanes did cover is green.

The one failing artifacts test is
`a runtime document read rejects an atomic path replacement during its snapshot`
in `packages/artifacts/test/run-documents.test.ts` — a known pre-existing flake on
main (it failed 1 of 3 re-runs of that file in isolation) in a package this PR does
not touch.

<h3>Greptile Summary</h3>

This PR makes read-only run health reporting tolerate sealed control documents that remain schema-valid but no longer re-derive consistently, while preserving strict failure behavior for execution and lifecycle operations.
- Converts graph/task-plan and completeness-binding re-derivation failures into control divergences only when tolerance is explicitly requested.
- Preserves a parseable sealed task manifest for observers and reports its planned-graph mismatch through the existing divergence channel.
- Adds integration coverage proving health remains readable, synchronization is skipped, and strict read, resume, and cancel paths remain closed.

The PR appears safe to merge because the new tolerance remains confined to the read-only health path while synchronization and execution paths continue to fail closed.

The tolerant caller reports re-derivation mismatches through the existing divergence channel, uses runner-derived status, and skips durable-state synchronization; strict readers and lifecycle operations still reject the same inconsistent evidence.

| Filename | Overview |
|----------|----------|
| packages/runtime/src/workflow-integrity.ts | Extends explicit read-only tolerance to runtime-control and completeness-binding re-derivation failures while retaining strict behavior when tolerance is absent. |
| packages/runtime/src/start-run.ts | Adds observer-only manifest parsing and merges semantic mismatches into the existing control-divergence channel after preserving downstream identity checks. |
| packages/runtime/test/runtime.test.ts | Covers degraded health reporting, skipped synchronization, unchanged strict lifecycle refusal, and preservation of divergent evidence. |

</details>

```mermaid
flowchart TD
  A[Read linked workflow evidence] --> B{Tolerance requested?}
  B -->|No| C[Strictly verify control relationships]
  C --> D{Any mismatch?}
  D -->|Yes| E[Return invalid control evidence]
  D -->|No| F[Permit lifecycle or execution path]
  B -->|Yes| G[Verify structural and identity contracts]
  G --> H{Re-derivation divergence?}
  H -->|No| I[Read runner status and synchronize]
  H -->|Yes| J[Record divergence warning]
  J --> K[Read runner status]
  K --> L[Skip durable-state synchronization]
```
</details>

<sub>Reviews (1): Last reviewed commit: ["fix(runtime): let a read-only run report..."](https://github.com/monad-developers/ultrafuzz/commit/792e392856d268e41f93dbf7f68706752ea051f2) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=56680450)</sub>

- Knowledge Base — [CLI workflows](https://app.greptile.com/monad-foudnation/-/custom-context/knowledge-base/monad-developers/ultrafuzz/-/docs/cli-workflows.md)
</details>


aviggiano:
LGTM

PR #874, opened 2026-08-25
Accept null for unset workflow input envelope keys

### LINKED ISSUE #868: Every local run dies at render: workflow input envelope keys reject the SQL nulls its own input table returns

aviggiano:
## Summary

- declare the workflow input envelope keys `.nullish()` rather than `.optional()`, because the runner's input-table projection returns SQL null — not undefined — for every column a submission did not set
- treat null as absence everywhere absence is tested, including the local/cloud envelope discrimination filters, so a null-filled local input is not read as a cloud dispatch
- keep null from standing in for a key the selected envelope requires: a local input whose `tasks` column is null and a cloud input whose `selected_task` column is null are both still rejected

Without this, `main` cannot render a local run at all. `parseWorkflowInput` rejects the workflow's own dispatch before the first task node is constructed:

```
ultrafuzz workflow input is invalid: cloud_worker: Invalid input: expected true; task_id: Invalid input: expected string, received null; ...
```

## Two cloud-worker assertions move by one layer

`selected_task: null` on a cloud dispatch is now absence, so it is refused by the envelope ("requires cloud_worker, task_id, attempt_id, execution_generation, and selected_task") instead of reaching the handoff contract ("cloud worker selected_task ..."). Two existing assertions in `cloud-worker-handoff.test.ts` pinned the later message and are updated to the earlier one, with the reason recorded inline.

Nothing is weakened by that. The dispatch is still refused outright, there is still no fallback to a compiled spec, and the refusal now happens one layer earlier with a message that names what is actually wrong.

## Validation

- new regression test `generated workflow input accepts the nulls its input table reads back for keys a dispatch never set`: proven load-bearing — with the source change stashed it fails (126/127 in `generated-workflow-verifier.test.js`), with it restored 127/127
- `pnpm --filter @ultrafuzz/runtime test`: supporting 753/753, the four `runtime.test.js` shards 61+51+71+62 executed with 0 failures, Bun adapter contract 40 pass / 1 skip / 0 fail
- `pnpm --filter @ultrafuzz/artifacts test`: 322/322 on rerun; the first run showed only the known pre-existing flake `a runtime document read rejects an atomic path replacement during its snapshot`, which reproduces 8/10 times on unmodified `main` and is unrelated to this change
- `pnpm -w lint` clean, `pnpm -w format:check` clean

Closes #868

<h3>Greptile Summary</h3>

The PR aligns generated workflow validation with the runner’s SQL-backed input projection while retaining strict local and cloud envelope requirements.

- Accepts null for unset projected envelope columns and consistently treats null as absence during envelope discrimination.
- Continues rejecting null for fields required by the selected local or cloud envelope.
- Adds regression coverage for projected local and cloud inputs and updates cloud-handoff assertions for earlier rejection.

The PR appears safe to merge with null handling consistently enforced across schema validation and envelope discrimination.

Unset projected columns are accepted as null, while required local and cloud fields remain non-null requirements and downstream optional prompt handling safely normalizes null.

| Filename | Overview |
|----------|----------|
| packages/runtime/src/templates/smithers/workflows/workflow.tsx | Makes projected envelope fields nullish and consistently treats null as absence while preserving required-key and mixed-envelope rejection. |
| packages/runtime/test/cloud-worker-handoff.test.ts | Updates null-handoff expectations to reflect rejection at the envelope-validation layer without permitting fallback behavior. |
| packages/runtime/test/generated-workflow-verifier.test.ts | Adds regression coverage for SQL-null-filled local and cloud inputs and verifies that required envelope values still cannot be null. |

</details>

```mermaid
flowchart TD
  A[Runner reads projected workflow input] --> B[Validate nullish flat schema]
  B --> C{Any non-null cloud keys?}
  C -->|Yes| D[Require complete cloud envelope]
  C -->|No| E[Require schema version, run ID, and tasks]
  D --> F[Cloud worker handoff validation]
  E --> G[Local workflow task rendering]
  D -->|Required key null or absent| H[Reject input]
  E -->|Required key null or absent| H
```
</details>

<sub>Reviews (1): Last reviewed commit: ["test(runtime): pin the null cloud handof..."](https://github.com/monad-developers/ultrafuzz/commit/1e8530ae981bff890022267e9357b1fcfcb73fd7) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=56690718)</sub>

**Context used:**


aviggiano:
LGTM

PR #887, opened 2026-08-25
Detect an atomic path replacement by link count, not by ctime

### LINKED ISSUE #879: Snapshot reads miss an atomic path replacement when the ctime bump lands in the same tick
## Summary

`a runtime document read rejects an atomic path replacement during its snapshot` (`packages/artifacts/test/run-documents.test.ts`) fails roughly half the time on `main`, in isolation as well as in suite. It is not a slow test with a tight budget: it is reporting a real gap in the check it exercises.

## Cause

`readRegularFileSnapshot` in `packages/artifacts/src/schema-registry.ts` compares `fstat` on the open descriptor before and after reading:

```ts
if (
  before.dev !== after.dev ||
  before.ino !== after.ino ||
  before.size !== after.size ||
  before.mtimeNs !== after.mtimeNs ||
  before.ctimeNs !== after.ctimeNs ||
  after.size !== BigInt(offset)
) {
  throw new Error(`file changed while it was read: ${filePath}`);
}
```

A rename over the path leaves that descriptor on the **original inode**. `dev`, `ino` and `size` are therefore all unchanged, `mtime` is untouched, and the only remaining evidence is the `ctime` bump from the unlink of the original name. The kernel records `ctime` at timestamp granularity, not instruction granularity, and open/read/rename/fstat routinely complete inside a single tick on an NVMe disk. When they do, the replacement goes undetected and the read returns content for a path that no longer refers to that inode.

The flakiness is the tell: the check is correct only when it wins a race with the clock.

## Fix

Compare `nlink` as well. The rename unlinks the original name, dropping that inode's link count to zero, which is exact and carries no dependence on the clock.

`VALIDATOR_BUILD_IDENTITY` digests the compiled validator modules, so touching `schema-registry.ts` moves it by design; both pinned copies (in `packages/artifacts/test/schema.test.ts` and `packages/runtime/test/runtime.test.ts`) need updating. The assertion is a tripwire that a validator change must be deliberate, and it still is.

## Note

This fix was the second commit of #820, which the maintainer closed because its first commit — the hand-rolled secret-scanning change — was superseded by #865. The `nlink` fix is unrelated to secret scanning and is re-filed here on its own.

aviggiano:
> **Maintainer decision, not merge-on-green.** This rotates `VALIDATOR_BUILD_IDENTITY`. Please do not merge without weighing that; see "The identity rotation" below.

## Summary

- detect an atomic path replacement during a snapshot read by comparing the open inode's link count, instead of depending on the ctime bump landing in a different tick
- run the existing regression test's race 24 times instead of once, so it stops being a coin flip
- update both pinned copies of `VALIDATOR_BUILD_IDENTITY`, which this change necessarily rotates

`readRegularFileSnapshot` compares `fstat` on the open descriptor before and after reading. A rename over the path leaves that descriptor on the **original inode**, so `dev`, `ino` and `size` are all unchanged, `mtime` is untouched, and the only remaining evidence is the ctime bump from the unlink of the original name. The kernel records ctime at timestamp granularity, not instruction granularity, and open/read/rename/fstat routinely complete inside a single tick. When they do, the replacement goes undetected and the read returns content for a path that no longer refers to the inode it read.

This is the trusted read path for run documents, sealed manifests and schema files, so the miss is not cosmetic. `nlink` closes it exactly: the rename unlinks the original name, dropping that inode's link count to zero, which carries no dependence on the clock.

## The flaky test was the bug reporting itself

`a runtime document read rejects an atomic path replacement during its snapshot` fails roughly half the time on `main`. It is not slow — it is correct only when it loses a race. Measured on this box, unmodified `main`: **8 failures in 10 runs**.

A single attempt therefore proves very little either way, so the test now runs the race 24 times against fresh paths and requires every attempt to be caught. That makes it deterministic in both directions, which is what a regression test for a clock-dependent check has to be.

## The identity rotation

`VALIDATOR_BUILD_IDENTITY` hashes five compiled modules — `json-file-validator`, `json-schema-validator`, `json-validation-worker`, `schema-registry`, `strict-json` — plus the `ajv`/`ajv-formats` versions. `readRegularFileSnapshot` lives in `schema-registry.ts`, so **any** fix to it rotates the identity. There is no version of this change that avoids it: moving the function to a new module still changes `schema-registry.js`, and changes the module set as well.

```
old  ultrafuzz-json-validator.v1:77b2461a78ff8a1e942a70d473460dab54158e204dba791e818717cc9674a6d4
new  ultrafuzz-json-validator.v1:028be3251e9ac213ad6e1c037d8da47c9d903e8be563c8f4fa2149839565bcab
```

This is a deliberate consequence of a security fix, not incidental churn. The tripwire is doing exactly its job — saying that a validator change must be a decision — and this is that decision. I am aware a previous cleanup pass reverted cosmetic edits to this module specifically to avoid rotating the identity; that was right for cosmetics and is why this PR is flagged for a maintainer rather than merged on green.

Both pinned copies are updated in this PR so it does not land red:

- `packages/artifacts/test/schema.test.ts`
- `packages/runtime/test/runtime.test.ts`

## The sealed-bundle compatibility question

The runtime pin's message — *"a compatibility-only bundle loader must retain the pre-upgrade validator identity"* — asks a real question, so here is what I found.

`artifactContractSchemaBinding` builds `validator_build` from the **running process's** `VALIDATOR_BUILD_IDENTITY`; it is not a stored constant. `assertPlannedGraphSemantics` then rejects any persisted output whose `validator_build` differs from the running one. Its `allowHistoricalSchemaBundle` escape hatch — the one `assertSealedPlannedGraph` opts into, and therefore the one `readPlannedGraphDocument` uses for every run's `graph.json` — covers `schema_bundle_sha256` **only**. There is deliberately no escape hatch for `validator_build`, in the sealed loader or anywhere else.

Two consequences:

1. **Nothing in the repository needs to retain the old digest.** Searching the tree for `77b2461a` finds exactly the two test pins and nothing else — no fixture, no schema, no sealed snapshot. In `schema.test.ts` the literal is used twice: as the pin, and as the `validator_build` of the historical-bundle fixture. Moving both together is what preserves the property under test: `assertPlannedGraph` still rejects the historical bundle digest, `assertSealedPlannedGraph` still accepts it, and the sealed loader still refuses a historical *validator* build. Leaving either literal stale would silently change what that test asserts.

2. **Runs sealed by a pre-rotation build cannot be resumed by a post-rotation build.** Their `graph.json` records the old `validator_build`, and `readPlannedGraphDocument` will refuse it. This is by design and is true of every merged change to any of the five modules, but it is a real operational cost and it is the thing worth weighing here. It does not affect completed runs or new runs — only in-flight ones across the upgrade.

I am confident about (1): it is a direct reading of the code plus an exhaustive search. I am confident about the mechanism in (2); what I cannot assess from here is how many in-flight runs that would actually strand in practice.

## My recommendation

Land it. The cost is a one-time, bounded resume break for runs in flight across the upgrade — the cost the tripwire is designed to make visible, not to forbid. The benefit is not "a flaky test goes green": the check as written misses the replacement it exists to catch **8 times out of 10** on this hardware, in the trusted read path for sealed control documents. A one-comparison, clock-independent fix to that is worth a deliberate rotation. But it is a maintainer's call, so this PR does not assume it.

## Validation

- **Load-bearing proof.** With the source change reverted and rebuilt, the strengthened test fails **12 out of 12** runs. With it restored, it passes **12 out of 12**. Before strengthening, the test failed 8 of 10 runs on unmodified `main` — the same bug, reported as flakiness.
- `pnpm --filter @ultrafuzz/artifacts test`: **322 tests, 322 pass, 0 fail**
- `pnpm --filter @ultrafuzz/artifacts run schema:check`: clean
- `pnpm -w lint`: clean. `pnpm -w format:check`: clean
- The full runtime suite was **not** run here — long invocations are killed on this box, so CI is the authority for it. The one runtime test the pin change touches was run directly and passes: `artifact gates validate a historical bundle through its active sealed schema snapshot`.

## Provenance

This was the second commit of #820, which was closed because its first commit — hand-rolled secret scanning — was superseded by #865. The `nlink` fix is unrelated to secret scanning, so it is re-filed on its own.

Closes #879

<h3>Greptile Summary</h3>

The PR strengthens trusted regular-file snapshot reads by detecting link-count changes, makes the atomic-replacement regression test repeat deterministically, and rotates the validator build identity pins to reflect the compiled validator change.
- Compares `nlink` before and after reading an open inode to detect rename-based path replacement without relying on timestamp granularity.
- Runs the replacement race against 24 fresh path pairs.
- Updates both validator identity assertions for the newly compiled validator.

The PR appears safe to merge, subject to the explicitly documented maintainer decision about validator identity rotation.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| packages/artifacts/src/schema-registry.ts | Adds a link-count consistency check to the trusted snapshot reader, closing the same-timestamp atomic-replacement gap. |
| packages/artifacts/test/run-documents.test.ts | Reworks the atomic-replacement regression test to exercise 24 fresh files and verify every replacement occurs during the read. |
| packages/artifacts/test/schema.test.ts | Updates the artifacts validator identity pin to match the compiled schema-registry change. |
| packages/runtime/test/runtime.test.ts | Updates the runtime compatibility test’s validator identity assertion to the rotated build identity. |

</details>

```mermaid
flowchart TD
  A[Open regular file without following symlinks] --> B[Capture initial fstat]
  B --> C[Read bytes from open descriptor]
  C --> D[Atomic rename may replace pathname]
  D --> E[Capture final fstat on original inode]
  E --> F{Identity, timestamps, size, and nlink unchanged?}
  F -->|No| G[Reject changed snapshot]
  F -->|Yes| H[Return captured bytes]
```
</details>

<sub>Reviews (2): Last reviewed commit: ["fix(artifacts): detect an atomic path re..."](https://github.com/monad-developers/ultrafuzz/commit/70b99daf62b5e4a65000789e983c1f9ea0aa5a4f) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=56731438)</sub>

**Context used:**

- Knowledge Base — [Artifact management](https://app.greptile.com/monad-foudnation/-/custom-context/knowledge-base/monad-developers/ultrafuzz/-/docs/artifact-management.md)


aviggiano:
LGTM

Issue #792, opened 2026-08-24
Remove the `full` audit profile and merge its differential/stateful lanes into `exhaustive`


aviggiano:
## Context

Maintainer decision (2026-08-24, corrected): v0.1.0 ships exactly five audit profiles — `default`, `exhaustive`, `invariant-only`, `low-cost`, `smoke`. The `full` profile is removed and its deep differential/stateful/dynamic specialist lanes **merge into `exhaustive`**. `default` stays exactly as it is today.

> Correction: the first version of this issue said the lanes should fold into `default`. Wrong — `default` does not run differential tests or stateful fuzzing, and should keep not doing so.

History: #708 (v0.0.21) established the five-profile lineup by removing `thorough`; PR #621 reintroduced a sixth profile `full` (packages/config/audit-profiles.yml + packages/config/topologies/full.yml) a day later during the release/v0.1.0 reconciliation. `full` and `default` were at one point synonyms; we keep only `default` to avoid confusion — and `full`'s extra lanes become what distinguishes `exhaustive`.

## Tasks

- [ ] Give `exhaustive` the deep differential/stateful/dynamic specialist lanes from `topologies/full.yml`. Today `exhaustive` = project topology with max settings (strategy_loops=5, unlimited dynamic enumerator, 8 agents/16 nodes, 86400s deadline); after this it also schedules the specialist lanes — e.g. via an exhaustive topology that extends the project topology while keeping those settings
- [ ] `default` unchanged — no new lanes
- [ ] Remove the `full` entry from `packages/config/audit-profiles.yml`
- [ ] Delete `packages/config/topologies/full.yml`
- [ ] Update docs/reference/audit-profiles.md (profile table + the "Ultrafuzz ships full, smoke, and invariant-only topology files" line) and the `ultrafuzz topology list/show/copy` surfaces and tests

## Acceptance

- An exhaustive-profile run schedules differential + stateful lanes
- A default-profile run's topology is identical to today's
- `ultrafuzz --audit-profile full` fails with a clear unknown-profile error
- audit-profiles.yml and docs list exactly the five profiles; full.yml is gone
- CI green

Pitch note: after this lands, slide 8's diagram (differential + invariant lanes in one run) describes the `exhaustive` workflow; `default` remains threat-model + property/strategy lanes.

aviggiano:
Corrected per maintainer: the lanes merge into `exhaustive`, not `default` — `default` keeps its current topology unchanged. Title and body updated accordingly.

Issue #648, opened 2026-08-18
simplify audit_profile


aviggiano:
- merge thorough and exhaustive: we need a single 'max' profile, not 2; I think 'exhaustive' is the best term
- remove fuzz-only: this was useful for v0.x vs v0.1 comparison, but once we show that v0.1 is clearly better we just deprecate 'fuzz-only'; we will still keep 'invariants-only' since it can be useful for actually authoring a stateful fuzzing campaign, which can show more bugs as it is manually developed

aviggiano:
**This looks already done — superseded by merged PR #651** (`fix(config): simplify audit profiles`, commit `7a064638` on `release/v0.1.0`). Recommending closure; flagging rather than closing since it is your issue and the "deprecate vs remove" wording below is a judgement call.

Both asks are satisfied at `origin/release/v0.1.0`:

1. **"merge thorough and exhaustive: we need a single 'max' profile, not 2; 'exhaustive' is the best term."** `packages/config/audit-profiles.yml` now declares exactly five profiles — `smoke`, `low-cost`, `default`, `exhaustive`, `invariant-only`. There is no `thorough`, and the surviving name is `exhaustive`, which is the term you picked.
2. **"remove fuzz-only ... we will still keep 'invariants-only'."** No `fuzz-only` profile remains, and `packages/config/topologies/` is down to `full.yml`, `invariant-only.yml`, `smoke.yml` — the `fuzz-only.yml` topology is gone too, not just the profile entry. `invariant-only` is retained, exactly as you asked.

One thing to confirm before closing, because it is the one place the shipped state may differ from your intent: you wrote *"we just **deprecate** 'fuzz-only'"*, and #651 **removed** it outright rather than deprecating it with a warning path. If a deprecation shim mattered for anyone with `fuzz-only` in an existing config, that is a separate follow-up; if not, this is fully done.

Found while sweeping for issues made obsolete by PRs #689/#690 — worth recording that this supersession comes from upstream #651, not from either of those PRs. It surfaced because #689 had to drop its own `fuzz-only.yml` topology change as moot for exactly this reason.

Issue #534, opened 2026-08-12
Rename the 'balanced' audit profile to 'default' with no alias or fallback


aviggiano:
## Problem

The default audit profile is named `balanced`, and which profile is the default is expressed by a
separate pointer key. `packages/config/audit-profiles.yml` currently opens with:

```yaml
schema_version: 1
default: balanced

profiles:
  balanced:
    description: Recommended default for normal audits.
    intended_use: General-purpose audits using the editable project topology.
    settings: {}
```

Two problems follow from that shape:

1. **The name carries a claim the profile does not make.** `balanced` implies a chosen point on a
   cost/depth tradeoff, but the profile declares `settings: {}` and no `topology_path` — it is
   literally "no overrides, use the project topology". It is the default, not a balance point. The
   other profile names (`smoke`, `low-cost`, `thorough`, `exhaustive`, `invariant-only`) all describe
   real deviations from it, which makes `balanced` the odd one out.
2. **Two things can disagree about what the default is.** The `default:` pointer and the profile
   identity are independent, so the catalog can name a default that means something other than
   "unmodified". Every consumer then has to read `catalog.defaultProfile` to find out.

## Proposal

Rename the profile to `default`, and make being the default a property of the name rather than of a
pointer.

Recommended shape — drop the `default:` key entirely and resolve the default as "the profile whose
id is `default`", failing closed if the catalog has no such profile:

```yaml
schema_version: 2

profiles:
  default:
    description: Unmodified project workflow with no profile overrides.
    intended_use: General-purpose audits using the editable project topology.
    settings: {}
```

This removes the indirection: there is exactly one way to be the default and it cannot drift. The
lower-churn alternative is to keep the pointer and write `default: default`, which works but
preserves the ability for the two to disagree and reads poorly. Recommending the first.

**No fallback and no alias.** `audit_profile = "balanced"` must be rejected with the existing
unknown-profile diagnostic that lists the available profiles. It must not be silently mapped to
`default`, and there must be no deprecation window that accepts both. A silent alias would leave the
old name in project configs and eval provenance indefinitely, which is the thing this change is meant
to end.

## Breaking change

Any project whose `ultrafuzz.toml` pins `audit_profile = "balanced"` will fail validation after this
lands and must be edited to `"default"` or have the key removed. This needs a `## Breaking changes`
section in the release notes for the version that ships it, per
`.agents/skills/ultrafuzz-release-skill/SKILL.md` step 3.

[...]

PR #936, opened 2026-08-26
fix(config)!: remove dead run.max_parallel_nodes concurrency key

### LINKED ISSUE #933: Remove dead run.max_parallel_nodes config key; max_parallel_agents is the only enforced concurrency limit
## Summary

`run.max_parallel_nodes` is parsed, defaulted, schema-validated, documented,
environment-overridable, and set to distinct values by four audit profiles — and
it is read by nothing. It has no effect on how a run executes. Only
`run.max_parallel_agents` reaches the scheduler.

It should be removed, and `max_parallel_agents` should remain the single
concurrency limit.

## Evidence (at `b0ffd4e2`)

`max_parallel_agents` reaches the scheduler:

```
packages/runtime/src/plan-run.ts:336
  requestedConcurrency: input.maxConcurrency ?? resolved.config.run.maxParallelAgents
packages/runtime/src/start-run.ts:728
  const requestedConcurrency = input.maxConcurrency ?? sealedConfig.run.maxParallelAgents
packages/runtime/src/workflow-control.ts:163-164
  const requestedConcurrency = Math.max(1, previous.concurrency?.requested_concurrency ?? 1)
  const availableSlots = Math.max(0, requestedConcurrency - activeWork)
```

`max_parallel_nodes` reaches nothing. Every reference outside `packages/config/`
either writes the key into a generated config file or documents it:

- `packages/config/src/{types,defaults,loader,resolve,resolved-config-schema,audit-profiles}.ts` — plumbing only
- `packages/config/schema/resolved-config.schema.json` — declared required
- `packages/config/audit-profiles.yml` — set to 4 / 4 / 16 / 8 across profiles
- `packages/modal/src/workspace-config.ts` — emits the key into a template
- `packages/evmbench/src/adapter.ts` — rewrites the key in a config
- `docs/SPECS.md`, `docs/reference/configuration.md`, `docs/reference/audit-profiles.md`, `docs/tutorials/first-campaign.md`

`git grep maxParallelNodes -- packages/runtime` returns nothing.

## Why this is worth fixing rather than leaving alone

A key that validates cleanly and silently does nothing is worse than an absent
key. `docs/reference/configuration.md:120` describes it as a "positive graph
planning parallelism limit", and the profiles give it plausible per-profile
values, so an operator tuning throughput reasonably concludes they have two
independent limits. They set `max_parallel_nodes` higher, observe no change, and
have no way to discover why short of reading the runtime source.

This surfaced during a long campaign where a run was executing at 6 concurrent
tasks with `max_parallel_nodes = 8` configured; the effective limit was
`max_parallel_agents` alone.

There is also a naming problem worth settling in the same change: the surviving
key is called `max_parallel_agents`, but it bounds *all* tasks, not just agent
tasks. Verification and preparation lanes — local, cheap, non-LLM — consume the
same slots as LLM agent lanes. An operator on an idle 32-core host who wants
verification lanes to run wider cannot express that without also raising LLM
concurrency against provider rate limits. Removing the dead key does not fix
that, but it should be recorded: either rename to something honest
(`max_parallel_tasks`), or genuinely split the two limits. Please decide which
before implementing, since it changes whether this is a pure deletion.

## Proposed change

1. Delete `max_parallel_nodes` / `maxParallelNodes` from config types, defaults,
   loader, resolver, Zod schema, and the JSON schema's `required` lists.
2. Drop it from all four entries in `packages/config/audit-profiles.yml`.
3. Drop `ULTRAFUZZ_MAX_PARALLEL_NODES` from the environment-override table.
4. Stop emitting it from `packages/modal/src/workspace-config.ts` and
   `packages/evmbench/src/adapter.ts` — `evmbench` currently writes
   `profile.max_concurrency` into it, so that value must move to
   `max_parallel_agents` or the profile's concurrency intent is silently lost.
5. Update the four docs files and `packages/config/test/audit-profiles.test.ts`.

## Compatibility

This is the part to get right. Existing `ultrafuzz.toml` files in the wild
contain `max_parallel_nodes`. If the loader becomes strict about unknown keys,
removing it turns every such file into a hard validation error on upgrade — a
worse outcome than the dead key.

Preference: accept the key for one release, ignore it, and emit a deprecation
diagnostic naming `max_parallel_agents` as the real limit. Only then delete the
compatibility shim. If the loader already ignores unknown keys silently, a
straight deletion is fine and this paragraph is moot — please confirm which
before implementing.

aviggiano:
Closes #933

## ⚠️ This is a breaking config change

`run.max_parallel_nodes` and the `ULTRAFUZZ_MAX_PARALLEL_NODES` environment override are **deleted outright**. There is no deprecation shim, no accept-and-ignore path, and no migration.

**The loader is strict about unknown keys.** I confirmed this in the code before writing anything: `parseProjectConfigToml` calls `collectUnknownKeys` for every table, and that helper builds its diagnostic with the `diagnostic()` constructor, which hardcodes `severity: "error"`. The function ends with `if (hasErrors(diagnostics)) return fail(diagnostics)`.

So the concrete consequence is: **any existing `ultrafuzz.toml` that still sets `max_parallel_nodes` will now fail to load**, with

```
CONFIG_UNKNOWN_FIELD: unknown config field run.max_parallel_nodes
```

**The fix for operators is to delete the line.** It never did anything. Release notes should call this out as breaking.

`ULTRAFUZZ_MAX_PARALLEL_NODES` is simply inert now — environment variables are not subject to the unknown-key check, so setting it is silently ignored rather than an error.

## Why the key is going away

`run.max_parallel_nodes` was parsed, defaulted, schema-validated, documented, environment-overridable, and set to four distinct values across the audit-profile catalog — and no runtime code ever read it. `git grep maxParallelNodes -- packages/runtime` returns nothing. Only `run.max_parallel_agents` reaches the scheduler, so an operator raising `max_parallel_nodes` to widen throughput observed no change and had no way to discover why short of reading the runtime source.

A key that validates cleanly and silently does nothing is worse than an absent key. After this change `run.max_parallel_agents` is the only concurrency limit in the contract.

## What changed

Removed from:

- `packages/config/src/{types,defaults,loader,resolve,resolved-config-schema,audit-profiles}.ts`
- `packages/config/schema/resolved-config.schema.json` — dropped from both the camelCase `run` `required` list and the snake_case `auditProfileSettings` block, and from the `auditProfileSettingName` enum
- `packages/config/audit-profiles.yml` — all four profile entries
- `ultrafuzz.toml` — the packaged default config
- `packages/modal/src/workspace-config.ts` and the now-unused `DEFAULT_MODAL_MAX_PARALLEL_NODES`
- `packages/evmbench/src/adapter.ts`
- `docs/SPECS.md`, `docs/config.md`, `docs/reference/configuration.md`, `docs/reference/audit-profiles.md`, `docs/tutorials/first-campaign.md`

### The persisted resolved-config contract

`maxParallelNodes` / `max_parallel_nodes` are gone from the sealed resolved-config schema entirely, `required` lists included. Both validators are strict, so a resolved-config artifact sealed by an **older** build — one that still carries the key — will not re-validate on a newer binary. That is a deliberate consequence of the no-shim decision, not an oversight; flagging it because it affects resuming a run that was started before the upgrade. The schema version is left at `resolved-config.v3` rather than bumped to v4.

The `resolved-config.valid.pre-same-agent-attempts.json` regression fixture had to have the key stripped for the same reason.

### evmbench keeps its concurrency intent

`adapter.ts` was writing `profile.max_concurrency` into *both* keys on consecutive lines:

```ts
config = replaceInteger(config, "max_parallel_agents", profile.max_concurrency);
config = replaceInteger(config, "max_parallel_nodes", profile.max_concurrency);
```

The second write was redundant with the first, so `profile.max_concurrency` already lands on `max_parallel_agents` and nothing had to be re-routed — the value is preserved exactly as before. Removing the line was also mandatory rather than cosmetic: `replaceInteger` throws `Ultrafuzz config is missing ${key}` when the key is absent, and it is now absent from the generated config.

`packages/modal/src/workspace-config.ts` stops emitting the key. `DEFAULT_MODAL_MAX_PARALLEL_AGENTS` is deliberately left at its existing value — the removed key had no effect, so raising the surviving limit to match it would be a real behaviour change smuggled into a removal PR.

## Out of scope

The issue notes that the surviving key is misnamed: `max_parallel_agents` bounds *all* tasks, including local non-LLM verification and preparation lanes, not just agent tasks. Renaming it to something honest, or genuinely splitting the two limits, is a separate decision for the maintainers and is **not** in this PR. This change is removal only.

## Remaining grep hits, all intentional

`git grep -n "maxParallelNodes\|max_parallel_nodes\|ULTRAFUZZ_MAX_PARALLEL_NODES"` returns only:

| Location | Why it stays |
| --- | --- |
| `packages/config/test/config.test.ts` (5 hits) | Two new regression tests: one asserts the key is now a hard `CONFIG_UNKNOWN_FIELD` load failure, the other asserts `ULTRAFUZZ_MAX_PARALLEL_NODES` no longer affects the resolved config. |
| `packages/config/test/fixtures/resolved-config.invalid-version.json` (3 hits) | Pre-existing fixture that must be **rejected** for its schema version. Left untouched; it stays invalid. |
| `docs/reference/configuration.md` (1 hit) | The paragraph telling operators the key was removed, that this is breaking, and to delete it. |

No source file outside tests and that one doc paragraph mentions the key.

## Regenerated / recomputed artifacts

- `docs/reference/audit-profiles.md` is generated from the catalog — regenerated with `pnpm docs:audit-profiles`.
- `resolved-config.valid.json` is asserted byte-equal to the serialization of the default resolve, so it was regenerated; its `catalogDigest` changes because `audit-profiles.yml` changed.
- The two schema digests pinned in `resolved-config-schema.test.ts` were recomputed for the edited JSON schema.

<h3>Greptile Summary</h3>

The PR removes the unused `run.max_parallel_nodes` setting and its environment override from configuration parsing, resolution, schemas, generated configurations, profile definitions, and documentation.

- Makes legacy project configurations containing the removed TOML key fail strict validation.
- Removes the field from persisted resolved-config validation while retaining the existing schema-version literal.
- Preserves EVMBench and Modal concurrency through `max_parallel_agents`.
- Updates fixtures, schema digests, profile catalog data, and regression tests.

The PR appears safe to merge because no blocking failure remains.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| packages/config/src/loader.ts | Removes the TOML key from the accepted run fields and parsing path, making legacy declarations fail strict unknown-field validation as documented. |
| packages/config/src/resolve.ts | Removes the setting from profile application, environment and runtime overrides, provenance tracking, and resolved-config serialization. |
| packages/config/schema/resolved-config.schema.json | Removes the field from the closed run schema and audit-profile setting contracts, including required lists and setting-name enums. |
| packages/config/src/resolved-config-schema.ts | Keeps the Zod resolved-config validator aligned with the edited JSON Schema. |
| packages/config/src/types.ts | Removes the field from public configuration types while deliberately retaining the resolved-config v3 version literal. |
| packages/evmbench/src/adapter.ts | Stops rewriting the removed key while continuing to apply benchmark concurrency through max_parallel_agents. |
| packages/modal/src/workspace-config.ts | Stops emitting the removed key in generated Modal target configurations. |
| packages/config/test/config.test.ts | Adds regression coverage for strict rejection of the removed TOML key and inert handling of the former environment override. |

</details>

```mermaid
flowchart LR
  A[Project TOML and environment] --> B[Strict config loader]
  B --> C[Resolved configuration]
  C --> D[Runtime scheduler]
  E[Audit profiles] --> C
  F[Modal and EVMBench generators] --> A
  G[Removed max_parallel_nodes] -. no longer accepted or emitted .-> B
  C --> H[max_parallel_agents]
  H --> D
```
</details>

<sub>Reviews (2): Last reviewed commit: ["docs(changelog): record the max\_parallel..."](https://github.com/monad-developers/ultrafuzz/commit/1dc8ec06f9486b93e6f8b5ad43d527b9fbd41399) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=57076564)</sub>


docs/reference/configuration.md
```diff
@@ -130,6 +128,15 @@ empty path components, and dot components fail validation.
...
 
+`max_parallel_agents` is the only concurrency limit the runtime enforces. It
+bounds every task the workflow submits, not only agent tasks.
+
+`max_parallel_nodes` was removed. It never reached the scheduler, so it had no
+effect on how a run executed. This is a breaking change: the loader rejects
+unknown keys, so a config that still sets `max_parallel_nodes` now fails to load
+with a `CONFIG_UNKNOWN_FIELD` error. Delete the key. The
+`ULTRAFUZZ_MAX_PARALLEL_NODES` environment override was removed with it.
+
```
aviggiano: remove it form here, changelog will have this

aviggiano (commented):
the deprecation should be on CHANGELOG

other than that, LGTM

aviggiano:
Both change requests applied, rebased onto current `main`.

## 1. > the deprecation should be on CHANGELOG
>
> other than that, LGTM

Added to `CHANGELOG.md` under `## Unreleased` → `### Breaking changes`, as the fourth entry alongside the `#792` / `#534` / `#648` profile removals and in their format (`- **[area] [area]** …` tags, present-tense description, issue reference last):

- **[config] [docs] [modal] [evmbench]** Removes the dead `run.max_parallel_nodes` key and its `ULTRAFUZZ_MAX_PARALLEL_NODES` environment override. … (#933)

The entry states what happens to a config that still sets the key, verified against the code rather than assumed — and the two paths differ:

- **TOML key: hard load failure.** `parseProjectConfigToml` calls `collectUnknownKeys` for the `run` table, that helper builds its diagnostic with `diagnostic()` which hardcodes `severity: "error"`, and the function ends with `if (hasErrors(diagnostics)) return fail(diagnostics)`. So an `ultrafuzz.toml` that still sets `max_parallel_nodes` fails to load with `CONFIG_UNKNOWN_FIELD` on path `run.max_parallel_nodes`. Operators delete the line. No shim, no warn, no accept-and-ignore. Regression test: `packages/config/test/config.test.ts` → "rejects the removed run.max_parallel_nodes key outright".
- **Environment override: silently ignored.** Environment variables are not subject to the unknown-key check, so `ULTRAFUZZ_MAX_PARALLEL_NODES` is simply inert. Regression test: "no longer honours ULTRAFUZZ_MAX_PARALLEL_NODES".

It also records the two consequences an operator cannot see from the diff: `run.max_parallel_agents` is now the only concurrency limit in the contract (and it bounds every submitted task, not only agent tasks), and the key is gone from the sealed resolved-config schema including its `required` lists, so a `config.resolved.toml` sealed by an older build no longer re-validates on a newer one (schema version deliberately left at `resolved-config.v3`).

## 2. > remove it form here, changelog will have this
> (`docs/reference/configuration.md` line 139)

Deleted the deprecation paragraph from the Run section of `docs/reference/configuration.md`:

```
`max_parallel_nodes` was removed. It never reached the scheduler, so it had no
effect on how a run executed. This is a breaking change: the loader rejects
unknown keys, so a config that still sets `max_parallel_nodes` now fails to load
with a `CONFIG_UNKNOWN_FIELD` error. Delete the key. The
`ULTRAFUZZ_MAX_PARALLEL_NODES` environment override was removed with it.
```

One judgement call to flag, since the comment anchored to the end of the whole added block: I kept the paragraph immediately above it —

> `max_parallel_agents` is the only concurrency limit the runtime enforces. It bounds every task the workflow submits, not only agent tasks.

— because it documents current behaviour of a surviving key rather than the removal, which is reference-page material and not changelog material. Say the word and it goes too.

The page's structure is intact: the Run key table and the surrounding `max_dynamic_nodes` / `keep_workspaces` paragraphs are untouched, no dangling heading, row, or list item. `docs/reference/configuration.md` is not generated — `scripts/docs-check.mjs` only asserts its existence and the only generator, `scripts/audit-profile-docs.mjs`, owns `docs/reference/audit-profiles.md` — so no generator change was needed.

No source or doc file mentions the key any more. `git grep -n "max_parallel_nodes\|maxParallelNodes\|MAX_PARALLEL_NODES"` returns only the new `CHANGELOG.md` entry, the two regression tests, and the pre-existing `resolved-config.invalid-version.json` fixture that must stay invalid.

## Validation

Rebased onto `origin/main` (`fca2551b`, after today's fleet) — no conflict, `CHANGELOG.md` included; the new entry sits under the same `### Breaking changes` heading as the entries that landed today.

| Check | Result |
| --- | --- |
| `pnpm docs:check` | pass (audit-profile drift check + docs check) |
| `pnpm format:check` (whole tree) | pass |
| `pnpm --filter @ultrafuzz/config test` | 74/74 pass, 4 files |
| `pnpm --filter @ultrafuzz/evmbench test` | 62/62 pass, 6 files |
| `pnpm --filter @ultrafuzz/modal test` | 629/629 pass, 34 files |
| runtime, the one assertion this PR edits | `audit profile selects its packaged topology and records portable provenance` — pass |
| runtime PR smoke (`scripts/run-pr-smoke-tests.mjs`) | 41/41 pass under `node --test`; the Bun leg ran `4 pass / 0 fail` but the script's own `(pass) <name>` string match did not find those lines in this sandbox's Bun output, so the wrapper exits 1. Pre-existing local harness/Bun-version quirk, not caused by this PR — it touches no runtime source, only that one test assertion. CI's runtime job is authoritative here. |

The 16 initial `@ultrafuzz/modal` failures were `Failed to resolve entry for package` — an unbuilt workspace, not a real failure; after `pnpm --filter @ultrafuzz/modal... build` all 34 files pass.

Issue #955, opened 2026-08-26
Upgrade the pinned Smithers runtime to 0.35.0


aviggiano:
## Problem

Ultrafuzz currently pins `smthrs` and the generated workflow dependency closure to 0.34.0. Smithers 0.35.0 includes recovery and durability fixes that generated controllers and execution snapshots cannot use until the exact pin, dependency cutoff, and compatibility surface are upgraded together.

In particular, 0.35.0 ignores stale continuation pointers from failed or cancelled attempts, preventing a resumed node from spending retry budget on a conversation that no longer exists. The upgrade must preserve in-place continuation of existing durable 0.34.0 runs rather than requiring a new Ultrafuzz or Smithers run ID.

## Required changes

- Bump `SMITHERS_VERSION` and the generated runner manifest to `smthrs@0.35.0`.
- Align `effect` and the complete `@effect/*` override family with the exact versions required by Smithers 0.35.0.
- Record the new package publication timestamps and move the deterministic npm resolution cutoff past every new pin.
- Re-evaluate Ultrafuzz Smithers compatibility patches and remove only those that are demonstrably upstream in 0.35.0; retain and test any still required by the generated runtime.
- Regenerate affected manifests, fixtures, documentation, and sealed dependency expectations.
- Validate Smithers database migrations and rollback/error behavior against a stopped 0.34.0 run before mutating it.

## Acceptance criteria

- [ ] Fresh generated workflows install and execute with the exact Smithers 0.35.0 dependency closure.
- [ ] `doctor`, manifest validation, and dependency-resolution-cutoff checks report the new exact pins.
- [ ] A stopped durable 0.34.0 workflow resumes in place under 0.35.0 with the same Ultrafuzz run ID and Smithers run ID.
- [ ] A completed producer remains completed and byte-identical while only unfinished or newly rendered downstream tasks execute after resume with an accepted workflow change.
- [ ] Failed and cancelled attempts do not contribute stale agent continuation pointers to a later attempt.
- [ ] Existing attempt history, checkpoints, artifact manifests, and finalization evidence remain readable and are not rewritten merely by the dependency upgrade.
- [ ] `resume --refresh-controller` seals an authenticated 0.35.0 execution snapshot and does not silently fall back to an older installed Smithers tree.
- [ ] Focused resume, retry, pause/cancel, workflow-change, generated-controller, and compatibility-patcher tests pass.

## References

[...]

PR #995, opened 2026-08-28
fix(runtime): scrub stale continuation pointers on a resumed dispatch

aviggiano:
## Problem

A resumed activation is a **new controller process re-dispatching the same attempt number**. Smithers 0.34 derives that dispatch's continuation pointer from the interrupted attempt's heartbeat (`@smthrs/engine/src/engine.js` ~5975-6010):

- `resumeSession` — a session id in the agent CLI's own local store.
- `continueSession` — the fallback when no id was captured, which means `--continue the latest session in this worktree`. Upstream's own comment flags the hazard: *"if the worktree is shared by concurrent tasks, --continue is cwd-scoped and may attach the most recent session."*

Neither pointer survives a resume: the controller process is new, and the worktree may have been restored on another machine, so the referenced session is gone. `continueSession` is the worse of the two — being cwd-scoped rather than id-scoped, after a workspace restore it can attach the task to whatever conversation happens to be most recent in that worktree, which need not be this task's.

`artifactAwareAgent` already scrubbed `resumeSession`, `continueSession` and `lastHeartbeat` — but only for `smithersAttempt >= 2`:

```ts
if (smithersAttempt <= 1 || !firstGenerationForAttempt) return args;
```

**Resume reuses attempt numbers**, so a resumed dispatch of attempt 1 fell straight through the gate and the dead pointer reached the adapter.

## Change

The pointer scrub moves *above* the attempt gate, so it runs on the first generation of any attempt in this controller process:

- **Attempt 1, first generation** — only the three pointers are dropped. `attemptedGenerations` is a fresh `Set` per process, so this branch is reached exactly once per attempt per controller. On a fresh run no pointer exists and the scrub is inert; on a resume it is the fix.
- **Attempt >= 2, first generation** — unchanged: pointers dropped *and* the error-agnostic reset of `messages`/`prompt` back to the original.
- **Later generations in the same attempt** — untouched. These are Smithers' correction turns against a session *this* process opened, so their pointer is live and must be preserved. Covered by the existing "artifact-aware agents preserve Smithers continuation and checkpoint capabilities" test, still green.

### What a resume deliberately keeps

Only the *pointers* are dropped. `messages` (Smithers' checkpoint conversation) and `resumeCheckpoint` are portable state Smithers stored itself, not references into a CLI session store. The engine passes them alongside the pointer, so keeping them means the resumed agent continues from the replayed transcript rather than restarting from scratch. This also protects the fork-seeding path (`forkSeedMessages`), which arrives as `messages` on a first generation and would have been destroyed by a blanket reuse of the retry branch.

## Test

`a resumed activation's first dispatch carries no stale continuation pointer` in `packages/runtime/test/generated-workflow-verifier.test.ts` builds a fresh wrapper (a new controller process), dispatches attempt 1 carrying a stale heartbeat, and asserts the adapter sees no pointer — for **both** shapes: a captured `resumeSession` id and the cwd-scoped `continueSession: true` fallback. It also asserts the checkpoint `messages` and effective prompt survive, and that a following correction turn in the same attempt keeps its live pointer.

Verified non-vacuous: with the source change reverted, the new test fails with `actual: 'interrupted-session'`.

The source-text guard in `generated retries do not inspect or inject previous failure text` is extended to assert the ordering — that the pointer scrub is constructed *before* `if (smithersAttempt <= 1)` — so a future edit cannot quietly re-gate it behind the attempt number.

## Relationship to #955

This is a **partial, pre-upgrade mitigation on the Ultrafuzz side. It is not a replacement for the upstream fix in Smithers 0.35.0** and #955 should stay open.

What this covers: the args actually handed to the agent adapter. A dead session id or a cwd-scoped `--continue` no longer reaches the CLI on the first dispatch after a resume.

**What remains broken until 0.35.0 lands**, because it happens inside the engine *before* the workflow's wrapper is called and the wrapper cannot reach it:

1. **Engine-side hijack selection.** `shouldAutoHijack = desc.onHijackExit === "reopen" || !priorContinuation` (engine.js:6082). A stale `priorContinuation` from the interrupted attempt still suppresses auto-hijack, and the workflow rewriting the adapter args afterwards does not undo that choice.
2. **The empty-prompt guard is satisfied by a dead pointer.** `TASK_EMPTY_PROMPT` (engine.js:6238) is skipped when `resumeSession || continueSession` is set. A node whose prompt renders empty on resume passes a guard on the strength of a pointer this PR then removes, so it fails later and less legibly at the adapter instead of failing fast with the node named.
3. **Attempt and checkpoint bookkeeping.** `resumeCheckpointRefConsumed`, `discardResumeSession` and retry-budget accounting are all engine-side and still treat the dispatch as a resume of a session that no longer exists.
4. **The scope is one task kind.** Only the agent task goes through `agentForTask`/`artifactAwareAgent`. Preparation and verifier tasks carry no agent today, so there is no gap right now — but this is a wrapper-level workaround, not a systemic one, and any future agent-bearing task would need the same treatment.

0.35.0's `ignore stale continuation pointers from failed or cancelled attempts` (smithers#1619) fixes this at the derivation site, which is where it belongs. #955's acceptance criterion *"Failed and cancelled attempts do not contribute stale agent continuation pointers to a later attempt"* is **not** met by this PR — this PR addresses the adjacent resume case (an *interrupted* attempt's pointer re-fed to the *same* attempt number in a new process), which upstream's wording does not obviously cover.

## Review notes

This touches the generated workflow template, so it changes the controller emitted for every new run. Worth a careful read of the branch ordering in `artifactAwareAgent.generate`.

## Validation

- `pnpm -r build`, `pnpm --filter @ultrafuzz/runtime typecheck`, `pnpm lint`, `pnpm format:check` — clean.
- `generated-workflow-verifier.test.ts` — 133/133 pass.
- `agent-adapter-boundaries.test.ts` — 10/10 pass.

<h3>Greptile Summary</h3>

The PR updates generated runtime controllers to remove stale continuation pointers from the first dispatch of each attempt while retaining portable checkpoint conversation state.
- Moves continuation-pointer scrubbing ahead of the attempt-number gate.
- Preserves live pointers on later correction generations within the same attempt.
- Adds regression coverage for resumed attempt-one dispatches and both continuation-pointer forms.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| packages/runtime/src/templates/smithers/workflows/workflow.tsx | Moves stale-pointer removal to every first generation while preserving retry-only conversation reset and later correction continuations. |
| packages/runtime/test/generated-workflow-verifier.test.ts | Adds focused resumed-dispatch regression coverage and verifies the generated source ordering. |

</details>

```mermaid
sequenceDiagram
    participant S as Smithers
    participant W as Artifact-aware wrapper
    participant A as Agent adapter
    S->>W: First dispatch for attempt
    W->>W: Clear resumeSession, continueSession, lastHeartbeat
    W->>A: Dispatch with checkpoint messages retained
    S->>W: Later correction generation
    W->>A: Dispatch with live continuation pointer retained
```
</details>

<sub>Reviews (2): Last reviewed commit: ["fix(runtime): scrub stale continuation p..."](https://github.com/monad-developers/ultrafuzz/commit/b2837bab5b966f3e3849b549d09eaf7b09266bbf) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=57754348)</sub>


aviggiano (commented):
LGTM

PR #983, opened 2026-08-27
fix(runtime): refresh task schema bundles

aviggiano:
Fixes #982.

## Summary

- mark generated current-controller workflows as explicit schema-refresh controllers while ordinary launch controllers remain strict
- atomically rematerialize current checked-in schema bytes over older physical task-local schema files only for that refresh path
- rebind retained static and dynamic task output contract/schema metadata to the current artifacts module selected by `--refresh-controller`
- retain no-symlink, regular-file, single-link, pinned-digest, and read-only protections; ordinary schema drift still fails closed

## Production reproduction

A campaign launched at `431c0c4` retained task-workspace `report.schema.json` bytes with SHA-256 `6ee780...`. Current source after `0b8405f` is `b4987c...`. After earlier native continuation blockers were fixed, Smithers accepted and submitted the same-ID retry, then every retried property task failed in `materialize-prompt-schemas` before its agent ran. The preserved run and an isolated copy have identical historical schema bytes.

## Validation

- `pnpm --filter @ultrafuzz/runtime... build`
- runtime and artifacts test TypeScript compilation
- focused artifacts tests: refresh replacement/idempotence and hard-link rejection (2/2)
- focused runtime tests: current-controller schema rebinding and strict launch controller generation (2/2)
- runtime typecheck
- ESLint on changed TypeScript files
- repository Prettier check
- docs check

<h3>Greptile Summary</h3>

This PR allows refreshed controllers to replace stale task-local schema bundles while keeping ordinary workflow launches strict.
- Rebinds retained static and dynamic output metadata to current artifact contracts and schema identities.
- Uses durable replacement for stale physical schema files while preserving regular-file, hard-link, symlink, digest, and read-only checks.
- Adds focused coverage for refresh replacement, hard-link rejection, refreshed metadata, and strict launch behavior.

<h3>Confidence Score: 5/5</h3>

The PR appears safe to merge.

No blocking failure remains.

<details><summary><h3>Important Files Changed</h3></summary>

| Filename | Overview |
|----------|----------|
| packages/artifacts/src/schema-bundle.ts | Adds an explicitly gated durable replacement path for stale schema files while retaining existing filesystem and digest protections. |
| packages/runtime/src/smithers.ts | Marks refreshed controllers for schema replacement and consistently rebinds static and dynamic task output metadata. |
| packages/runtime/src/templates/smithers/workflows/workflow.tsx | Passes the generated controller’s refresh policy into task-workspace schema materialization. |
[...]

PR #996, opened 2026-08-28
feat(runtime): upgrade the pinned workflow engine to Smithers 0.35.0

### LINKED ISSUE #955: Upgrade the pinned Smithers runtime to 0.35.0

aviggiano:
Closes #955.

Moves the pinned workflow engine from `smthrs@0.34.0` to `0.35.0`, moves the deterministic npm
resolution cutoff to `2026-08-18T06:00:00Z`, re-anchors the six compatibility patches whose upstream
anchors moved, and widens every Ultrafuzz surface that strictly parses the runner so the release's
new fields and states are accepted instead of throwing.

---

## The Effect override family and `pnpm-workspace.yaml` are deliberately unchanged

**Read the zero diff on `pnpm-workspace.yaml` as a verified result, not an oversight.**

`smthrs@0.35.0` requires exactly the same Effect 4 tree as `0.34.0`. Scanning `effect` and every
`@effect/*` specifier across all 41 `@smthrs/*` subpackages at both releases yields exactly five
packages — `effect`, `@effect/opentelemetry`, `@effect/platform-bun`, `@effect/platform-node-shared`,
`@effect/sql-sqlite-bun` — each at a single value, `4.0.0-beta.105`, in both. `0.35.0` promotes
`@effect/platform-node-shared` from transitive to a direct dependency, but that package was already in
the override family at the same version, so nothing to add.

The issue's "align `effect` and the complete `@effect/*` override family with the exact versions
required by Smithers 0.35.0" is therefore satisfied by the pins that are already there. Verified on
the merged tree: `packages/runtime/node_modules/smthrs` is `0.35.0`, its `effect` dependency is
`4.0.0-beta.105`, and `pnpm-lock.yaml` contains **zero** occurrences of `0.34.0`.

---

## None of the 31 compatibility patches were retired; 6 were re-anchored

`SMITHERS_COMPATIBILITY_PATCHES` still has **31 entries**. Six anchors stopped matching upstream
source in `0.35.0` and were re-derived against the new bodies:

`detached_snapshot_transfer`, `supervisor_descriptor`, `resume_snapshot_transfer`,
`workflow_path_import`, `manifest_relaunch`, `engine_insert_workflow_path`.

The remaining 25 anchor unchanged.

**Why retiring the five spawn-path patches would have been a silent data-loss bug.** `0.35.0`
introduces `smithersRuntimeSpawn()` / `smithersRuntimeReentry()` in `@smthrs/cli/src/node-loader/`,
which looks like upstream finally doing what those patches do. It is only half of it. The new helper
does the bun-vs-node runtime selection; it has **no equivalent of Ultrafuzz's fd-3 execution-snapshot
descriptor transfer**. Confirmed by absence in the `0.35.0` tree:
`ultrafuzzExecutionSnapshotChildTransfer`, `ULTRAFUZZ_SNAPSHOT_INHERITED_DESCRIPTOR` and
`relaunchSnapshotTransfer` do not appear anywhere in `@smthrs/cli@0.35.0`. Dropping those patches on
the "it's upstream now" reading would have deleted snapshot-descriptor transfer from every spawn
path while every test still passed.

Also note what the new helper does under Bun, the only interpreter Ultrafuzz actually uses:
`smithersRuntimeSpawn` returns `{command: "bun", args}` — byte-for-byte what `0.34.0` hardcoded — and
`smithersRuntimeReentry` returns `{command: process.execPath, args}` — byte-for-byte what `0.34.0`'s
relaunch site hardcoded. So these are pure re-anchors that preserve exact current semantics, not
behaviour changes. Composing with the helper instead would also have reintroduced a bare-string
`"bun"` `PATH` lookup into a spawn the executable-capability layer exists to keep on a verified
absolute interpreter.

The same reasoning was applied to `terminal_state_restore` and `event_probe_index`: both workarounds
are still absent upstream (`restoreTerminalTaskStates` is not in the `0.35.0` scheduler,
`_smithers_events_insert_probe_idx` is not in `EXTRA_INDEX_STATEMENTS`).

---

### 1. The new `succeeded-with-failures` run state

`0.35.0` adds a `RunState` the issue does not mention and that Ultrafuzz's closed enum rejected
outright. It is the release's **ordinary** terminal state for a run that finished while tolerating a
child failure, so under the old enum a large class of successful runs would have failed to parse at
all. Admitted in `SMITHERS_RUN_STATES` and handled at every consuming site: the derived-status mirror
(upstream renames both successful states to `finished` for `ps` consumers), `workflowSucceeded`,
`finalRunStatus`, and the exhausted-loop reconciliation. Sealed schemas regenerated.

Widening `workflowSucceeded` matters on its own: without it, `WORKFLOW_TASK_EVIDENCE_MISSING` was
silently disarmed for exactly the run class `0.35.0` made common.

### 2. The manifest migrator, without which no scaffolded project converges

With `SMITHERS_VERSION` bumped, `assertSmithersPackageManifest` rejects every existing `0.34.0`
`.smithers/package.json`. `MIGRATABLE_PRIOR_SMITHERS_VERSIONS` generalizes that to a list of versions
this repository actually pinned, and an otherwise byte-equal prior-pin manifest is now normalised
forward in place rather than refused — so an already-scaffolded project converges onto the new pin
without `--force` regenerating the run. A manifest that differs anywhere *else* is still refused
rather than silently rewritten.

**One correction to the research plan, and it is load-bearing for AC3.** The plan called this a
blocker for resume-in-place. It is not — resume never reads the target project's
`.smithers/package.json`. `ensureSmithersDependencies`, the only caller of
`assertSmithersPackageManifest` outside `init`, runs against the operator controller's own temporary
root, into which it has just written `renderSmithersPackageJson()`. The migrator's real value is at
`init`; resume already worked. Stated this way in the CHANGELOG.

`0.33.0` was also dropped from the migratable list: `git log -p` shows `0.34.0` is the only
`SMITHERS_VERSION` this repository has ever committed, so `0.33.0` was a version we never pinned.

---

## Six strictly-parsed runner surfaces, not one

The issue and the plan both scoped "the inspect envelope". Ultrafuzz strictly parses **six** runner
surfaces, and `0.35.0` widened all six. Each was diffed `0.34.0` → `0.35.0` against the real tarballs:

| Surface | New in 0.35.0 | Handling |
|---|---|---|
| `inspect` | `data.tokenUsage`, `data.run.cancellationSource`, `data.runState.warnings` | allowlisted |
| `why` | `warnings` on **every** return path | required key; joined into the public `notes` alongside `information`, matching the runner's own rendering |
| `why` blockers | `stalled` blocker kind | added to `CURRENT_WHY_BLOCKER_KINDS`, `RunBlockerKind`, **and** `packages/cli/schema/cli-result.schema.json` — `adaptBlocker` passes the kind through verbatim and the envelope is producer-validated, so a missing enum member fails the whole `why --json` result |
| `status` | `counts.stalled` | admitted, then folded into `failed` — Ultrafuzz has no `stalled` node status and `statusFromWorkflowState` already maps it to `failed`; an 11th public count would contradict every other surface |
| `node` | `tokenUsage.freshInputTokens` | added to both node token-usage parsers; no public schema change, since `adaptNodeDetail` never exposes token counts |
| `TokenUsageReported` | `freshInputTokens`, `costUsd` | both allowlisted; `freshInputTokens` gets the token-count bound, `costUsd` a finite-non-negative bound (fractional USD, not a count) |
| event types | `NodeStalled`, `RunConcurrencySaturated` | added to the default lifecycle set and the category map |

Downstream of the new `stalled` node state: a stalled node now resets under `resume --retry-failed`
alongside failed ones (both `smithersFailedTasks` and `failedWorkflowTaskIds`), and the evmbench
adapter no longer aborts a converged run on `0.35.0`'s widened `degraded` verdict — `degraded` now
also fires on `failedChildren > 0`, and the two causes are separable because `finalRunStatus` forces
`failed` whenever `inspect.exhaustedLoops` is non-empty.

Two generated agent adapters (`deepseek`, `kimi`) passed Ultrafuzz-only options into Smithers agent
constructors, which `0.35.0` now rejects as unknown; both split those keys onto a private field
before `super()`. Enumerating every option each of the seven adapters passes shows only those two are
affected — OpenRouter inherits `CompatibleCodexAgent` and passes only CodexAgent keys.

Four forward-only SQLite migrations (0041–0044) come with the release: all `ALTER TABLE ADD COLUMN` /
`CREATE INDEX IF NOT EXISTS`, with 0043 recording `skipped: "postgres_only"`.

## The structural fix: the mirror test

`runtime.test.ts` gains a test that diffs **all six** parsed surfaces against the pinned runner's own
sources, so a future runner bump that widens an enum or an envelope fails a test instead of a
production run. Every new assertion is mutation-checked: deleting `warnings`, `stalled`,
`freshInputTokens`, `costUsd`, the `stalled` blocker kind, `NodeStalled`, `RunConcurrencySaturated`
or `steers` from the Ultrafuzz side makes it fail, and so does adding an event type the runner does
not categorise. The fake-Smithers fixtures now emit `0.35.0` documents, so every pre-existing
status/why/node/sync test exercises shapes the pinned runner can actually produce.

---

### AC1 — Fresh generated workflows install and execute with the exact 0.35.0 dependency closure
**Met for install; execution is covered only by the fake-runner harness.**
The generated manifest was rendered from `renderSmithersPackageJson()` and installed in a throwaway
directory with the real `smithersDependencyInstallArgs()` (including `--before=2026-08-18T06:00:00Z`):
`added 655 packages`, exit 0, `smthrs` `0.35.0`, exactly one `effect` root at `4.0.0-beta.105`,
`@esbuild/linux-x64` present. Re-run on the merged tree for this PR.
Honest limit: "execute" here means the full runtime/CLI suites against Ultrafuzz's fake-Smithers
harness plus the real `0.35.0` tree for patch anchoring. No real generated workflow was executed
end to end with real agent CLIs — see *Not verifiable here*.

### AC2 — `doctor`, manifest validation, and cutoff checks report the new exact pins
**Met.** On the merged tree:
- `doctor --project .` → `workflow-engine-registry: ok - pinned workflow engine 0.35.0 is the latest published stable release`, with `bundled: 0.35.0`, `required by generated project: 0.35.0`, `latest published stable: 0.35.0`. (`doctor` still exits non-zero here for environment-only reasons: missing `forge`/`kimi`/`pi` binaries and the repo root not being a project.)
- `assertSmithersResolutionCutoff()` → `ok 0.35.0 2026-08-18T06:00:00Z`, and the cutoff is in the past.
- Manifest validation: `smithers-package.test.ts`, 9/9.

### AC3 — A stopped durable 0.34.0 workflow resumes in place with the same Ultrafuzz and Smithers run IDs
**Met at the code and unit-test level; NOT proven against a real stopped 0.34.0 run.**
`start-run.ts` resolves `smithersRunId` from `metadata.workflow.run_id` and never regenerates it; the
existing same-ID continuation tests pass. The manifest blocker the plan predicted does not exist on
this code path (see above) — resume asserts the manifest only against the operator controller's own
freshly rendered temporary root. `init`-time convergence for an existing `0.34.0` project is covered
by the new prior-pin migration tests.
What is missing is an actual staged resume of a captured stopped `0.34.0` run under `0.35.0`. That
needs real agent CLIs and provider credentials; it was not faked or stubbed.

### AC4 — A completed producer stays completed and byte-identical; only unfinished or newly rendered downstream tasks execute
**NOT verified.** This needs the Tier-1 fixture described in the plan: a scrubbed stopped-`0.34.0`
run plus the post-resume ledgers from a staged `0.35.0` resume, replayed through
`packages/evals/src/recovery-equivalence.ts`. Producing those ledgers requires the same staged run
AC3 lacks. Nothing here asserts it, and nothing here pretends to.
Worth flagging for review scope: PR #983 set `replacePromptSchemas: true` for controller rendering,
so a `--refresh-controller` resume rewrites `.ultrafuzz/schemas/*.json` inside every task workspace,
completed producers included. "Byte-identical" must be scoped to the producer's declared artifact
files, verifier markers and finalization evidence.

### AC5 — Failed and cancelled attempts do not contribute stale continuation pointers to a later attempt
**Met.** This is the release's headline fix (`@smthrs/db/src/attempt-resume-pointers.js`), and it is
pinned by a new Bun-lane contract test read directly from the pinned runner —
`the pinned runner drops resume pointers only from dead attempts`: a `failed` attempt's meta loses
all five of `agentResume`, `agentConversation`, `lastHeartbeat`, `resumedFromConversation`,
`resumedFromSession`; `in-progress` and `waiting-*` keep them; a `hijackHandoff` attempt keeps them;
unparseable meta is untouched. Registered in the PR smoke lane so it gates future PRs.

### AC6 — Existing attempt history, checkpoints, artifact manifests and finalization evidence stay readable and are not rewritten by the upgrade
**Met for the durable store; the artifact/finalization half is argued, not fixture-proven.**
New Bun-lane test `pinned store migrations are additive over a 0.34.0 database` builds a
`0.34.0`-shaped SQLite store, seeds `_smithers_runs` / `_smithers_attempts` /
`_smithers_agent_checkpoints` including a failed attempt carrying all five pointer keys, runs
`0.35.0`'s migrator and asserts: 0041/0042/0044 apply and 0043 records `skipped: "postgres_only"`;
the ledger rows and checksums land; **every pre-existing attempt row's `meta_json` is byte-identical
afterwards**; a second run is a no-op; and downgrade tolerance holds, since every added column is
nullable. The upstream pointer fix strips pointers only on the transition into a non-success terminal
state and neutralises stale pointers on the read side, so it does not bulk-rewrite history.
Artifact manifests and finalization evidence are untouched by any change in this PR, but that is an
argument from the diff, not a replayed fixture.

### AC7 — `resume --refresh-controller` seals an authenticated 0.35.0 execution snapshot and does not silently fall back to an older installed tree
**NOT met. Deliberately left open rather than papered over.** Two independent reasons, both
re-verified on the merged tree:
1. `refreshedSmithersControllerSnapshot` has **no production caller** — grepping all of `packages/`
   excluding `node_modules` and tests, the only non-test hit is its own definition at
   `smithers.ts:1537`. `resume --refresh-controller` calls `assertSmithersControllerRefreshable` and
   then `renderCurrentSmithersController`, which writes an *unsealed* `.tsx` and never materializes a
   sealed execution snapshot.
2. `applyRefreshedSmithersCompatibilityPatches` throws
   `controller refresh <pkg> dependency must remain at 0.35.0` (`smithers.ts:1866`), so a
   `0.34.0`-sealed snapshot is rejected the moment `SMITHERS_VERSION` moves; and the exact-path-set
   assertion cannot hold either, because `0.35.0` adds five files under `@smthrs/cli/src/node-loader/`
   that no `0.34.0` seal contains.
So it does *not* silently fall back — it fails closed — but it also does not seal a new generation.
The fix is a design decision (relax version equality to "previous or current pin" and path equality
to "no path disappears" and re-seal, **or** declare that a `0.34.0` run continues natively under
`0.35.0` with no sealed refresh and reword the criterion). That belongs in its own change with its
own test, not smuggled into a pin bump.

### AC8 — Focused resume, retry, pause/cancel, workflow-change, generated-controller and compatibility-patcher tests pass
**Met.** Post-merge on this branch, at `umask 022`:
- compatibility patcher, 8/8 including `every runner compatibility patch still anchors in the pinned Smithers release` (reads the real installed `0.35.0`) and the new `pinned runner state and envelope contracts match Ultrafuzz's mirrors`
- `lifecycle-inspection.test.ts` 43/43, including `diagnoseProject reports a posture for every tracked compatibility patch`
- `smithers-package.test.ts` 9/9
- Bun adapter contract lane: 42 pass / 1 skip / 0 fail
- PR smoke lane (`run-pr-smoke-tests.mjs`): exit 0
- `@ultrafuzz/artifacts` 324/324, `@ultrafuzz/evmbench` 64/64, `test:ci-scripts` 109/109
- `build`, `typecheck`, `lint`, `format:check`, `docs:check`: all exit 0

Two lane fixes were needed and are included:
- **The PR smoke lane could never actually pass its Bun post-condition.** `assertBunTestsPassed`
  looked for a `(pass) <name>` line per expected test; `bun` emits a per-test line only for failures
  (verified on both 1.3.14, the `ci.yml` pin, and 1.4.0). It only looked green because a failing test
  exited the script first. It now reads the pass/fail counts off the run summary; renaming a
  registered test still makes it fire.
- **The Kimi contract test constructed the pinned `KimiAgent` with Ultrafuzz-only options** and died
  on `0.35.0`'s new option assertion. It now builds the pinned adapter from Smithers-known options
  only and pins the rejection in both directions, so the option split can be retired deliberately if
  upstream ever widens `CLI_AGENT_OPTION_KEYS`.

---

## Not verifiable in this environment

- **An end-to-end resume of a real stopped `0.34.0` durable run under `0.35.0`** — AC3's real proof
  and the prerequisite for AC4's Tier-1 fixture. Needs a real multi-package Smithers install in a
  target worktree, Bun executing a real controller, real agent CLIs and provider credentials. CI has
  none of those: dependency installs are faked, Smithers itself is faked, and release-validation
  lanes run with no provider secrets. **Recommendation: run one staged resume of a captured `0.34.0`
  run against a real `0.35.0` install before merging, commit its ledgers and a pre/post `smithers.db`
  pair as fixtures.** Then AC3/AC4/AC6 are proven once and regression-guarded thereafter.
- **AC7 has no implementation to test** (above).
- **`validate:release` was not run as a single wrapper invocation** — it re-runs the ~5h runtime suite
  serially. Its gates were run individually: `validate:pack`, `docs:check`,
  `security:dependency-advisories`, `validate-audit-profile-package`, and every package suite.
- **The Modal seed image and the evmbench overlay image were not rebuilt.** Both install the pinned
  closure; the closure install is proven above, but the images themselves need a rebuild before merge.
- **`packages/modal` and `packages/dashboard`** were typechecked and linted only; no image build and
  no cloud integration was exercised.
- One pre-existing runtime failure is unrelated to this branch and reproduces on `main`:
  `a half-published dynamic expansion stays readable while execution stays closed`, confirmed failing
  identically at the merge base `b0d2268e` in a separate worktree.
- Note for local reproduction: two OpenRouter Bun tests fail under `umask 002` because `initProject`
  creates `.ultrafuzz` group-writable and trips `assertSafeDirectory`. Everything here was run at
  `umask 022`, CI's default.

## One pre-existing bug found and deliberately left alone

`packages/cli/schema/cli-result.schema.json`'s blocker `kind` enum rejects five kinds `adaptBlocker`
passes through verbatim: `bound-stale`, `binding-missing`, `stale-task-heartbeat`,
`approval-decided-resume-required`, `side-effect-boundary-crossed`. Any run whose diagnosis carries
one makes `ultrafuzz why --json` throw `CLI producer result does not match ultrafuzz.cli.result.v2`.
It predates this branch. The schema also carries three entries (`binding`, `side-effect-boundary`,
`other`) that look like an intended *mapping* vocabulary `adaptBlocker` never implemented, so the
right fix is a design decision about mapping vs. widening — it deserves its own issue, not a
five-line enum edit inside a pin bump. `stalled` is not in that category: with pass-through semantics
the schema simply must list it, which this PR does.

<h3>Greptile Summary</h3>

Upgrades the pinned workflow engine to Smithers 0.35.0 and updates Ultrafuzz’s strict runtime mirrors for the release’s expanded states, events, usage fields, and inspection envelopes.

- Re-anchors compatibility patches and advances the deterministic npm resolution cutoff.
- Migrates canonical Smithers 0.34.0 generated manifests to the new pin.
- Updates lifecycle synchronization, retry behavior, generated agent adapters, schemas, fixtures, and smoke-test enforcement.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| packages/runtime/src/smithers.ts | Updates the engine pin boundary, compatibility patches, strict inspect parsing, and stalled-task retry handling without an accepted blocking issue. |
| packages/runtime/src/lifecycle-inspection.ts | Admits Smithers 0.35.0 why, node-usage, blocker, and lifecycle-event fields. |
| packages/runtime/src/smithers-package.ts | Advances the deterministic resolution cutoff and adds narrowly scoped migration of canonical 0.34.0 manifests. |
| packages/runtime/src/workflow-sync.ts | Projects the new run and node states into durable status, evidence, and retry reconciliation. |
| packages/runtime/src/state-export.ts | Updates strict status parsing and folds stalled Smithers counts into Ultrafuzz’s existing failed count. |
| packages/runtime/src/templates/smithers/agents/deepseek.tsx | Keeps Ultrafuzz-only authentication options out of the stricter upstream constructor. |
| packages/runtime/src/templates/smithers/agents/kimi.tsx | Separates Ultrafuzz-specific authentication and reasoning options from Smithers constructor options. |
| packages/runtime/scripts/run-pr-smoke-tests.mjs | Validates Bun smoke execution using emitted summary counts rather than unavailable per-test success lines. |
| packages/cli/schema/cli-result.schema.json | Extends the public blocker-kind contract with the newly introduced stalled state. |
| pnpm-lock.yaml | Resolves the workspace against Smithers 0.35.0 and its updated dependency closure. |

</details>

```mermaid
flowchart LR
  A[Smithers 0.35.0 outputs] --> B[Strict runtime parsers]
  B --> C[Lifecycle and state projection]
  C --> D[Durable Ultrafuzz state]
  C --> E[CLI inspection results]
  F[Canonical 0.34.0 manifest] --> G[Manifest migration]
  G --> H[Smithers 0.35.0 dependency install]
```
</details>

<sub>Reviews (2): Last reviewed commit: ["fix(changelog): resolve the Other change..."](https://github.com/monad-developers/ultrafuzz/commit/06986a6a6b9d067a405215ac5d6135cc8d79e7a4) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=57811793)</sub>

<details><summary><h4>Context used (4)</h4></summary>

- Knowledge Base — [CLI workflows](https://app.greptile.com/monad-foudnation/-/custom-context/knowledge-base/monad-developers/ultrafuzz/-/docs/cli-workflows.md)
- Knowledge Base — [Fix generated workspace dependency drift](https://app.greptile.com/monad-foudnation/-/custom-context/knowledge-base/monad-developers/ultrafuzz/-/reverts/incident-mitigation_384-20260808-generated-workspace-resolution-d2fdf77.md)
</details>


aviggiano (commented):
LGTM

Issue #1008, opened 2026-08-31
pnpm install fails on a clean checkout on macOS: allowBuilds is missing @smthrs/jj-darwin-arm64


mrthankyou:
## Summary

`allowBuilds` in `pnpm-workspace.yaml` records an explicit decision for `@smthrs/jj-linux-arm64` and `@smthrs/jj-linux-x64`, but not for `@smthrs/jj-darwin-arm64`. The pinned workflow engine (Smithers 0.35.0) resolves the darwin binary on macOS, so `pnpm install` fails on a pristine checkout and nothing can be built.

This is not a build-time-only annoyance: install fails *before* any build step, and it also mutates a tracked file.

## Reproduction

macOS arm64, pristine checkout of `main` (`ab3c03f4`):

```
$ git status --porcelain pnpm-workspace.yaml     # clean
$ pnpm install --frozen-lockfile
[ERR_PNPM_IGNORED_BUILDS] Ignored build scripts: @smthrs/jj-darwin-arm64@0.35.0

Run "pnpm approve-builds" to pick which dependencies should be allowed to run scripts.
$ echo $?
1
```

## Second-order effect: the tracked file is rewritten

The failed install writes a placeholder into `pnpm-workspace.yaml`:

```diff
 allowBuilds:
   "@moonshot-ai/kimi-code": false
   "@parcel/watcher": false
+  '@smthrs/jj-darwin-arm64': set this to true or false
   "@smthrs/jj-linux-arm64": false
```

So a first-time macOS contributor ends up with a dirty tracked file. If they then add the real entry without noticing the placeholder, every subsequent pnpm command fails on the YAML instead:

```
[ERROR] duplicated mapping key (30:3)
```

## Impact

No one can install or build on macOS from a clean checkout. `pnpm install`, `pnpm build`, `pnpm test`, and `pnpm ci` are all unreachable until the entry is added by hand.

## Fix

Add `"@smthrs/jj-darwin-arm64": false`, matching the two sibling `@smthrs/jj-*` platform entries.

`false` is the correct value — the build script is a no-op:

- the package's only script is `postinstall`: `node -e "const fs=require('fs');const p=__dirname+'/bin/jj';if(fs.existsSync(p))fs.chmodSync(p,0o755)"`
- the vendored `bin/jj` already ships mode `0755` (`-rwxr-xr-x`)

so running it would change nothing.

## Verified

With the one-line entry applied to an otherwise pristine checkout:

| step | before | after |
[...]

Issue #988, opened 2026-08-27
ultrafuzz clean cannot remove sealed execution snapshots, and the leak eventually makes new campaigns unlaunchable


mrthankyou:
## Summary

`ultrafuzz clean` fails for any run that has published a workflow execution snapshot:

```
$ ultrafuzz clean <run-id> --yes
error: CLEAN_REMOVE_FAILED: generated artifact cleanup failed; run source refs were preserved
```

`sealSnapshotPermissions` seals snapshot directories read-only, leaving them `dr-x------`. Removing an entry requires the write bit on its **parent directory**, not on the entry itself, so `fs.rmSync(..., { force: false })` cannot unlink anything inside a sealed snapshot.

`--dry-run` reports success for the same run:

```
$ ultrafuzz clean <run-id> --dry-run
Removed: runs/<run-id>
```

so the plan and the real outcome disagree.

## Impact

This is not only wasted disk. Snapshots accumulate at roughly **1 GB and ~59,000 files per run**, and because `.ultrafuzz` is untracked, every leaked file joins the untracked-file enumeration that workflow submission performs.

Observed on one project after twelve runs:

| | |
|---|---|
| files under `.ultrafuzz` | 533,313 |
| runs directory | 8.8 GB |

The next submission then died with:

```
WORKFLOW_SUBMISSION_FAILED: workflow runner command failed (exit ENOBUFS, signal SIGTERM)
```

The enumeration overflowed the child process's stdout buffer. So the leak escalates from disk usage to **campaigns that cannot launch at all**, and the documented way to recover is the command that cannot work. Recovering by hand requires `chmod -R u+w` before `rm -rf`.

## Fix

PR #985 restores owner write permission on directories before removal:

- directories only (unlinking depends on the parent's write bit, not the file's)
- only the owner write bit is added
- symlinks are never followed, so permissions outside the tree cannot be affected
- the removal root is already resolved and path-guarded by the caller

Verified end to end: `Removed: runs/<run-id>`, 90,357 files / 1.8 GB reclaimed. `packages/runtime/test/clean.test.ts` passes 4/4.

## Notes

- Not platform-specific; sealed directories behave the same on Linux.
- On macOS that test suite needs `TMPDIR` set to a canonical path, because `os.tmpdir()` returns a symlinked `/var/folders/...` and the suite trips a symlink guard during `planRun`. Unrelated to this issue, but it fails on `main` too.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1016, opened 2026-08-31
feat(runtime): run campaigns on darwin via volfs inode-rooted descriptor paths

### LINKED ISSUE #1011: Campaigns cannot run on macOS: descriptor-rooted traversal assumes /proc, and the failure surfaces late as a tampering error
## Summary

Campaign submission anchors snapshot publication, sealing, and execution to an open directory descriptor, addressing that descriptor as a path so that renaming or symlink-swapping a component mid-publication cannot redirect where bytes are written. Node exposes no `openat(2)`, so this is done by spelling a descriptor as a path — and the only spelling the code knows is `/proc`.

macOS has no `/proc`. Nothing declares that: `package.json` has no `os` field, the CLI builds and reports itself as `darwin-arm64`, and `init`, `validate`, `references`, `doctor`, and full graph planning all succeed. The campaign then fails **after** planning, prompt rendering, reference materialization, and a multi-gigabyte controller snapshot publication, with:

```
workflow execution snapshots descriptor path changed at the controller command boundary
```

which reads as though tampering was detected. It means the platform has no `/proc`.

## Root cause

`packages/runtime/src/workflow-execution-snapshot-capability.ts` (main @ `ab3c03f4`, L545-563):

```ts
function directoryDescriptorPath(descriptor: number): string | undefined {
  for (const candidate of [`/proc/self/fd/${descriptor}`, `/dev/fd/${descriptor}`]) {
    try {
      if (fs.statSync(candidate).isDirectory()) return candidate;   // identity never checked
    } catch { /* next */ }
  }
  return undefined;
}

function controllerDirectoryDescriptorPath(descriptor: number): string | undefined {
  const candidate = `/proc/${process.pid}/fd/${descriptor}`;        // Linux-only spelling
  ...
}
```

On macOS `/dev/fd/<n>` **does** stat as a directory, so `directoryDescriptorPath` returns it on `isDirectory()` alone — but that entry reports devfs's own `st_dev`, and it cannot be traversed at all. `verifiedDirectoryDescriptorPath` (L565) then compares device and inode, they disagree, and it throws the message above. The check that would have caught this cheaply runs too late and reports the wrong cause.

## Evidence

Same script, both platforms. `NEW` is the proposed logic (identity-checked candidate selection, with a volfs candidate appended).

**macOS (Darwin 25.6.0, arm64)**

```
platform: darwin/arm64   opened dev=16777232 ino=15344302
  main   directoryDescriptorPath   -> /dev/fd/11              verify: THROWS: ...descriptor path changed at the controller command boundary
  main   controllerDescriptorPath  -> undefined               verify: undefined
  NEW    directoryDescriptorPath   -> /.vol/16777232/15344302 verify: OK
  NEW    controllerDescriptorPath  -> /.vol/16777232/15344302 verify: OK
  traversal through NEW path       -> ok
```

`/dev/fd/11` reports `dev=18446744071813286000` against the opened directory's `dev=16777232`, and `/dev/fd/11/child` is `ENOENT`.

**Linux (arm64 container, same script)**

```
platform: linux/arm64   opened dev=63 ino=46563662
  main   directoryDescriptorPath   -> /proc/self/fd/17        verify: OK
  main   controllerDescriptorPath  -> /proc/1/fd/17           verify: OK
  NEW    directoryDescriptorPath   -> /proc/self/fd/17        verify: OK
  NEW    controllerDescriptorPath  -> /proc/1/fd/17           verify: OK
  traversal through NEW path       -> ok
```

Linux resolves to exactly the same paths before and after. The `/proc` candidates are tried first and win, and the added identity check is a strengthening that `/proc` already satisfies.

## Proposed approach

macOS exposes volfs: `/.vol/<dev>/<ino>` names a path rooted at an inode rather than at a name, which is the property this code actually needs — renaming or symlink-swapping any component cannot redirect resolution. Traversal, `mkdir`, `O_NOFOLLOW|O_DIRECTORY` open, `O_CREAT|O_EXCL`, `symlink`, and `chmod` all work through it. Because the descriptor stays open for as long as the path is used, the inode cannot be recycled, so the volfs path resolves to exactly the directory that was opened. It is also process-independent, so unlike `/proc/<pid>/fd` it satisfies the cross-process requirement.

1. Offer `/.vol/<dev>/<ino>` as a descriptor-path candidate, and apply the existing `isDirectory` + device + inode identity check during *selection* rather than only afterwards. Nothing is relaxed to make volfs pass — the same check that rejects `/dev/fd` accepts volfs.
2. Canonicalize through the tracked lexical path where `realpath(3)` is needed, since volfs paths intentionally have no realpath resolution. Every such site already asserts that its access and lexical paths name the same device and inode.
3. Verify the sealed Bun module root by comparing the held descriptor's device and inode when `/proc` is absent, instead of realpath equality.

Beyond snapshot publication there are two further macOS blockers, both reached only once publication succeeds:

- **Detached engine spawn.** The controller hands the snapshot to the engine as fd 3 and rewrites paths to the literal `/proc/self/fd/3` (`smithers.ts` L318, L328, L396, L482, L523), with the parent addressing the same directory as `/proc/<pid>/fd/<n>`. The engine dies before admission on `ENOENT: /proc/self/fd/3/controls/bunfig.toml`. A single volfs string is valid in both parent and child and needs no pid or fd number. fd 3 inheritance is unchanged.
- **Claude subscription auth.** Covered separately by #986.

## What works on macOS today

`init` (67 files created on a fresh project), `validate`, `references`, `doctor`, and graph planning all succeed — which is what makes the late failure costly. Every failed attempt also leaves several gigabytes of run evidence behind.

## Status

I have this working on a branch, but I am not opening a PR yet: it needs a real rebase onto current `main`. Two of the three substantive commits cherry-pick clean; the third conflicts in `smithers.ts` across 5 hunks, because #1004 and the Smithers 0.35.0 upgrade moved the same regions. Happy to do that rebase if the approach looks right — checking on direction first, since it is a platform-support commitment rather than a bug fix.

## Environment

- ultrafuzz `main` @ `ab3c03f4`
- macOS (Darwin 25.6.0) arm64, node v26.3.1, pnpm 11.1.1
- Linux comparison: `node:22-slim`, linux/arm64
- Note: building on macOS at all currently requires #1008 / #1009.

mrthankyou:
Fixes #1011.

Campaign submission addresses an open directory descriptor as a path, so that
renaming or symlink-swapping a component mid-publication cannot redirect where
bytes are written. The only spelling it knew was `/proc`, which macOS does not
have. `directoryDescriptorPath` accepted `/dev/fd/<n>` on `isDirectory()` alone
-- an entry that reports devfs's own `st_dev` and cannot be traversed -- and
`verifiedDirectoryDescriptorPath` then rejected it as

```
workflow execution snapshots descriptor path changed at the controller command boundary
```

which reads as tampering for what is really a missing platform capability.

macOS exposes volfs: `/.vol/<dev>/<ino>` is rooted at an inode rather than a
name, which is exactly the swap-immunity this code needs, and it is
process-independent so it also satisfies the cross-process requirement that
`/proc/<pid>/fd` served. The identity check is now applied during candidate
*selection* rather than only afterwards, so nothing is relaxed to make volfs
pass: the same check that rejects `/dev/fd` accepts volfs.

## Linux is unchanged

The `/proc` candidates are tried first and win, and the added identity check is
a strengthening `/proc` already satisfies. Verified by running the selection
logic on both platforms:

| | `main` today | this branch |
| --- | --- | --- |
| Linux (`linux/arm64`) | `/proc/self/fd/N`, `/proc/<pid>/fd/N` | **identical** |
| macOS (darwin/arm64) | `/dev/fd/N` -> throws | `/.vol/<dev>/<ino>`, traversable |

## Regression coverage

The two snapshot capability tests were gated on `fs.existsSync("/proc/self/fd")`,
so they skipped on darwin whether the runtime worked or not -- which is why this
class of bug could exist at all. Gating them on a probe of the capability
instead makes them the regression test:

| `workflow-execution-snapshot-capability` on darwin | result |
| --- | --- |
| pristine `main` + probe | **0/2** -- fails with the error quoted above |
| this branch + probe | **2/2** |

The other eleven `/proc` gates are deliberately left alone: those tests read
`/proc` directly in their bodies (enumerating `/proc/self/fd`, building
`/proc/<pid>/fd/<n>` pathnames), so there the gate describes what the test
needs, not what the runtime supports.

## An upgrade break this surfaced

Not a conflict -- a real defect that rebasing onto 0.35.0 exposed.

`SMITHERS_CLI_RESUME_SNAPSHOT_TRANSFER_PATCH` is derived from its predecessor by
dropping `--preserve-symlinks`. Substituting the inline bun startup arguments
for a descriptor-rooted helper call removed that replacement's target, so the
derivation became a no-op and the patch equalled its own predecessor.
`classifyRequiredSmithersPatch` treats that as `incompatible`, so an
installation patched by the current release would have failed on its next launch
with *"pinned workflow runner resume snapshot transfer implementation is
incompatible"*. The predecessor now keeps the inline literal it has always had,
the current patch is derived from the post-0.35.0 form, and that form is
registered as a second predecessor.

## Open question for review

Five other patch sites change shape under this work and carry no predecessor for
the form the current release writes: `detached_snapshot_transfer`,
`supervisor_descriptor`, `workflow_path_import`, `manifest_relaunch`, and
`process_snapshot_anchor`. The same upgrade break fixed above plausibly applies
to each.

I have not guessed at this, because it is a policy decision rather than a
mechanical one: either every changed patch carries its previous shape as a
predecessor, or the patch set gets a version and a changed set requires a fresh
controller install. The first grows this PR substantially (the anchor patch
alone is ~180 lines of previous shape); the second is a smaller change with
different operational consequences. Happy to implement whichever you prefer.

## Verification

- `pnpm build`, `tsc --noEmit`: pass
- Full runtime suite, `main` vs branch with `TMPDIR` held constant: **0 regressions**, measured both with the default tmpdir (362 vs 362 distinct failures) and with a canonical one (45 distinct failures, every one of them also failing on `main`)
- `pnpm validate:pack`: passes for all 12 packages on darwin
- A real campaign against a public Solidity project on macOS reaches workflow submission -- planning, prompt rendering, reference materialization and snapshot publication all succeed, where `main` cannot start one at all

## Notes

- Building on macOS at all currently needs #1008 / #1009 (`allowBuilds` is missing `@smthrs/jj-darwin-arm64`).
- Two pre-existing darwin problems are **not** addressed here and are worth separate issues: `os.tmpdir()` returns a symlinked `/var/folders/...`, which fails the runtime's canonical-root checks (44 baseline failures are literally `must be canonical`, and `pnpm validate:pack` fails with `SYMLINK_PATH: /var` unless `TMPDIR` is canonical); and a full test run leaves roughly 180k files of sealed temp state behind that it cannot remove, the suite-side twin of #988.

<h3>Greptile Summary</h3>

This PR enables descriptor-rooted workflow snapshot publication and execution on macOS through volfs while retaining the existing `/proc` path on Linux.
- Selects descriptor paths only after verifying directory device and inode identity.
- Adapts snapshot publication, executable verification, Bun startup controls, and child-process snapshot transfer to volfs.
- Adds capability-based test gating and compatibility migration coverage for the resume snapshot-transfer patch.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| packages/runtime/src/workflow-execution-snapshot-capability.ts | Adds identity-verified volfs candidates for local and cross-process directory-descriptor paths. |
| packages/runtime/src/workflow-integrity.ts | Extends descriptor-rooted snapshot publication and verification to macOS volfs paths while retaining inode checks. |
| packages/runtime/src/smithers.ts | Makes generated Smithers snapshot-transfer and Bun startup patches use platform-specific descriptor roots and recognizes prior resume-patch forms. |
| packages/runtime/src/smithers-executable-capability.ts | Adds identity-aware handling for executable paths reached through volfs. |
| packages/runtime/test/descriptor-paths.ts | Introduces a behavioral probe for traversable descriptor-rooted directory paths. |
| packages/runtime/test/runtime.test.ts | Updates compatibility-patcher coverage for platform-specific snapshot roots and predecessor migration. |
| packages/runtime/test/workflow-execution-snapshot-capability.test.ts | Uses the descriptor-path capability probe so snapshot tests run on supported macOS hosts. |

</details>

<sub>Reviews (5): Last reviewed commit: ["Merge remote-tracking branch &#39;origin/mai..."](https://github.com/monad-developers/ultrafuzz/commit/a4cc9d04541430845dd11e01ed6fd54527afcc3f) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=58788002)</sub>

**Context used:**


mrthankyou:
Moving this to draft deliberately, to separate it from the stability work in #1019, #1021, #1023 and #1025.

Those four are mechanical and low-risk: canonicalization one-liners, comment markers, and documentation. This one is not the same kind of change. It adds roughly ten functions to security-critical paths and changes how identity is established there:

- `ultrafuzzSameDirectory`, `canonicalSnapshotPath` and `canonicalExecutablePath` compare device and inode where the code previously compared `realpath` equality. That is what volfs forces -- volfs paths have no realpath resolution -- and it is what those checks already meant, but it is a change to a trust boundary rather than a tidy-up.
- `isVolfsPath` special-casing is introduced in two files.
- `BUN_MODULE_CONFINEMENT_SOURCE`, the sealed module confinement, is modified.

The open question in the description is also unresolved: five other patch sites change shape under this work and carry no predecessor for the form the current release writes, so the upgrade break fixed here for `resume_snapshot_transfer` plausibly applies to them too. Whether to encode that lineage or to version the patch set and require a fresh controller install is a maintainer decision, and it materially changes this PR's size.

Nothing here is blocked on my side -- the branch is green apart from the shared secret-scanner gate that #1021 clears, Linux output is byte-identical, and the snapshot capability tests go 0/2 on `main` to 2/2 with it. I would rather it be reviewed as the platform-support feature it is, once the lineage question has an answer, than ride along with the stability batch.

mrthankyou:
Rebased onto `fix/gitleaks-test-fixture-allowlist` (#1021) and retargeted this PR's base to it.

This PR was failing `External static analysis` (and `release-gates` with it) on pre-existing secret-scanner findings, not on anything it adds. The gate runs with `VALIDATE_ALL_CODEBASE: false`, so gitleaks scans changed *files* rather than changed *lines* -- editing any part of a file that already contains a fake credential fixture fails the gate on content this change never touched. #1021 marks those fixtures.

Stacking rather than duplicating the markers keeps this diff to its own delta (unchanged at +274/-54) and makes the dependency explicit instead of implicit. It should be retargeted back to `main` once #1021 lands; the rebase was clean, so that will be a no-op.

To be explicit, since the failure name invites the wrong conclusion: no credential is added by this branch. Scanning only the lines it adds reports no findings, and the flagged fixtures trace to #865.

PR #985, opened 2026-08-27
fix(runtime): let clean remove sealed execution snapshots

mrthankyou:
## Problem

`ultrafuzz clean` cannot remove any run that has published a workflow execution snapshot:

```
CLEAN_REMOVE_FAILED: generated artifact cleanup failed; run source refs were preserved
```

`sealSnapshotPermissions` seals snapshot directories read-only, leaving them `dr-x------`. Unlinking an entry requires the write bit on its **parent directory**, not on the entry itself, so `fs.rmSync(..., { force: false })` cannot remove anything inside a sealed snapshot.

`--dry-run` still reports `Removed: runs/<id>`, so the plan and the actual outcome disagree.

## Why it matters beyond disk usage

The tool cannot remove state it created. Snapshots accumulate at roughly **1 GB and ~59,000 files per run**. Because `.ultrafuzz` is untracked, every leaked file joins the untracked-file enumeration that workflow submission performs.

Observed on a real project after twelve runs:

| | |
|---|---|
| files under `.ultrafuzz` | 533,313 |
| runs directory | 8.8 GB |

The next submission then died with:

```
WORKFLOW_SUBMISSION_FAILED: workflow runner command failed (exit ENOBUFS, signal SIGTERM)
```

The untracked-file listing overflowed the child's stdout buffer. So the leak eventually makes new campaigns **unlaunchable**, and the documented way to recover is the command that cannot work.

## Fix

Restore owner write permission on directories before removing the tree.

- Directories only; regular files are left alone (unlinking depends on the parent's write bit).
- Only the owner write bit is added (`mode | 0o700`).
- Symlinks are never followed, so permissions outside the tree cannot be affected.
- The removal root is already resolved and path-guarded by the caller.

## Verification

Before the fix, on a run with a published snapshot:

```
$ ultrafuzz clean mm-13 --yes
error: CLEAN_REMOVE_FAILED: generated artifact cleanup failed; run source refs were preserved
```

After:

```
$ ultrafuzz clean mm-13 --yes
Removed: runs/mm-13
```

90,357 files / 1.8 GB reclaimed, and `.ultrafuzz/runs` returned to empty.

`packages/runtime/test/clean.test.ts` passes (4/4).

[...]

PR #1019, opened 2026-08-31
test(runtime): create temporary roots that are canonical and removable

### LINKED ISSUE #1017: Runtime tests build temporary roots under a symlinked os.tmpdir(), so ~314 tests fail on macOS before any code change
## Summary

The runtime test suite builds its temporary roots with `fs.mkdtempSync(path.join(os.tmpdir(), ...))`. On macOS `os.tmpdir()` is `/var/folders/...`, and `/var` is a symlink to `/private/var`, so every one of those roots is non-canonical. The runtime deliberately requires canonical roots, so the tests fail against their own fixtures.

```
macOS   tmpdir   : /var/folders/4j/5w5ycfps4r54cr1xxf8_76zc0000gn/T
        realpath : /private/var/folders/4j/5w5ycfps4r54cr1xxf8_76zc0000gn/T
        canonical: false

Linux   tmpdir   : /tmp
        realpath : /tmp
        canonical: true
```

Linux is unaffected, which is why CI has never seen this.

## Impact

Running `pnpm --filter @ultrafuzz/runtime test` on macOS produces **362 distinct failing tests**. The dominant error shapes are the canonicality assertions, not real defects:

```
34x  Error: artifact verification marker root must be canonical
 4x  Error: pinned source root is not a canonical directory
 2x  Error [ArtifactPathError]: untracked governance input root must be canonical
 2x  Error [ArtifactPathError]: provider route config root must be canonical
```

44 failures are literally `must be canonical`, and 75 error lines reference `/var/folders`.

Pointing `TMPDIR` at a canonical directory takes the same suite on the same commit to **48 distinct failures**, with 674 passing. So roughly 314 of the 362 failures are an artifact of the temporary directory rather than anything about the code under test.

The same root cause breaks `pnpm validate:pack` on macOS, which creates its consumer under `os.tmpdir()`:

```
$ pnpm validate:pack
Error: packed validate failed (... ultrafuzz validate --project /var/folders/.../initialized-project --json)
  "code": "SYMLINK_PATH",
  "message": "Topology path contains a symlink component: /var",
```

With a canonical `TMPDIR` it passes for all 12 packages.

## Why this matters beyond convenience

A macOS contributor cannot tell a real regression from the noise: the suite is red before they change anything, and the failures name canonicality rather than the tmpdir, so the cause is not obvious from the output. It also means no darwin-specific defect can be caught by the suite, because the suite never gets far enough to exercise one.

## Scope

136 raw `mkdtempSync(path.join(os.tmpdir(), ...))` sites across 39 test files. Exactly one call site currently canonicalizes.

## Suggested fix

Canonicalize once in a shared test helper and use it everywhere, rather than sprinkling `realpathSync` at 136 call sites. This pairs naturally with the cleanup problem in the companion issue, since both want the same helper: one function that returns a canonical temporary root and reliably removes it afterwards.

The production code should keep rejecting non-canonical roots; that check is doing its job here.

## Environment

- macOS (Darwin 25.6.0) arm64, node v26.3.1, pnpm 11.1.1
- ultrafuzz `main` @ `ab3c03f4`

### LINKED ISSUE #1018: Runtime tests leave sealed temporary state behind, and an accumulated tmpdir silently slows the suite from 2m30s to 90+ minutes
## Summary

The runtime test suite creates sealed execution snapshots and controller roots in temporary directories and never removes most of them. `sealSnapshotPermissions` leaves directories `dr-x------`, and unlinking an entry needs the write bit on its **parent**, so even the fixtures that do call `rmSync` cannot remove what they created.

This is the test-side twin of #988, which fixed exactly this for `ultrafuzz clean` by restoring owner write permission on directories before removal.

## Measured

One run of `pnpm --filter @ultrafuzz/runtime test` left behind:

```
top-level entries                 1,232
files                           567,065
sealed directories (dr-x------)  64,331
```

Sample of what survives:

```
dr-x------  <tmp>/ultrafuzz-controller-rcRqfm/operator-npm
dr-x------  <tmp>/ufz-snapshot-attestation-SLhURB/snapshots/generation
dr-x------  <tmp>/ufz-snapshot-attestation-Q8oOod/snapshots/generation
```

Across repeated runs a single temporary directory reached **1,047,808 files and 11 GB**.

## Impact

It is not only disk. Once the temporary directory is large, every later run slows down dramatically: a suite that completes in about 2m30s against a clean temporary directory took over 90 minutes against an accumulated one, with individual tests going from tens of milliseconds to 55-85 seconds. The degradation is silent and looks like a hang or a regression in whatever you happen to be working on.

Because the directories are sealed, `rm -rf` on the temporary directory also fails for a developer trying to recover; it needs `chmod -R u+w` first, which is the same trap #988 described for `ultrafuzz clean`.

## Current state

```
test files creating temporary roots   39
                    calling rmSync    25
        restoring permissions first   11
```

`snapshotFixture` in `workflow-execution-snapshot-capability.test.ts` is representative: it creates a snapshot root and never removes it at all.

## Suggested fix

A shared helper that creates a temporary root and registers it for removal, restoring owner write permission on directories on the way down -- the same treatment `restoreRemovableDirectoryPermissions` already applies in `clean.ts` (#985). Call sites then get cleanup for free rather than each fixture having to remember both halves of the problem.

This pairs with #1017, which needs the same helper to canonicalize the root; fixing them together is one change rather than two passes over the same 136 call sites.

## Environment

- macOS (Darwin 25.6.0) arm64, node v26.3.1
- ultrafuzz `main` @ `ab3c03f4`
- Not macOS-specific: the sealing behaviour and the missing cleanup are platform-independent, though the slowdown is most visible where the temporary directory persists between runs.

mrthankyou:
Fixes #1017. Fixes #1018.

Both issues come from the same line, repeated 136 times:

```ts
fs.mkdtempSync(path.join(os.tmpdir(), "some-prefix-"))
```

**It is not canonical.** On macOS `os.tmpdir()` is `/var/folders/...` and `/var`
is a symlink to `/private/var`, so the runtime rejects the fixture's own root --
correctly, since that check exists to stop a component being swapped underneath
a resolved path. The suite failed against itself rather than against anything
under test. Linux is unaffected (`/tmp` is already canonical), which is why CI
never saw this.

**It is not removable.** Sealed execution snapshots are left `dr-x------`, and
unlinking an entry needs the write bit on its *parent* rather than on the entry
itself, so `rmSync` cannot remove a tree that contains one. Of 39 files creating
temporary roots, 25 attempted removal and 11 restored permissions first.

## Measured on macOS, `main` vs this branch, default tmpdir throughout

| | `main` | this branch |
| --- | --- | --- |
| distinct failures | 362 | **219** |
| new failures | — | **0** |
| `must be canonical` failures | 44 | **0** |
| leaked directories | 1,232 | **0** |
| leaked files | 567,065 | **0** |

The remaining 219 all fail on `main` too. 159 are the darwin `/proc` limitation
tracked in #1011 and fixed by #1016; the rest are unrelated to temporary roots.

The one directory that still survives a run is a deliberately retained
native-continuation controller root, which production removes through its own
exit hook -- not a leak, and left alone.

## What changed

`temporaryRoot(prefix)` builds on a canonicalized tmpdir and registers the root
for removal, restoring owner write permission on directories on the way down --
the same treatment `restoreRemovableDirectoryPermissions` already applies in
`clean.ts` (#985).

`registerTemporaryPath(dir)` covers paths a fixture places *beside* its root
rather than inside it, which root removal cannot reach. These were the last
survivors: `${project}-fake-bin`, `-fake-ps-bin`, `-redaction-bin`,
`-sealed-runner-interpreter` and friends.

Two production sites are canonicalized for the same reason the tests are -- the
runtime requires canonical roots, so it should create them:

- `smithers.ts` `ultrafuzz-controller-`
- `workspace-handoff.ts` `ultrafuzz-workspace-index-`

Incidental cleanups the rewrite forced:

- `trusted-cli.test.ts` had a local `temporaryRoot()` wrapper that the rename
  made self-recursive; removed, with its prefix inlined at its 16 call sites.
- `workflow-execution-snapshot-capability.test.ts` had a local const shadowing
  the import; renamed to `fixtureRoot`.
- 36 files no longer need `os`, and 3 no longer need `mkdtempSync`.

## Why this is worth more than the failure count suggests

A macOS contributor currently cannot tell a real regression from the noise: the
suite is red before they touch anything, and the failures name canonicality
rather than the temporary directory, so the cause is not evident from the
output. The accumulation problem compounds it -- the slowdown is silent and
reads as a hang or as a regression in whatever you happen to be editing.

It also means darwin-specific defects could not be caught by the suite at all,
because it never got far enough to exercise one. #1016 is a concrete instance:
the tests that would have caught it were skipping.

## Verification

- `tsc --noEmit` (both `tsconfig.json` and `tsconfig.test.json`), `eslint`, `prettier --check`: clean
- `pnpm build`: clean
- Full runtime suite on macOS with the default tmpdir, before and after, with the temporary directory cleared between runs so the leak delta is real

<h3>Greptile Summary</h3>

The PR centralizes runtime-test temporary-root creation and cleanup while canonicalizing the two production temporary roots that feed trusted runtime paths.

- Adds a shared canonical temporary-root helper with exit-time permission restoration and recursive cleanup.
- Registers fixture-created sibling paths that root cleanup cannot otherwise reach.
- Migrates runtime fixtures to the helper and removes obsolete filesystem imports and local wrappers.
- Canonicalizes Smithers controller and workspace-index temporary roots before creation.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| packages/runtime/test/temporary-root.ts | Adds the shared canonical-root registry and best-effort exit cleanup used across runtime tests. |
| packages/runtime/src/smithers.ts | Creates the operator-controller project beneath the canonical system temporary directory. |
| packages/runtime/src/workspace-handoff.ts | Creates temporary Git index directories beneath the canonical system temporary directory. |
| packages/runtime/test/runtime.test.ts | Migrates the largest runtime fixture surface to canonical registered roots and explicitly tracks sibling fixture paths. |
| packages/runtime/test/trusted-cli.test.ts | Removes a conflicting local temporary-root wrapper and uses the shared helper with explicit prefixes. |

</details>

```mermaid
flowchart TD
  A[Test or runtime requests temporary root] --> B[Resolve canonical system tmpdir]
  B --> C[Create uniquely prefixed directory]
  C --> D{Test fixture?}
  D -->|Yes| E[Register root and sibling paths]
  D -->|No| F[Use runtime-specific cleanup ownership]
  E --> G[Process exit]
  G --> H[Restore owner permissions recursively]
  H --> I[Remove registered paths]
```
</details>

<sub>Reviews (4): Last reviewed commit: ["Merge remote-tracking branch &#39;origin/mai..."](https://github.com/monad-developers/ultrafuzz/commit/b743d0ff2f6f8d851eeb960cc85c8d410500b106) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=58798884)</sub>

**Context used:**


mrthankyou:
Rebased onto `fix/gitleaks-test-fixture-allowlist` (#1021) and retargeted this PR's base to it.

This PR was failing `External static analysis` (and `release-gates` with it) on pre-existing secret-scanner findings, not on anything it adds. The gate runs with `VALIDATE_ALL_CODEBASE: false`, so gitleaks scans changed *files* rather than changed *lines* -- editing any part of a file that already contains a fake credential fixture fails the gate on content this change never touched. #1021 marks those fixtures.

Stacking rather than duplicating the markers keeps this diff to its own delta (unchanged at +292/-211) and makes the dependency explicit instead of implicit. It should be retargeted back to `main` once #1021 lands; the rebase was clean, so that will be a no-op.

To be explicit, since the failure name invites the wrong conclusion: no credential is added by this branch. Scanning only the lines it adds reports no findings, and the flagged fixtures trace to #865.

PR #1037, opened 2026-09-01
fix(runtime): preserve admitted Pi environment across detached continuations

### LINKED ISSUE #1035: Detached resume drops external Pi CLI PATH before agent preflight
## Summary

A native same-run resume can lose an external user-local `PATH` entry before a `PiAgent` task reaches Smithers preflight. The `ultrafuzz resume` submit process can resolve `pi`, the detached controller starts, and the selected node is reopened, but agent preflight fails with:

```text
[diagnostics] pi: cli_installed=fail: pi not found on PATH
```

The failure occurs before a provider request. Re-running `resume --reset-node` repeats the same preflight failure.

## Minimal synthetic fixture

Use a temporary Git project with a one-node agentic topology whose default model profile selects `PiAgent`. Put either the real Pi CLI or an executable stub named `pi` in an external directory such as `/tmp/uf-path-repro/bin`, and prepend that directory only in the submitting environment:

```bash
export PATH="/tmp/uf-path-repro/bin:$PATH"
command -v pi

ultrafuzz resume synthetic-run \
  --project /tmp/uf-path-repro/project \
  --reset-node node:worker \
  --max-concurrency 1 \
  --json

ultrafuzz node synthetic-run node:worker \
  --project /tmp/uf-path-repro/project \
  --attempts \
  --json
```

The fixture needs a previously failed or stopped same-ID run so native continuation is selected. A stub is sufficient to demonstrate this defect because the failure is Smithers' `which pi` check before invocation. The fixture must keep the target project and the external bin as separate directories.

## Actual behavior

- `command -v pi` succeeds in the submit environment.
- `ultrafuzz resume` reports the same run as submitted.
- The detached run emits `RunStarted` and `NodeStarted`.
- The node fails within seconds because `pi` is absent from the preflight environment's `PATH`.

## Expected behavior

External, non-target `PATH` entries admitted by `composeSmithersCommandPath()` should remain available to the detached controller and its generated Pi adapter for the lifetime of a native continuation. Target-owned bin directories must remain excluded.

## Relevant boundary

This appears to be in the Ultrafuzz/Smithers detached-environment handoff, not model configuration. `composeSmithersCommandPath()` explicitly retains external absolute entries, while the generated Pi adapter later builds a filtered child environment before Smithers runs `which pi`.

## Acceptance criteria

- A native continuation test places `pi` in an external temporary bin and proves resumed Pi preflight resolves it.
- The test exercises a detached same-run resume, not only direct adapter construction.
- Target-owned PATH entries remain rejected.
- Provider credentials remain route-scoped and are not included in diagnostics or test output.
- Other agent adapters and fresh-run behavior remain unchanged.

aviggiano:
A native detached continuation advertises a persisted workflow under the target. The shared child-environment filter then blanks the entire admitted PATH, including external Pi, and removes Pi's isolated target-local home. The original Pi-specific override did not survive this filter.

Preserve the controller-admitted PATH in both shared filter passes and distinguish a native target-local Pi home from actual execution snapshot aliases. Controller-only variables, arbitrary path aliases, and unrelated provider credentials remain excluded. Remove the ineffective Pi-specific override.

Validation: Node 24 runtime build/typecheck and test compilation; 10 adapter boundary tests; 8 focused Bun contracts covering Pi, detached continuation, credential isolation, Codex, DeepSeek, and OpenRouter; strict lint and formatting. The detached fixture runs Ultrafuzz resume/reset-node and detach, a separate child, generated Pi adapter, pinned Smithers preflight, and an external stub CLI. Smithers inspect/reset responses are fixture-backed. Reverting only the shared environment fix reproduces `pi not found on PATH`.

Closes #1035.

<h3>Greptile Summary</h3>

This PR preserves the controller-admitted command path and native target-local Pi home across detached continuations while continuing to remove execution-snapshot aliases, controller-only variables, and unrelated provider credentials.

- Introduces separate root selection for Pi’s home based on whether explicit snapshot markers are present.
- Exempts the already-admitted PATH from both shared alias-filtering passes.
- Adds focused generated-environment and detached continuation coverage.
- Updates the adapter boundary fingerprint and complexity ceiling for the environment template.

The PR appears safe to merge, with the continuation behavior and isolation boundaries covered by focused tests.

No actionable failure remains: production PATH values pass through controller admission, actual snapshots carry explicit roots used by the stricter Pi-home filter, and the new tests exercise both native continuation preservation and snapshot rejection.

| Filename | Overview |
|----------|----------|
| packages/runtime/src/templates/smithers/agents/environment.tsx | Preserves admitted PATH and conditionally preserves Pi’s target-local home while retaining snapshot and credential isolation. |
| packages/runtime/test/runtime.test.ts | Adds direct environment contracts and an end-to-end detached Pi continuation fixture covering path admission, home propagation, and credential isolation. |
| packages/runtime/test/agent-adapter-boundaries.test.ts | Updates the reviewed source fingerprint and syntax-node ceiling for the changed environment template. |

</details>

```mermaid
flowchart TD
  A[Controller environment] --> B[Controller PATH admission]
  B --> C[Detached continuation]
  C --> D[workflowControlChildEnvironment]
  D --> E{Environment variable}
  E -->|PATH| F[Preserve admitted PATH]
  E -->|PI_CODING_AGENT_DIR with explicit snapshot roots| G[Reject snapshot alias]
  E -->|PI_CODING_AGENT_DIR in native continuation| H[Preserve target-local Pi home]
  E -->|Controller-only or unrelated credentials| I[Blank value]
  F --> J[Generated Pi child]
  H --> J
```
</details>

<sub>Reviews (1): Last reviewed commit: ["Preserve admitted CLI paths and Pi homes..."](https://github.com/monad-developers/ultrafuzz/commit/a474b368d46cf1dae0c43c4a892a4ac70139ad3a) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=61583753)</sub>


aviggiano:
Holding this PR. The bug is real, but this change has no effect on the Pi child environment, and the fix belongs in `environment.tsx`.

**Root cause on main**
A native continuation sets `ULTRAFUZZ_WORKFLOW_PERSISTED_PATH` to the target tree (`start-run.ts:510-515`, `:567`). `environment.tsx:379-386` derives the project root as a snapshot root. `workflowControlChildEnvironment` (`environment.tsx:96-98`, `:104-106`) blanks every variable whose value contains that root. The controller PATH starts with `<project>/.ultrafuzz/runs/<id>/trusted-bin`, so PATH becomes `""` in `opts.env` (`pi.tsx:424-429`). `PI_CODING_AGENT_DIR` is blanked too. Smithers' `which pi` preflight reads `{...process.env, ...this.env}` (`opts.env`), not `command.env`.

**Why this PR does not fix it**
The PATH value added to `command.env` passes through the same filter and is blanked. Two independent simulations with the real templates under a native-continuation environment showed identical effective PATH (`""`) on main and on this branch. In a fresh run the added PATH equals what the child already inherits. The new assertions never set a persisted path, so no root exists and nothing is blanked; they fail on main only because the key is absent. The Bun CI lane that runs them was cancelled.

**Rework**
Skip PATH in both blanking loops of `workflowControlChildEnvironment`. Do not filter PATH per entry: the prepared forge-guard bin lives inside the target and would be dropped. Add a continuation-shaped test (persisted path under `<project>/.smithers/workflows`, PATH starting with `<project>/.ultrafuzz/runs/<id>/trusted-bin` plus an external bin) that asserts `agent.opts.env.PATH` keeps the external bin, and cover `PI_CODING_AGENT_DIR`. Update the `environment.tsx` fingerprints and drop the `pi.tsx` change. This fixes every adapter that uses the helper.

PR #621, opened 2026-08-15
fix(prompts): align strategy framing with output contracts

aviggiano:
## Summary

Closes #5. Supersedes the follow-ups tracked in #633.

Completes the strategy-role disposition from #5 and closes the remaining
pipeline-integration gaps behind it, so every strategy node's intake and every
published report row is bound to an authenticated producer rather than to
agent-authored or host-rendered content.

**Strategy roles (the original #5 decision).** `externalized-state-accounting`,
`rounding-direction-audit`, `state-machine-boundaries`,
`market-exhaustion-boundaries`, `lifecycle-view-boundaries`, `round-trip`, and
`packed-action-parity` become explicit source-backed bug search: they keep
`findings@2` and the passing-property guard, treat tests and PoCs as optional
evidence, and no longer require `generated-tests@3`. `boundary-tests`,
`admin-config-boundaries`, `payable-fallback-accounting`,
`order-replacement-collateral`, `stateful-invariant-handlers`, and
`dynamic-strategy-generator` are refined. `time-warp-sequences` is deferred
unchanged.

**Bounded ancestor intake.** Prompt rendering replaces the unbounded ancestor
collection helpers with `ancestor_contract_artifact_authority` and
`ancestor_artifact_path_authority`. Both render a fixed-size pointer to the
task-local `<workspace>/.ultrafuzz/authorities/<attemptId>.json` document the
runtime derives from the sealed task manifest, so a prompt never inlines a
rendered ancestor array. Legacy helpers are rejected with a migration message.
A 2,000-ancestor regression asserts the rendered Markdown is byte-identical to
the single-ancestor rendering and leaks no ancestor id or artifact path.

**Optional prerequisites, recovery, and TOCTOU.** Nodes in a `failure_policy:
continue` group no longer block dependents; verifier-persisted dependency
admission is authenticated against the sealed closure; recovery refuses to
resume unless every blocking failed attempt carries exact sealed task and
attempt identity; required-command preflight separates blocking from optional
specialist commands and never degrades an opaque probe failure to a warning;
deterministic Modal archives are built through directory file descriptors with
`O_NOFOLLOW` and identity re-checks.

**Report lifecycle.** Report rows must reconcile against the authenticated
severity-classified or bounded deduped population and its paired lifecycle
ledger; dropped candidates surface as non-production outcomes instead of
disappearing; a topology with no dedupe producer cannot authorize
agent-authored rows. Handler nodes publish qualifying production candidates
[...]

PR #1038, opened 2026-09-01
fix(runtime): reconcile reset attempt digests across refresh

### LINKED ISSUE #1036: reset-node reuses an immutable attempt-ledger identity
## Summary

After a failed node attempt has already been projected into the immutable node-attempt ledger, a same-run `resume --reset-node` can reopen the same runner attempt identity with different timestamps and state. The next status synchronization then rejects the replacement record even though the workflow is running.

## Minimal synthetic fixture

1. Create a temporary Git project with a one-node workflow and a deterministic stub agent.
2. Let attempt 1 fail, then invoke `ultrafuzz status --json` once so the terminal attempt is projected into the node-attempt ledger.
3. Run `ultrafuzz resume synthetic-run --project /tmp/uf-reset-ledger/project --reset-node node:worker --max-concurrency 1 --json`.
4. Keep the reopened attempt running long enough to invoke `ultrafuzz status synthetic-run --project /tmp/uf-reset-ledger/project --json` again.

The status payload reports a running workflow but also returns `ok: false` with:

```text
NODE_ATTEMPT_LEDGER_WRITE_FAILED
node attempt ["synthetic-run", 42] was already recorded with different immutable data
```

No model or target-specific artifact is required; the collision is between reset semantics and the generic immutable attempt-ledger contract.

## Expected behavior

A reset execution must receive a distinct immutable ledger identity, or be represented as a new reset revision without rewriting the prior attempt row. Historical terminal rows remain unchanged, while status and stats continue to emit schema-valid snapshots during the reopened execution.

## Acceptance criteria

- A model-free fixture projects a failed attempt, resets that node, and synchronizes while the replacement attempt is running.
- The prior ledger row remains byte-for-byte unchanged.
- The replacement execution has a distinct stable identity and lifecycle.
- `status --json` and `stats --json` do not emit an attempt-ledger write diagnostic.
- Repeated synchronization is idempotent.

aviggiano:
## Summary

- canonicalize attempt input manifest hashing so JSON object key order does not change immutable evidence
- reconcile legacy order-sensitive hashes only when they re-derive from an authenticated task snapshot with identical semantics
- preserve distinct terminal source-event identities when a reset reuses the runner attempt number

## Validation

- focused reset/legacy-digest regression test
- existing forged immutable-provenance rejection test
- Prettier, ESLint, and `git diff --check`

Fixes #1036


aviggiano:
Holding this PR. The stated root cause is not demonstrated, and the change alters a durable format.

- Ledger identity is `[workflow_run_id, source_event_sequence]` (`attempt-ledger.ts:771-775`). Smithers 0.35 allocates `seq` as MAX+1 and `timetravel` does not delete event rows, so a reset cannot reuse a sequence. Main already drops a recorded superseded occurrence when the same attempt number restarts (`workflow-sync.ts:4964-4998`; test at `runtime.test.ts:21275`).
- No path on main reorders task JSON keys between syncs. Sync reads the digest-checked sealed bytes, the controller snapshot keeps `contents.tasks`, and the dynamic rewrite uses `structuredClone`.
- The new test fails on main only at `assert.notEqual(input_sha256, legacyInputDigest)`. It hand-edits `attempts.jsonl` and tests a migration path, not a reset conflict.
- Changing `manifests.input_sha256` for every new row makes a v0.0.23 reader reject rows written by this version as "different immutable data".
- The new throw `sealed workflow is missing its task manifest` runs outside the diagnostic conversion. `status` and `inspect` would crash on seals created before #621.

To reopen: add a test that reproduces the #1036 flow against main (record a failed attempt, `--reset-node` while in progress, sync) and names the immutable field that differs. If key-order drift is proven, accept both digests at the reconcile step in two lines. Do not change the write format for new rows.

aviggiano:
Closing this implementation during stabilization. The proposed canonical digest write format changes immutable ledger hashes and is incompatible with existing readers, while the actual #1036 reset/key-order failure has not been reproduced. The existing recorded-occurrence reset regression passes.

Keep #1036 open. A replacement should first reproduce the reported reset flow, identify the exact immutable field that differs, preserve the existing write format, and cover repeated synchronization and reader compatibility.

PR #1053, opened 2026-09-02
fix(cli): accept additive dynamic tasks in stats

### LINKED ISSUE #1052: stats rejects additive workflow tasks after dynamic graph expansion
## Summary

`ultrafuzz stats` rejects a coherent run after supported runtime graph expansion adds workflow tasks. Run metadata retains the sealed launch-time task IDs, while the current planned graph contains that baseline plus dynamically generated tasks. Statistics currently requires exact set equality and fails before producing a snapshot.

## Minimal reproduction

1. Create a synthetic linked run with one sealed workflow task in run metadata and the planned graph.
2. Apply the supported dynamic-runtime expansion so the current graph contains the original task plus one generated task.
3. Keep the workflow binding, graph fingerprint, state provenance, and ledgers coherent.
4. Run:

```text
ultrafuzz stats synthetic-run --project /tmp/uf-dynamic-stats/project --json
```

Actual result:

```text
RUN_STATS_FAILED
planned graph workflow task IDs do not match run metadata
```

## Expected behavior

Statistics should treat the metadata task IDs as the sealed baseline and accept a current graph that is an additive superset, matching the runtime integrity rule for dynamic expansion. It must still reject a graph that removes or replaces a baseline task.

## Affected component

CLI statistics evidence binding validation.

## Acceptance criteria

- A model-free regression covers one baseline task plus one additive dynamic task.
- `stats` derives a typed snapshot containing both nodes.
- Removing the baseline task remains an error.
- Static exact-match behavior remains unchanged.

aviggiano:
## Summary

- accept additive workflow task IDs introduced by supported dynamic graph expansion
- continue rejecting removal of sealed launch-time task IDs
- cover both cases with a model-free statistics regression

## Tests

- `pnpm --filter @ultrafuzz/cli... build`
- `pnpm --filter @ultrafuzz/cli exec tsc -p tsconfig.test.json`
- `node --test --test-concurrency=1 packages/cli/dist-test/test/run-statistics.test.js`
- focused Prettier and ESLint checks

Fixes #1052


aviggiano:
Closing as superseded by #1076.

Both PRs fix #1052 at `run-statistics.ts:335`. This PR replaces the exact match with a subset check, which also accepts extra task IDs on static nodes. Issue #1052 requires static exact-match behavior to stay unchanged. The regression test here builds the extra node with `graphDocument()`, which sets no `dynamic_generated` field, so it locks in the wider acceptance. On the report-bundle path the task-ID match is the only content binding between `graph.json` and `run.json`, so the widening matters there.

#1076 accepts added task IDs only when the node carries `dynamic_generated`, keeps rejecting removals and static additions, and tests the static-extra reject case. The two PRs conflict textually, so only one can land.

PR #1072, opened 2026-09-02
fix(runtime): drain streamed inspection output

### LINKED ISSUE #1071: Streaming inspection drops buffered stdout after child close
## Generic symptom

A one-shot streaming inspection can return only an early prefix of valid NDJSON and report `truncated: false` when the child emits a burst and exits immediately.

## Minimal reproduction

1. Bind a synthetic local runner under `/tmp/uf-stream-tail/runner.js`.
2. Have it synchronously write several thousand numbered NDJSON records to stdout and exit successfully.
3. Call the runtime streaming inspection helper with a line limit above the emitted count.

## Expected behavior

The helper waits for both child termination and line-reader drain, delivers every numbered record in order, and reports a successful non-truncated result.

## Actual behavior

The child `close` handler resolves the stream promise first. Finalization then closes the readline interface and destroys stdout while buffered records remain, so the consumer receives a prefix with `truncated: false`.

## Affected component

Runtime streaming Smithers command lifecycle used by event and node watch/inspection surfaces.

## Non-sensitive diagnostics

The failure uses a synthetic executable, local pipes, and numbered JSON only. Buffered one-document execution of the same runner returns the complete output.

aviggiano:
## Summary

- wait for both child termination and readline drain
- preserve buffered NDJSON emitted immediately before process exit
- cover a 4,000-record final stdout burst

## Validation

- `pnpm --filter @ultrafuzz/runtime... build`
- `pnpm --filter @ultrafuzz/runtime exec tsc -p tsconfig.test.json`
- focused streaming executable-capability tests
- focused Prettier and ESLint checks

Fixes #1071


aviggiano:
Closing: the reported truncation does not reproduce on main, and the change introduces a hang.

**Not reproducible on main**
- The PR's own new test passes against main source (PR test file, `origin/main` src).
- A 20,000 × 1 KB burst against the main build delivered all 20,000 records in order with `truncated: false`. Node emits ChildProcess `close` only after the stdout pipe has emitted `end`, and readline has already emitted every complete line by then. The `finally` block's `reader.close()` has nothing left to drop.
- The only prefix that can be produced is child-side: the child calls `process.exit(0)` before its queued pipe writes drain. Main and this PR both deliver the same 58 of 20,000 records in that case.

**New defect: hang on abort during preparation**
If the abort signal fires while `prepareSmithersExecutableEnvironment` is pending, the existing post-prepare check (`smithers.ts` ~4868-4870) calls `stopStreaming()`, which calls `reader.close()`. readline emits `close` synchronously, before `reader.once("close")` is attached inside the promise executor (~4901). `readerClosed` stays false, `resolveWhenDrained` never resolves, and the `finally` cleanup never runs. Reproduced: the PR build timed out at 8 s; main resolved in 386 ms with `stoppedByCaller: true`. The path is reachable through the public runtime API whenever a caller passes a signal (`lifecycle-inspection.ts` 426, 482, 555). The CLI does not pass a signal today.

Reopen with a reproduction that fails on `origin/main` under Node 24. If a drain guarantee is still wanted, create the reader-closed promise right after `createInterface`, before any `stopStreaming()` call, and await it after `exit`.

PR #1078, opened 2026-09-02
Rehydrate verified results from identical dynamic generations

### LINKED ISSUE #1077: Identical dynamic expansion recovery does not rehydrate verified completed attempts
## Generic symptom

An explicit retry archives a complete dynamic-expansion generation. If the producer later emits a byte-identical manifest set and the runtime renders identical prompts, verified completed generated attempts remain only in history and are scheduled again.

This preserves evidence but wastes completed agent work and can make successful progress appear to regress after recovery.

## Minimal reproduction

1. Start a synthetic workflow with one dynamic producer and a generated fan-out.
2. Complete at least one generated attempt so its rendered prompt, publications, and verification marker are durable.
3. Explicitly retry the producer, which archives the active dynamic generation and its attempt state.
4. Regenerate a byte-identical manifest set and render the same prompt.
5. Resume the workflow.

## Expected

The runtime rehydrates an archived attempt only when all recovery authority matches:

- the complete manifest generation is identical;
- the exact rendered prompt matches;
- the verification marker is schema-valid and names the same attempt;
- every declared publication is a regular file inside the archived artifact directory and matches its recorded digest.

Failed, unverified, incomplete, changed-prompt, or changed-generation attempts remain pending.

## Actual

The new generation starts with empty active artifact and verification roots. Previously verified attempts are scheduled again even though the compatible result remains in dynamic-expansion history.

## Scope

This is runtime dynamic-expansion retry/recovery behavior. Reproduction does not require a provider, network call, or particular target.

aviggiano:
## Summary

- match archived and active dynamic manifest generations before considering reuse
- require the exact rendered prompt, current output-contract authority, a valid matching verification marker, and every publication digest
- republish only verified publications and the prompt; never restore mutable workspaces or unverified sidecars
- emit normal Smithers preparation, agent-process, and verification outputs without another model call
- keep recovery idempotently active across workflow rerenders

## Tests

- runtime package build
- runtime test TypeScript compile
- dynamic-expansion tests
- generated-workflow-verifier tests
- focused ESLint and Prettier checks

This draft is stacked on #1064, which preserves the complete dynamic generation needed as recovery authority.

Fixes #1077


aviggiano:
Deferring this PR past v0.1.0. It is a feature, and it has a confirmed regression.

- #1077 describes wasted work, not incorrect behavior. The state exists only after #1064 archives a generation; main has no archive step.
- The rehydrated agent-process Task has no agent, so Smithers records no `attemptMeta.agentChainIndex`. `workflow-sync` inspects every terminal attempt of a local task, and `reconcileSmithersAttemptAgentSelection` throws "no sealed agent-chain selection"; `syncRun` returns `WORKFLOW_ATTEMPT_INSPECT_FAILED`. Confirmed by running a compute Task on the pinned Smithers 0.35.0 and inspecting the persisted meta. `stats` fails, and `status` / `inspect` stop advancing run state, in exactly the retry scenario this PR targets.
- `listRegularArtifactFiles` throws `DYNAMIC_RETRY_REHYDRATION_CHANGED` on any symlink or non-regular entry in an active artifact dir. One stray entry makes every render fail while a matching archive exists.
- The recovery match ignores `admitted_dependency_attempt_ids` and dependency artifact digests.
- No test executes the rendered Tasks or `syncRun`. CI-mode strict lint reports 5 errors in `dynamic-expansion-retry.ts` (function length, statements, complexity, max-lines); the advisory gate hid them.

If pursued later: record rehydrated attempts as reuse ledger entries bound to the archive evidence, or restore Smithers state instead of rendering no-agent Tasks, and add an integration test that renders, completes, and syncs a rehydrated attempt. This branch also needs a rebase once #1064 lands in trimmed form.

aviggiano:
Note on the close: GitHub closed this PR automatically at 00:02 UTC when #1064 merged and its head branch `fix/dynamic-retry-manifest-1063`, the base of this stacked PR, was deleted. The deferral above still stands. When this work is picked up after v0.1.0, rebase `fix/rehydrate-dynamic-results-1077` onto `main` (which now contains the trimmed #1064 with the state.json prune) and reopen against `main`.

PR #1082, opened 2026-09-02
Rebind stale preparation authority after dependency replacement

### LINKED ISSUE #1081: Persisted preparation authority survives a replacement dependency publication
## Generic symptom

A downstream task has valid persisted pre-agent preparation evidence from an earlier execution. An upstream dependency is later reopened and publishes a different, authenticated workspace patch from the same pinned source tree. When the downstream task is reopened, preparation restores its old tree and then strict dependency replay rejects the replacement patch with a base-tree mismatch.

## Minimal reproduction

1. Create a synthetic workflow with producer A and downstream producer B.
2. Let A publish an authenticated patch from tree `base` to `result-1`.
3. Let B prepare successfully and persist `result-1` as its pre-agent preparation tree.
4. Reopen A and publish a different authenticated patch from `base` to `result-2`.
5. Reopen B and run its pre-agent preparation.

## Expected behavior

The runtime recognizes that B’s schema-valid preparation record belongs to the superseded dependency publication, preserves strict patch validation, restores the authenticated current dependency base, replays the replacement patch, and durably binds B’s new preparation to `result-2`.

Malformed or modified preparation evidence, unauthenticated dependency artifacts, non-chain dependency captures, and unrelated worktree drift must still fail closed. The post-agent path must continue preserving current model-authored source.

## Actual behavior

B restores `result-1`; applying A’s replacement patch then fails because that patch requires `base`. Automatic preparation retries repeat the same deterministic mismatch.

## Affected component

Generated workflow pre-agent workspace restoration and dependency-patch replay authority.

## Relationship to existing fixes

This is distinct from #949: that issue restores a reopened producer over unchanged dependency authority. Here the restored preparation is internally valid but stale because an authenticated dependency publication changed after it was recorded.

aviggiano:
## Summary
- detect a reopened producer whose authenticated dependency publication changed
- restore and replay from the authenticated replacement base
- atomically rebind runtime-owned preparation evidence while preserving strict drift checks

## Tests
- dependency-inclusive runtime build
- runtime typecheck
- focused ESLint
- workspace-patch replay suite (23 tests)

Fixes #1081


aviggiano:
Holding this PR. The bug is real and deterministic on main (`workflow.tsx:4285-4307`, `4117-4123`; `workspace-handoff.ts:433-436`). The fix does not hold across preparation, and an existing test fails on the branch.

1. **Existing test fails.** `generated-workflow-verifier.test.ts` "generated dependency admission retains one exact snapshot epoch and never hydrates a replacement patch pair" fails on this branch with `ReferenceError: supersededPreparationDependencyBase is not defined`. The harness slices the template from `materializeWorkspacePatchDependencies` and injects a fixed collaborator list; the two new helpers sit before the slice start. CI skipped that lane because the advisory gate failed first.
2. **The rebind is undone by the next preparation step.** In create mode `restore-invariant-suite-snapshot` runs right after the materializer (`workflow.tsx:3842-3847`). It rewrites `src/`, `contracts/`, `test/`, `tests/` from the generation-1 run-root snapshot (main `5000-5072`). Capture and reconcile return early while that snapshot exists (`4926-4928`, `6522-6523`), and no reset path removes it. Dependency patches cannot change `src/` or `contracts/`, so the replacement's changes live in `test/` and are reverted. The worktree ends at result-1 content while preparation and baseline say result-2. B's later patch, captured against the rebound baseline, encodes a reversion of A's replacement. On main this scenario fails closed before any mutation; the PR turns it into a silent inconsistent state. This was established by code trace on both branches, not executed.
3. Minor: the `expectedBaseline/persistedBaseline` check at `4221` duplicates the one at `4155` and cannot throw.

To reopen: rebind or invalidate every persisted pre-agent evidence store of the attempt (run-root invariant workspace snapshot, invariant-suite baseline and its protected copy, any stale patch pair), or fail closed with a message that names the stale stores. Add an end-to-end test through `prepareArtifactMirror` with a real worktree, a persisted generation-1 snapshot, and a replacement patch that changes a `test/` file; assert the final tree equals the chain end. Repair the verifier harness.

aviggiano:
Superseded by #1115. Rebinding only the preparation tree and patch baseline leaves the old invariant workspace snapshot active; the full preparation sequence can restore generation-1 source after authority has moved to generation 2.

The replacement validates and archives all affected pre-agent stores, replays the authenticated replacement, and regenerates coherent evidence. It includes real-Git full preparation and interruption tests, plus protected prior dependency provenance for future runs. Legacy preparations without that provenance are not silently migrated. The original branch is preserved.

PR #1089, opened 2026-09-04
fix(runtime): make the observer synchronization deadline opt-in

aviggiano:
## Summary

`observationSynchronizationDeadline` no longer supplies a 15-second default. When `ULTRAFUZZ_OBSERVATION_SYNC_TIMEOUT_MS` is unset or empty it returns `undefined`, so `getRunStatus`, `getRunHealth`, `diagnoseRun`, and `stats` wait for full synchronization. A positive integer still bounds the refresh and is still capped at 60000 ms; `0` and `off` still mean no deadline; unparseable values now also mean no deadline instead of falling back to a default. The warning text, the downgrade to a warning in `getRunHealth`, the atomic node write group, and `describeObservationSynchronizationDeadline` from #1059 are unchanged. The unit tests in `observation-deadline.test.ts` are updated for the new default, and `docs/reference/configuration.md` now documents the variable as opt-in.

## Why

#1059 introduced a 15 s default deadline for the read-only observers. On GitHub CI runners the runtime test `getRunStatus synchronizes without appending duplicate events` now fails on main (run 33814847949, shard 4/4): the observe-side synchronization hits the default deadline, the status stays `running` with a `WORKFLOW_SYNC_DEADLINE_EXCEEDED` warning, and the test's strict equality on the synchronized status fails. It passes on a fast local machine only. A default that trips on ordinary CI hardware will also trip for users with large histories or slow disks and make `status` and `stats` stale by default, which is a behavior regression before v0.1.0. Observers are the only CLI path that converges local run state from runner evidence, so the safe default is to wait; the bound stays available for anyone who wants it. The `getRunHealth` deadline test already set the variable explicitly to `1`, so no runtime test relied on the default.

## Tests

- `node --test --test-name-pattern "observation" dist-test/test/observation-deadline.test.js`: 5 pass, 0 fail.
- `node --test --test-name-pattern "getRunHealth returns live runner health when observation synchronization exceeds its deadline" dist-test/test/runtime.test.js`: 1 pass.
- `node --test --test-name-pattern "getRunStatus synchronizes without appending duplicate events" dist-test/test/runtime.test.js`: 1 pass.
- `node scripts/docs-check.mjs`: exit 0.
- `CI=1 ESLINT_PLUGIN_DIFF_COMMIT=origin/main ULTRAFUZZ_STRICT_LINT=1 eslint --max-warnings 0` on the changed TypeScript files: clean. `prettier --check` on all changed files: clean.

<h3>Greptile Summary</h3>

This PR makes the observer synchronization deadline opt-in so read-only commands wait for complete synchronization by default.

- Returns no deadline for unset, empty, disabled, unsafe, or otherwise invalid timeout values.
- Retains positive timeout handling and the 60000 ms cap.
- Updates focused tests, runtime documentation, and synchronization comments to match the new behavior.

The PR appears safe to merge with no concrete blocking or independently actionable non-blocking issues identified.

The implementation consistently propagates the documented opt-in deadline semantics, retains bounded explicit overrides, and updates focused tests and documentation accordingly.

| Filename | Overview |
|----------|----------|
| packages/runtime/src/workflow-sync.ts | Removes the default 15-second observer deadline while preserving strict positive-integer parsing and the 60-second cap. |
| packages/runtime/src/state-export.ts | Updates synchronization documentation to describe the deadline as opt-in without changing runtime behavior. |
| packages/runtime/test/observation-deadline.test.ts | Updates focused coverage for unset, empty, invalid, disabled, positive, and capped timeout values. |
| docs/reference/configuration.md | Documents that observer synchronization is unbounded by default and explains the opt-in timeout behavior. |

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(runtime): make the observer synchron..."](https://github.com/monad-developers/ultrafuzz/commit/d04745819124fb58e20e7b044cddcef993e285d3) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=60220670)</sub>


aviggiano:
Merging with the smoke job red. The only failed step is "Enforce production dependency advisory policy", which has been timing out against registry.npmjs.org/-/npm/v1/security/advisories/bulk since about 22:20 UTC; the build and pr-smoke steps before it passed, and the same step fails on main for the same reason. The change was verified locally: the observation-deadline unit tests, the getRunHealth deadline test, and the previously failing `getRunStatus synchronizes without appending duplicate events` all pass, and an independent review confirmed the delta touches only the default. This restores the pre-#1059 default behavior of the observers and clears main's shard 4 failure.

PR #1114, opened 2026-09-08
fix(runtime): recover report retry authority after restart

### LINKED ISSUE #1100: Final-report retry prompt loses earlier attempts across controller restarts
After a final-report producer is retried across a controller restart, the report prompt's `run_metadata.agent_execution` can omit earlier failed attempts. A later verifier reconstructs the full durable history and rejects the report that copied the prompt authority.

Observed recovery required replacing only this metadata from Smithers' recorded attempts and rerendering the report. Prior failed attempts existed in durable history but were absent from the prompt-derived report. Fixing the separate missing `smithersRunId`/`smithersNodeId` reconstruction fields allowed the verifier to reach this mismatch; it did not restore the prior attempts to the prompt projection.

Relevant current code: `packages/runtime/src/templates/smithers/workflows/workflow.tsx`.

- `artifactAwareAgent` builds selections from the process-local `finalReportAgentSelectionAuthority` map, defaulting to an empty array.
- It appends the current attempt and writes `materializeFinalReportPromptAuthority`.
- `authoritativeFinalReportAgentExecution` can reconstruct all attempts from `smithers node --full-output` after a fresh process starts.

Thus prompt-time and verifier-time authority can describe different histories after a restart. This is an identified path and an observed symptom; a complete model-free restart regression remains to be added.

Suggested fixture: fail attempt 1, stop the controller, resume attempt 2, have a stub producer copy the supplied prompt metadata into a valid report, and restart before verification. Both prompt and verifier should include attempt 1 as failed and attempt 2 as producer. Repeat with multiple chain entries and with an attempt number that differs from the chain index.

Acceptance criteria:

- Seed report provenance from authenticated durable history at the retry boundary, before constructing the prompt authority.
- Keep actual attempt numbers distinct from planned chain positions.
- Preserve the same metadata across rerenders and verifier-only restarts.
- Do not require a model call or a manual report edit merely to repair orchestration-owned attempt metadata.

Sanitized: no report findings, target identity, provider details, run IDs, host paths, or raw logs are included.

aviggiano:
After a final-report retry crosses a controller restart, the prompt loses earlier attempts while a fresh verifier later recovers them and rejects the report. Seed missing prompt provenance from durable Smithers history, preserve physical attempt numbers separately from chain positions, and share the strict history projection with verifier recovery. Keep complete in-process history for ordinary retries and retain the cloud worker's independent single-attempt behavior.

Validation: Node 24 runtime build/test compilation; 152 broader regressions; final 4 history tests and 2 real Smithers restart integrations. Each integration runs three separate controller processes through producer failure, resumed report generation, and verification; the quota case retries within one chain entry. The pre-fix integration reproduces #1100, and the fix passes without editing the produced report. Strict lint and formatting pass.

Closes #1100.


aviggiano:
Validation for exact head `ca6c4787d86bfc79714c83c0bdd471019c55d9ed` is complete. All build/static checks, all four integration shards, runtime support/Bun contracts and the release aggregate pass in CI run 34227815471.

The original shard 1 reached its 75-minute job limit while still advancing, with 49 assertions passing and output 30 seconds before timeout. Its unchanged registry/pricing fixtures ran several times slower than the corresponding prior passing run; these paths do not execute the new report-history helper. The next selected test also passed locally on this exact head. One retry of that shard alone passed in 35m51s; the other successful lanes were retained. No source changes, timeout increases or check bypasses were used to obtain the pass.

